PR #1, opened 2019-04-30
Remove previous params

vijay120:
This PR does the following:

1) Removes the `previous_params ` field since it is not being used in any application and I am unclear why we have it
2) Fix the bug where the request object does not persist the previous turn's responder

### DIFF
--- a/mindmeld/app_manager.py
+++ b/mindmeld/app_manager.py
@@ -117,9 +117,9 @@ class ApplicationManager:
 
     def _pre_dm(self, processed_query, context, params, frame, history):
         request = self.request_class(context=context, history=history, frame=frame,
-                                     params=FrozenParams(previous_params=params), **processed_query)
+                                     params=params, **processed_query)
 
-        response = self.responder_class(frame=frame, params=Params(previous_params=params),
+        response = self.responder_class(frame=frame, params=Params(),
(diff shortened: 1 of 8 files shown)


mindmeld/app_manager.py
```diff
@@ -117,9 +117,9 @@ def load(self):
 
     def _pre_dm(self, processed_query, context, params, frame, history):
         request = self.request_class(context=context, history=history, frame=frame,
-                                     params=FrozenParams(previous_params=params), **processed_query)
+                                     params=params, **processed_query)
```
minhtuev: We are assuming that params is a frozen params object here?

minhtuev:
Why do we need user state? Can we pass it in one of the existing fields, such as context?

mindmeld/app_manager.py
vijay120: The params to this function passes in a frozen params:

```
        params = freeze_params(params)
.
.
.

        request, response = self._pre_dm(processed_query=processed_query,
                                         context=context, history=history,
                                         frame=frame, params=params)
```

vijay120:
@minhtuev We need user state since there is no way to persist a custom key,value in the dialogue handler for a single turn.

minhtuev:
We should add tests for the bug fix

PR #4, opened 2019-05-02
Small update for duckling and docker

minhtuev:
1. Relax the constraint to find duckling pid
2. Update docs with port mapping for duckling docker
3. Docker container does not need mindmeld config
4. ES cluster name to be mindmeld

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -52,8 +52,6 @@ RUN chown -R mindmeld:mindmeld /elasticsearch /data /var/log
 # Add Config Files
 COPY ./conf/elasticsearch.yml /elasticsearch/config/elasticsearch.yml
 
-COPY ./conf/config.cfg /root/.mindmeld/config
-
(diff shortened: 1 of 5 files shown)


docker/Dockerfile
```diff
@@ -67,6 +65,7 @@ RUN pip3 install --upgrade pip
 RUN pip3 install -U mindmeld
 RUN pip3 install click-log==0.1.8
 
+ARG REBUILD_BLUEPRINT=unknown
```
vijay120: what is this?

minhtuev: This is a trick to help rebuild docker container after a specific step

https://stackoverflow.com/questions/35154219/rebuild-docker-image-from-specific-step

minhtuev: Otherwise, Docker will use a cached version of the assistant

vijay120: :)

PR #7, opened 2019-05-03
Update build process

minhtuev:
Update layout to include Google Analytics


minhtuev:
@vijay120 : I checked individual commit; it is possible that I might miss something but we can look at it from the staging server for reference.

PR #15, opened 2019-05-14
Add the 'sleep' directive

racampos:
Added the 'SLEEP' directive name and the sleep() function.

### DIFF
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -48,6 +48,8 @@ class DirectiveNames:
     SUGGESTIONS = 'suggestions'
     """A view for a list of suggestions."""
 
+    SLEEP = 'sleep'
+    """A directive to sleep."""
 
 class DirectiveTypes:
     """A constants object for directive types."""
@@ -655,6 +657,10 @@ class DialogueResponder:
         """
         self._logger.warning('prompt() is deprecated. Please use reply() and listen() instead')
         self.reply(text)
+
+    def sleep(self):
+        """Adds a 'sleep' directive."""
+        self.act(DirectiveNames.SLEEP)
 
     @staticmethod
     def _choose(items):
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -658,9 +658,9 @@ class DialogueResponder:
         self._logger.warning('prompt() is deprecated. Please use reply() and listen() instead')
         self.reply(text)
 
-    def sleep(self):
+    def sleep(self, delay=0):
         """Adds a 'sleep' directive."""
-        self.act(DirectiveNames.SLEEP)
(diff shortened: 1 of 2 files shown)


mindmeld/components/dialogue.py
```diff
@@ -656,6 +658,10 @@ def prompt(self, text):
         self._logger.warning('prompt() is deprecated. Please use reply() and listen() instead')
         self.reply(text)
 
+    def sleep(self):
```
minhtuev: Can you add an optional param `delay` with default `0`?

Delay is the number of milliseconds to wait before putting the client to sleep.

mindmeld/components/dialogue.py
```diff
@@ -656,6 +658,10 @@ def prompt(self, text):
...
+        """Adds a 'sleep' directive."""
+        self.act(DirectiveNames.SLEEP)
```
minhtuev: `payload={'delay': delay}`

racampos:
I added the optional 'delay' param and documented the new function.

mindmeld/components/dialogue.py
```diff
@@ -656,6 +658,10 @@ def prompt(self, text):
         self._logger.warning('prompt() is deprecated. Please use reply() and listen() instead')
         self.reply(text)
 
+    def sleep(self, delay=0):
+        """Adds a 'sleep' directive."""
```
minhtuev: document the delay argument - see line 655

https://github.com/cisco/mindmeld/pull/15/files#diff-c21e7ac893d7b0e1d68df2df2ce7ff52R655

racampos:
I documented the 'delay' param in the sleep directive.

source/userguide/dm.rst
```diff
@@ -241,6 +241,9 @@ The table below details the ``responder`` methods to send actions, also termed `
 +-------------------------------+----------------------------------------------------------------+
 | :py:meth:`responder.direct`   | Used to send an arbitrary directive object                     |
 +-------------------------------+----------------------------------------------------------------+
+| :py:meth:`responder.sleep`    | Used to put the client to sleep during a specified number of   |
+|                               | milliseconds                                                   |
```
serapio: *during* or *after*? It would be good clarify this in dialogue.py:52 also, that this `sleep` is dismissing to a standby mode after a delay, not like another meaning of `sleep` like sleeping a process for a duration.

minhtuev: I think it should be "after" right?

minhtuev: Is this an accurate description: "Dismissing to a standby mode after a delay."

racampos: I changed the word "during" for after" in dm.rst:244 and added a better description in dialogue.py:52.

PR #28, opened 2019-06-21
Loosen dependency

minhtuev:
Remove bleach, add pytest-asyncio and loosen numpy


setup.py
```diff
@@ -9,16 +9,15 @@
...
+    'click-log==0.1.8',
     'distro~=1.3',
     'elasticsearch5~=5.5',
     'Flask~=0.12',
     'Flask-Cors~=3.0',
     'future~=0.16.0',
     'nltk~=3.2',
-    'numpy>=1.14,<=1.14.5; python_version < "3.7"',
+    'numpy~=1.14; python_version < "3.7"',
     'numpy~=1.15.0; python_version >= "3.7"',
```
jre21: Since this is a library meant for external consumption, I'd rather we not be overly restrictive in our version constraints unless certain versions of a library are known to break our software.

@vijay120 , were there specific reasons that you wanted us to only use numpy 1.15.x versions?  If not, I'd be more comfortable with using `numpy~=1.14` regardless of python version.  The exact workings of the ~= specifier are described [here](https://www.python.org/dev/peps/pep-0440/#compatible-release).

vijay120: I would need to test it with the 1.14 versions for py3.7, doing that now.

vijay120: @jre21 python 3.7 is only supported in numpy>=3.7. (https://github.com/numpy/numpy/releases?after=v1.15.2), which is why this was done.

jre21: Okay, that much makes sense, although having downstream packages manage the dependencies of their dependencies is kind of icky.

More to the point, though, right now we are allowing numpy 1.16.x on python 3.6 but not on python 3.7, which makes no sense.  We need to pick one or the other.  My preference would be to change the >= 3.7 requirement to `numpy~=1.15` so as to allow upgrades to future versions short of an eventual 2.0 release, but if you'd rather be more conservative than that, at least be _consistently_ conservative.

vijay120: That makes sense. https://github.com/numpy/numpy/releases/tag/v1.15.4 is the last version of numpy that supports python 2.7 and 3.4-3.7, so it makes sense to just do `numpy~=1.15`. @minhtuev Can you add that?

jre21: Incidentally, numpy 1.14.6 also supports python 3.7.

minhtuev: @vijay120 Done

vijay120: @jre21 can you approve?

PR #30, opened 2019-06-26
Docker setup for MindMeld deps

vijay120:
## Problem
Developers have a hard time installing MindMeld dependent services, namely Elasticsearch (required Java) and Num-Parse (requires a binary executable that run on only select OSs). 

We already have a docker container that has ALL of MindMeld deps and the MindMeld project in the container, but this is cumbersome to use since developers need to sh into the container, make changes using bash editors like vi, save and commit those changes and then run the MindMeld server to make any improvement on their project.

## Idea
Create a docker container containing only MindMeld dependent services, namely Elasticsearch and Num-Parse, while letting the developer run the MindMeld project locally. This would involve installing the `mindmeld` lib locally, which is not an issue.

# Workshop Instructions

## Install libraries need to for local development (python3, pip and virtualenv):

```bash
sh docker_containers/mindmeld_dep_docker/run.sh
```

## Run Mindmeld Dependent services in Docker:

We assume the developer has Docker installed on their local machine. If not, we will direct them to install it and get going.

```bash
docker pull mindmeldworkbench/dep:latest
docker run -p 0.0.0.0:9200:9200 -p 0.0.0.0:7151:7151 -p 0.0.0.0:9300:9300 mindmeldworkbench/dep -ti -d
```

## Create and Deploy MindMeld project

Open up and new terminal and run the following:

```bash
mkdir home_assistant
curl https://devcenter.mindmeld.com/bp/home_assistant/app.tar.gz  -O -J -L
# wget https://devcenter.mindmeld.com/bp/home_assistant/app.tar.gz
tar -xvf app.tar.gz -C ./home_assistant
virtualenv -p python3 .
source bin/activate
pip install mindmeld
python3 -m home_assistant build
python3 -m home_assistant converse
```


source/userguide/getting_started.rst
```diff
@@ -24,138 +24,78 @@ If you're going to be using MindMeld often, **we recommend you do the virtualenv
...
 .. code-block:: shell
 
-   docker pull mindmeldworkbench/mindmeld
-   docker run -p 0.0.0.0:7150:7150 mindmeldworkbench/mindmeld -ti -d
-
-The Docker container contains Elasticsearch, the numerical parsing service, the MindMeld library and the Home Assistant application for you to test. The container will build and serve the application on port 7150 which is exposed to the external environment.
-
-The application code and data is located at directory ``/root/home_assistant`` on the docker container.
+   docker pull mindmeldworkbench/dep:latest
+   docker run -p 0.0.0.0:9200:9200 -p 0.0.0.0:7151:7151 -p 0.0.0.0:9300:9300 mindmeldworkbench/dep -ti -d
```
arushir: Can we include a specific note in the documentation that ports 7151, 9200 and 9300 will be used by the Docker container, so users should make sure that any other applications don't clash with those ports. And if they already have ES installed and are using it for other applications, then virtualenv setup is recommended

PR #31, opened 2019-06-28
Updated requests lib

vijay120:
Updated the requests library version due to version conflicts


jjjacksn:
I think this restricted the requirements a bit more than you intended. For references, here are the docs for `~=`, or [compatible releases](https://www.python.org/dev/peps/pep-0440/#compatible-release) in requirements.

Before:
`requests>=2.20, requests==2.*`

After:
`requests>=2.20.1, requests==2.20.*`

This means we no longer support requests 2.21 or newer. 

What was the issue you were seeing?

setup.py
```diff
@@ -26,9 +26,9 @@
     'scipy>=0.9,<2.0',
     'scikit-learn>=0.18.1,<0.20; python_version < "3.7"',
     'scikit-learn>=0.19.2,<0.20; python_version >= "3.7"',
-    'requests~=2.20',
+    'requests~=2.20.1',
     'tqdm~=4.15',
-    'urllib3==1.25.3',
+    'urllib3==1.24.3',
```
jjjacksn: We shouldn't be locking any package to any single specific version. If there is some issue with 1.25 we can disable that version.

PR #38, opened 2019-07-09
Adding Travis definition

minhtuev:
Adding Travis CI to run builds and tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 dist: xenial
 language: python
 python:
-  - "3.4"
   - "3.5"
   - "3.6"
   - "3.7"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,22 @@
+dist: xenial
+language: python
+python:
+  - "3.6"
```
vijay120: Where are we only validating 3.6? MindMeld supports 3.4-3.7

minhtuev: We were running into some issues with Travis CI for other versions
https://travis-ci.org/cisco/mindmeld/builds/556563795?utm_source=github_status&utm_medium=notification

PR #39, opened 2019-07-10
QuestionAnswerer for unstructured text

vembar:
Currently, we support QA on knowledge bases that are well structured i.e, fuzzy matching on phrases, keywords or short spans of text. This change supports QA over larger chunks of unstructured text (passages) by introducing a new analyzer that tokenizes, removes stop words, does stemming of text before performing matching. We also introduce some fuzziness in filter queries using char_ngram matching.

Users can now specify query_type (keyword/text) along with their queries to trigger the right analyzers.


minhtuev:
Rebase to master to enable Travis automation

mindmeld/components/question_answerer.py
```diff
@@ -765,15 +793,16 @@ def validate(self):
...
+        def __init__(self, field, field_info=None, value=None, query_type='keyword', range_gt=None,
+                     range_gte=None, range_lt=None, range_lte=None):
             """Initialize a knowledge base filter clause. The filter type is determined by whether
             the range operators or value is passed in.
             """
 
             self.field = field
             self.field_info = field_info
             self.value = value
+            self.query_type=query_type
```
arushir: Nit: fix lint errors like this `self.query_type=query_type` -> `self.query_type = query_type`

mindmeld/components/question_answerer.py
```diff
@@ -78,6 +78,7 @@ def get(self, index, size=10, **kwargs):
         Args:
             index (str): The name of an index.
             size (int): The maximum number of records, default to 10.
+            query_type (str): Whether the search is over structured or unstructured text
```
minhtuev:
```suggestion
            query_type (str): Whether the search is over structured or unstructured text.
```

mindmeld/components/question_answerer.py
```diff
@@ -470,6 +475,7 @@ def filter(self, **kwargs):
                 >>> s.filter(field='price', gte=1, lt=10)
 
         Args:
+            query_type (str): Whether the filter is over structured or unstructured text
```
minhtuev:
```suggestion
            query_type (str): Whether the filter is over structured or unstructured text.
```

mindmeld/components/question_answerer.py
```diff
@@ -797,13 +826,22 @@ def build_query(self):
...
                     }
                 else:
-                    clause = {
-                        "match": {
-                            self.field + ".normalized_keyword": {
-                                "query": self.value
+                    if self.query_type == 'text':
+                        clause = {
+                            "match": {
+                                self.field + ".char_ngram": {
```
mhuang2173: just wonder is this supposed to be against `processed_text` rather than `char_ngram` field?

mindmeld/components/question_answerer.py
```diff
@@ -455,6 +458,8 @@ def filter(self, **kwargs):
         For example, in food ordering domain the resolved restaurant entity can be used as a filter
         to resolve dish entities. The exact knowledge base field to apply these filters depends on
         the knowledge base data model of the application.
+        If the entity is not in the canonical form, a fuzzy filter can be applied by setting the
```
mhuang2173: is there an example of the use case of adding fuzzy filters?

mindmeld/components/question_answerer.py
```diff
@@ -661,34 +668,55 @@ def build_query(self):
...
+                            {
+                                "match": {
+                                    self.field + ".processed_text": {
+                                        "query": self.value
+                                    }
                                 }
                             }
-                        },
-                        {
-                            "match": {
```
mhuang2173: looks like we don't do matching on character ngrams anymore. guess some of the sub-word matching is taken care of by the matching on word stem in `processed_text` field?

In the past i remember we intentionally did not do stopword removal and stemming to avoid losing information and tried to let the data decide if the specific text span is important in the context. (based on TF/IDF score for example)  But guess from our eval looks like we still have performance gain from them for the unstructured data we are experimenting with?

mindmeld/components/question_answerer.py
vembar: In the case of Cisco product documents, the title could be `catalyst 3000` but the query can have `cat 3k`. So if we wanted to filter by matching against titles, we'd need a character n-gram match.

mindmeld/components/question_answerer.py
vembar: So far we've only handled matching against short phrases or sentences. If we have big passages, intuitively, char-ngrams will always match with most passages and do not help in the ranking process. Like you mentioned, in the evaluation set, we see a steady 7% gain in accuracy without them.

mindmeld/components/question_answerer.py
vembar: For filters, the default system uses the normalized words. if the query_type is text, we want to incorporate some fuzziness in the filter as well to give the user flexibility.

minhtuev:
Remember to add documentation on how to use it under `/sources`

vijay120:
LGTM, could you add documentation to this PR as well for completeness?

mindmeld/components/question_answerer.py
```diff
@@ -661,34 +668,55 @@ def build_query(self):
...
-                                self.field + ".char_ngram": {
-                                    "query": self.value
+                        ]
+                    }
+                }
+            else:
```
minhtuev: we should do the check here if the type is keyword, otherwise throw error

mindmeld/components/question_answerer.py
```diff
@@ -797,13 +826,22 @@ def build_query(self):
...
-                                "query": self.value
+                    if self.query_type == 'text':
+                        clause = {
+                            "match": {
+                                self.field + ".char_ngram": {
+                                    "query": self.value
+                                }
+                            }
+                        }
+                    else:
```
minhtuev: check whether the query_type is keyword, otherwise throw UnsupportedQueryTypeException

vembar:
Documentation will be addressed in a different PR since it requires adding a FAQ use-case into the HR blueprint.

PR #44, opened 2019-07-19
Reduce ES requirement for MindMeld

minhtuev:
A couple of things:

1. Catch and process EntityResolverConnectionError; ES should not be a strong requirement for MindMeld application to run.
2. Increase timeout for Travis.
3. Add the correct relative path for icon.

### DIFF
--- a/mindmeld/components/nlp.py
+++ b/mindmeld/components/nlp.py
@@ -1134,7 +1134,7 @@ class EntityProcessor(Processor):
             self.role_classifier.load(incremental_model_path if incremental_timestamp else model_path)
             self.entity_resolver.load()
         except EntityResolverConnectionError:
-            logger.warning('Cannot connect to ES; Entity Resolver is not loaded.')
+            logger.warning('Cannot connect to ES, so Entity Resolver is not loaded.')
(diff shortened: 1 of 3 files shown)


mindmeld/components/nlp.py
```diff
@@ -1128,10 +1128,13 @@ def _dump(self):
...
-            self._app_path, self.domain, self.intent, self.type, timestamp=incremental_timestamp)
-        self.role_classifier.load(incremental_model_path if incremental_timestamp else model_path)
-        self.entity_resolver.load()
+        try:
+            model_path, incremental_model_path = path.get_role_model_paths(
+                self._app_path, self.domain, self.intent, self.type, timestamp=incremental_timestamp)
+            self.role_classifier.load(incremental_model_path if incremental_timestamp else model_path)
+            self.entity_resolver.load()
+        except EntityResolverConnectionError:
+            logger.warning('Cannot connect to ES; Entity Resolver is not loaded.')
```
vijay120: Maybe this: `Cannot connect to ES, so Entity Resolver is not loaded.`

PR #50, opened 2019-08-13
Enable reading the ES configuration from file

minhtuev:
With this change, we will attempt to create ES client basing on:
1) Function params
2) If none, from env variables
3) If none, from config file

The current PR attempts to solve this problem for EntityResolver and QuestionAnswerer. This change should enable MindMeld application to specify ES configuration more easily.


mindmeld/components/entity_resolver.py
```diff
@@ -55,7 +55,14 @@ class EntityResolver:
...
     def __init__(
-        self, app_path, resource_loader, entity_type, es_host=None, es_client=None
+        self,
+        app_path,
+        resource_loader,
+        entity_type,
+        es_host=None,
+        es_username=None,
+        es_password=None,
+        es_client=None,
```
mhuang2173: just wonder is there any reason that we have `es_client` in constructor of Entity Resolver but not Question Answerer? Does the passed-in `es_client` take precedence?

mindmeld/components/entity_resolver.py
```diff
@@ -64,16 +71,43 @@ def __init__(
             resource_loader (ResourceLoader): An object which can load resources for the resolver
             entity_type: The entity type associated with this entity resolver
             es_host (str): The Elasticsearch host server
+            es_username (str): The Elasticsearch username server
```
mhuang2173: typo: remove `server`?

mindmeld/components/entity_resolver.py
```diff
@@ -64,16 +71,43 @@ def __init__(
...
+            es_password (str): The Elasticsearch password server
```
mhuang2173: typo: remove `server`?

vijay120:
@minhtuev Could you address the feedback?

PR #62, opened 2019-09-06
Added stemmers module to the platform

vijay120:
This PR abstracts away the default english stemmer into an abstract Stemming class and implements two NLTK stemmers: a spanish and english one.


mindmeld/stemmers/stemmers.py
```diff
@@ -0,0 +1,73 @@
...
+            return [word]
+        else:
+            return [stem]
+
+
+class SpanishNLTKStemmer(Stemmer):
+
+    def __init__(self, stemmer=None):
+        del stemmer
+        super().__init__(nltk.stem.SnowballStemmer('spanish'))
```
serapio: Let's make language an init arg. Also, how might a passed in stemmer be used?

vijay120: I would like the stemmers to be generic, ie., the developer can implement his/her stemmer given the contract specified in the abstract class `Stemmer`. Therefore, it make sense to have an in-built mindmeld spanish stemmer as it's own class, which the developer can choose to use (or not), rather than a param passed into a Stemming class that is opinionated on how spanish words should be stemmed.

mindmeld/query_factory.py
```diff
@@ -31,10 +29,10 @@ class QueryFactory:
         tokenizer (Tokenizer): the object responsible for normalizing and tokenizing processed
             text
     """
-    def __init__(self, tokenizer, preprocessor=None):
+    def __init__(self, tokenizer, preprocessor=None, stemmer=None):
         self.tokenizer = tokenizer
         self.preprocessor = preprocessor
-        self.stemmer = nltk.stem.PorterStemmer()
+        self.stemmer = stemmer or EnglishNLTKStemmer()
```
jjjacksn: Can we move this default into `create_query_factory()` to be consistent with how the tokenizer default is set?

mindmeld/query_factory.py
```diff
@@ -15,10 +15,8 @@
 
 from __future__ import absolute_import, unicode_literals
 
-import nltk
-
 from . import ser as sys_ent_rec
-
+from .stemmers import EnglishNLTKStemmer
```
jjjacksn: Import order: `.stemmers` should be after `.core`

mindmeld/stemmers.py
```diff
@@ -0,0 +1,81 @@
...
+class Stemmer(ABC):
+
+    def __init__(self):
+        self._stemmer = None
+
+    def set_stemmer(self):
+        """
+        This function sets the underlying stemmer. It is lazily initialized
+        """
+        raise NotImplemented
```
jjjacksn: Should this be abstract?

mindmeld/stemmers.py
```diff
@@ -0,0 +1,81 @@
...
+        if stem == '':
+            return [word]
+        else:
+            return [stem]
+
+
+class SpanishNLTKStemmer(Stemmer):
+
+    def set_stemmer(self):
+        self._stemmer = nltk.stem.SnowballStemmer('spanish')
```
jjjacksn:
```suggestion
        @property
        def _stemmer(self):
            if not hasattr(self, '__stemmer'):
                setattr(self, '__stemmer', nltk.stem.SnowballStemmer('spanish'))
            return getattr(self, '__stemmer')
```

mindmeld/stemmers.py
```diff
@@ -0,0 +1,81 @@
...
+        Returns:
+            list(str): A list of the stemmed version of a word.
+        """
+        raise NotImplemented
+
+
+class EnglishNLTKStemmer(Stemmer):
+
+    def set_stemmer(self):
+        self._stemmer = nltk.stem.PorterStemmer()
```
jjjacksn:
```suggestion
    @property
    def _stemmer(self):
        if not hasattr(self, '__stemmer'):
            setattr(self, '__stemmer', nltk.stem.PorterStemmer())
        return getattr(self, '__stemmer')
```

mindmeld/stemmers.py
jjjacksn: Further -- I think this should just be an abstract property. The implementation can decide whether it's lazy.

mindmeld/query_factory.py
vijay120: resolved

mindmeld/query_factory.py
vijay120: good point, done.

mindmeld/stemmers.py
vijay120: Great suggestion, resolved.

mindmeld/stemmers.py
vijay120: Great suggestion, resolved.

vijay120:
@jjjacksn Many languages have permutations of the stemmers, tokenizers and pre-processors that are all acceptable. 

For example, there are many ways to tokenize Mandarin (https://medium.com/the-artificial-impostor/nlp-four-ways-to-tokenize-chinese-documents-f349eb6ba3c3). 

Stemmers can be opinionated depending on how much "stemming" of the word has to be done. The concern with aggressive stemming is semantically different words will be transformed to the same stem, thus sacrificing NLP accuracy. 

Finally, pre-processors are the most opinionated/app-specific that are operations that need to be done to transform the raw text, for example, "lol" -> "laugh out loud". Such transformations are not handled by stemmers or tokenizers, so this component casts a wider bucket in terms of transformations than the more linguistic-specific transformations of the stemmer and tokenizer.

By allowing developers access to these three components, they are given the flexibility to perform advanced transformations of strings that a more opinionated NLP pipeline would not provide. 

@serapio , thoughts?

mindmeld/stemmers.py
vijay120: Resolved

serapio:
Yeah, in a sense they are all preprocessors. In the English pipeline it has made sense to only use the stemmer in the feature extractor, let the tokenizer process text after the system entity recognizer but before any of the other classifiers, and have the preprocessor as a catch-all for anything we want to do before any other NLP. But it's plausible to me that some app in some language would want to do tokenizing something like what the Spanish stemmer is doing here, and the character tokenization for Chinese (which is what we used previously) is pretty much the same idea. So in a way it might make sense to have tokenizers and stemmers share an interface, but I think it would be premature to try to make that work until we have more experience with different languages.

serapio:
@vembar do you remember having to do anything to remap token indices? it somehow didn't make it into the spanish branch, and without it, the indices are off when it tries to extract the query feature.

PR #71, opened 2019-10-04
Adding explicit error handling in reading query files

minhtuev:
If there is an issue reading a query file, handle it gracefully, log the error and yield an empty list

### DIFF
--- a/mindmeld/markup.py
+++ b/mindmeld/markup.py
@@ -134,7 +134,7 @@ def read_query_file(file_path):
                 if query_text:
                     yield query_text
     except IOError:
-        logger.error('Problem reading file {0}.'.format(file_path))
+        logger.error('Problem reading file %s.', file_path)
         yield from ()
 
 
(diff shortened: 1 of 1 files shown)


mindmeld/markup.py
```diff
@@ -127,13 +127,17 @@ def read_query_file(file_path):
...
-                yield query_text
+    try:
+        with codecs.open(file_path, encoding='utf-8') as queries_file:
+            for line in queries_file:
+                line = line.strip()
+                # only create query if line is not empty string
+                query_text = line.split('\t')[0].strip()
+                if query_text:
+                    yield query_text
+    except IOError:
```
chrisliu156: just wondering, if the file can't be opened or parsed, are we ok with the application to keep running?  It seems like this should be a service ending error, let me know if I'm wrong.

minhtuev: My idea is that we still want to keep the application running even if one file in the training data set is corrupted.

chrisliu156: my main concern is the error might be missed or ignored by the deployer because the error is  not escalated enough because the app is still running and will give a false sense that the app is running correctly.  Don't think the issue is big enough to stop merge though.

Issue #70, opened 2019-10-03
webex_bot_server is not importable


megamit:
The examples provided with the blueprints for example hr_assistant try to import with
`from mindmeld.bot import WebexBotServer`
This fails saying it doesnt exist as a valid __init__.py needs to be put into the bot folder

minhtuev:
Thanks for the find!

PR #73, opened 2019-10-07
Allow WebexBotServer to be imported directly

minhtuev:
Allow WebexBotServer to be imported directly `from mindmeld.bot import WebexBotServer`

Address: https://github.com/cisco/mindmeld/issues/70

### DIFF
--- a/mindmeld/bot/webex_bot_server.py
+++ b/mindmeld/bot/webex_bot_server.py
@@ -7,10 +7,11 @@ This module contains the Webex Bot Server component.
 import logging
 import json
 from flask import request, Flask
-from mindmeld.components import NaturalLanguageProcessor
-from mindmeld.components.dialogue import Conversation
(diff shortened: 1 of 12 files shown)


mindmeld/bot/webex_bot_server.py
```diff
@@ -60,7 +61,7 @@ def __init__(self, name, app_path, nlp=None, webhook_id=None, access_token=None)
         self.access_token_with_bearer = ACCESS_TOKEN_WITH_BEARER + self.access_token
 
         @self.app.route('/', methods=['POST'])
-        def handle_message():
+        def handle_message():  # pylint: disable=unused-variable
```
vijay120: If this function is not being used, we should delete it

minhtuev: It is the main function to handle message

PR #75, opened 2019-10-09
Make markup error messages more useful

serapio:
When reading in query files with bad markup, from some kinds of badness you can get an error that just says where in the query something bad happens, without enough detail to find the bad query. This adds the query to the message.


mindmeld/models/taggers/taggers.py
```diff
@@ -219,7 +219,11 @@ def get_tags_from_entities(query, entities, scheme='IOB'):
             '' if the IOB status is 'O'. The last two are like the first two, \
             but for system entities.
     """
-    iobs, types = _get_tags_from_entities(query, entities, scheme)
+    try:
+        iobs, types = _get_tags_from_entities(query, entities, scheme)
+    except IndexError:
+        logger.error("Invalid entities %s in '%s'", entities, query)
+        raise
```
minhtuev: No exception message here?

PR #77, opened 2019-10-17
Changed function names

vijay120:
Changed function names since they are consumed externally by clients


minhtuev:
Which external clients consume these functions?

PR #78, opened 2019-10-24
Open the application to pass in any server class

minhtuev:
Allow the client to pass in any implementation of abstract MindMeld server into the application


vijay120:
How are clients going to use this? Could you also add documentation that supports this feature?

jjjacksn:
I'm still not 100% convinced we should have server bundled along by default. IIRC we still don't document it well.

In any case, you can already override the CLI to use a custom server (as we do for our own internal apps). Perhaps we should just add illustrative examples for that?

PR #81, opened 2019-11-06
Small update to MindMeld server

minhtuev:
From a previous conversation with @jjjacksn about MindMeld server:

1. Remove some of the server logic that is not used
2. Expose the Flask web app as an attribute

### DIFF
--- a/mindmeld/server.py
+++ b/mindmeld/server.py
@@ -59,21 +59,8 @@ class MindMeldServer:
 
         server.request_class = MindMeldRequest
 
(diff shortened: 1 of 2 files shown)


mindmeld/server.py
```diff
@@ -140,11 +127,9 @@ def handle_server_error(error):
             logger.error(json.dumps(response_data))
             return response
 
-        @server.route("/_status", methods=["GET"])
+        @server.route("/health", methods=["GET"])
```
vijay120: Are we sure no downstream services might be affected by this change? If people were using the `_status` endpoint before, it will no longer work once the lib is updated.

PR #88, opened 2019-12-06
Refactor SystemEntityRecognizer

minhtuev:
~~Add env variable to set duckling url
Default duckling url is: http://localhost:7151/parse~~

After discussing with Lucien and VJ we have decided to refactor SystemEntityRecognizer for a clearer interface. I have split this class into two classes: DucklingRecognizer and DummySystemEntityRecognizer.

The DucklingRecognizer will take in a URL and query against this URL, and return a list of entities and a 200 response code if successful.
DummySystemEntityRecognizer will just return an empty list and a -1 error code.

This change hopefully will create a clearer interface:
1) DucklingRecognizer to be invoked without depending on the application's config.
2) All application-specific logic (such as decoding an entity label) will utilize the system entity recognizer that is configured by the application.

I will update the documentation in a follow-up PR.


mindmeld/components/_config.py
```diff
@@ -508,7 +508,7 @@ def is_duckling_configured(app_path):
 
 def get_system_entity_url_config(app_path):
     if not app_path:
-        return DEFAULT_DUCKLING_URL
+        return os.environ.get('MM_DUCKLING_URL', DEFAULT_DUCKLING_URL)
```
serapio: Is this better than setting it from within the config file? e.g.
```
NLP_CONFIG = {
    'system_entity_recognizer': {'type': 'duckling', 'url': os.environ.get('MM_DUCKLING_URL')}
}

Comment (minhtuev) [reply mindmeld/components/_config.py]: I am thinking of cases where you want to call the numerical parsing service independently from the app

Comment (vijay120) [reply mindmeld/components/_config.py]: But it is called from the app here as well.

Comment (megamit) [reply mindmeld/components/_config.py]: I just looked for how to do this, sad that you cant. Only way to set up mind meld on a local machine using docker compose seems to be with network mode host?

**Comment by minhtuev** [comment]
After discussing with VJ we have decided to refactor SystemEntityRecognizer for a clearer interface. I have split this class into two classes: DucklingRecognizer and DummySystemEntityRecognizer.

The DucklingRecognizer will take in a URL and query against this URL.
DummySystemEntityRecognizer will just return an empty list and a -1 error code.

This change hopefully will create a clearer interface and allow DucklingRecognizer to be invoked without depending on the application's config.

File: mindmeld/ser.py [hunk 2]
```diff
@@ -133,6 +133,7 @@ def parse_numerics(
     locale=None,
     time_zone=None,
     timestamp=None,
+    url=None,
```
vijay120, line 136: You added a URL param here but the caller to this function has not been changed.

mindmeld/system_entity_recognizer.py
```diff
@@ -41,61 +44,101 @@ class SystemEntityError(Exception):
...
+
+        return DummySystemEntityRecognizer._instance
+
+    def get_response(self, data):
+        del data
+        return [], NO_RESPONSE_CODE
+
+
+class DucklingRecognizer(SystemEntityRecognizer):
+    _instances = {}
```
vijay120, line 104: why have this dict?

mindmeld/system_entity_recognizer.py
```diff
@@ -41,61 +44,101 @@ class SystemEntityError(Exception):
...
+            raise Exception("DucklingRecognizer is a singleton")
 
-        self.app_path = app_path
-        SystemEntityRecognizer._instance = self
+        self.url = url
+        DucklingRecognizer._instances[url] = self
 
     @staticmethod
-    def get_instance(app_path=None):
+    def get_instance(url=None):
```
tmehlinger, line 121: Assign the default here to `DEFAULT_DUCKLING_URL` and you don't have to do the conditional assignment below.

mindmeld/components/_config.py
```diff
@@ -468,6 +468,10 @@
 }
 
 
+class NlpConfigException(Exception):
```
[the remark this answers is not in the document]
minhtuev: Ah, I was wondering about whether I should use error or exception.

mindmeld/converter/rasa.py
```diff
@@ -163,19 +163,14 @@ def _read_templates(self):
     def create_entity_files(self, mm_entry):
         entity_value, entity = mm_entry.strip("{}").split("|")
         gazetteer_location = (
-            self.mindmeld_project_directory
-            + "/entities/"
-            + entity
-            + "/gazetteer.txt"
+            self.mindmeld_project_directory + "/entities/" + entity + "/gazetteer.txt"
```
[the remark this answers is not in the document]
minhtuev: done

mindmeld/system_entity_recognizer.py
minhtuev: I am thinking of having a singleton instance per URL. What do you think? We can change to a singleton for all URL as well.

mindmeld/ser.py
minhtuev: Hmm I thought about not exposing the URL to the caller, but you are right. The caller should have the option to expose the URL as well.

mindmeld/system_entity_recognizer.py
minhtuev: The caller of this function can still pass None into get_instance, so I thought it is probably neat to keep the conditional assignment below.

mindmeld/ser.py
```diff
@@ -216,16 +229,17 @@ def parse_numerics(
             # Convert a second grain unix timestamp to millisecond
             timestamp *= 1000
         data["reftime"] = timestamp
-    return SystemEntityRecognizer.get_instance().get_response(data)
+    return DucklingRecognizer.get_instance(url=url).get_response(data)
```
serapio, line 232: Can we do this in a way that doesn't assume the entity recognizer is Duckling? Like if someone has their own special system entity recognizer for Japanese, how would they use this function? Perhaps instead of passing in `url` we could pass in a SystemEntityRecognizer, defaulting to Duckling.

minhtuev: I guess we could, but `parse_numerics` itself (and its immediate callers) is highly specific to Duckling. So one way to do it is to look at the caller of functions such as `get_candidates` and `resolve_system_entity` and see if it makes sense to allow a SystemEntityRecognizer to be passed in there, and if none is passed we need to default to Duckling (with the default URL).

Maybe in a follow-up PR we can explore an actual entity recognizer that is not Duckling (such as spacy). We should note that Duckling is currently doing **both** entity recognition and entity resolution, so whatever alternative that we pick must satisfy both.

tests/components/test_nlp.py
```diff
@@ -617,7 +617,7 @@ def test_sys_entity_feature(kwik_e_mart_nlp):
 
 
 test_data_dyn = [
-    ("kadubeesanahalli", None, "store_info", "", ""),
+    ("kadubeesanahalli", None, "store_info", "get_store_hours", ""),
```
minhtuev, line 620: This test sometime fails when "kadubeesanahalli" is classified as "get_store_hours", so I am pinning a more narrow scope here

mindmeld/markup.py
```diff
@@ -78,7 +78,7 @@ def load_query(
     query_factory = query_factory or QueryFactory.create_query_factory(app_path)
     query_options = query_options or {}
     _, query, entities = process_markup(
-        markup, query_factory=query_factory, query_options=query_options
+        markup, query_factory=query_factory, query_options=query_options,
```
serapio, line 81: super nit: this comma isn't expected by the linter, is it?

tests/test_ser.py
```diff
@@ -10,8 +10,7 @@
 import pytest
```
serapio, line 10: > Tests for `ser` module (Duckling)

=> `system_entity_recognizer`

mindmeld/markup.py
minhtuev: not sure, i think it was probably added by black

tests/test_ser.py
minhtuev: Ah, done

mindmeld/components/entity_recognizer.py
```diff
@@ -206,6 +210,18 @@ def load(self, model_path):
         self.ready = True
         self.dirty = False
 
+    def _predict(self, query, time_zone=None, timestamp=None, dynamic_resource=None):
```
vijay120, line 213: why cant you just override the `predict` function here instead of naming it `_predict `  ?

mindmeld/components/entity_recognizer.py
```diff
@@ -206,6 +210,18 @@ def load(self, model_path):
...
+        if not self._model:
+            logger.error("You must fit or load the model before running predict")
+            return None
+        if not isinstance(query, Query):
+            query = self._resource_loader.query_factory.create_query(
+                query, time_zone=time_zone, timestamp=timestamp
+            )
+        return self._model.predict(
+            [query], dynamic_resource=dynamic_resource, sys_resolver=self.sys_recognizer
+        )[0]
```
vijay120, line 223: why return only the first index?

mindmeld/components/entity_recognizer.py
```diff
@@ -41,17 +42,20 @@ class EntityRecognizer(Classifier):
...
             resource_loader (ResourceLoader): An object which can load resources for the classifier
             domain (str): The domain that this entity recognizer belongs to
             intent (str): The intent that this entity recognizer belongs to
+            sys_recognizer (SystemEntityRecognizer): The system entity recognizer, default to
+                Duckling
         """
         super().__init__(resource_loader)
         self.domain = domain
         self.intent = intent
+        self.sys_recognizer = sys_recognizer or DucklingRecognizer.get_instance()
```
vijay120, line 58: I think `sys_recognizer ` should be renamed to `system_entity_recognizer`

mindmeld/components/entity_recognizer.py
```diff
@@ -206,6 +210,18 @@ def load(self, model_path):
...
+    def _predict(self, query, time_zone=None, timestamp=None, dynamic_resource=None):
+        if not self._model:
+            logger.error("You must fit or load the model before running predict")
+            return None
+        if not isinstance(query, Query):
+            query = self._resource_loader.query_factory.create_query(
+                query, time_zone=time_zone, timestamp=timestamp
+            )
+        return self._model.predict(
+            [query], dynamic_resource=dynamic_resource, sys_resolver=self.sys_recognizer
```
vijay120, line 222: I dont think it makes sense to pass the recognizer to the model for every predict operation. The model should be instantiated with the system entity recognizer during construction and it should use that everytime.

mindmeld/components/entity_recognizer.py
vijay120, line 213: I dont think this function is necessary at all if the `_model` was constructed with the system entity recognizer.

mindmeld/components/entity_recognizer.py
```diff
@@ -258,7 +274,9 @@ def predict_proba(
             query = self._resource_loader.query_factory.create_query(
                 query, time_zone=time_zone, timestamp=timestamp
             )
-        predict_proba_result = self._model.predict_proba([query])
+        predict_proba_result = self._model.predict_proba(
```
vijay120, line 277: Again, we should not be passing in the system entity recognizer everytime. We should be modeling the call patterns of the system entity recognizer similar to the resource loader.

mindmeld/markup.py
```diff
@@ -270,13 +272,18 @@ def process_markup(markup, query_factory, query_options):
...
+    Args:
+        query (Query)
+        annotations (list)
+        sys_resolver (SystemEntityRecognizer): Default to Duckling
 
     Returns:
         list of ProcessedQuery:
     """
     stack = []
+    sys_resolver = sys_resolver or DucklingRecognizer.get_instance()
```
vijay120, line 286: I dont like this pattern of `sys_resolver = sys_resolver or DucklingRecognizer.get_instance()` everywhere since it doesnt isolate the responsibility of one class to instantiate the system_ent_recognizer. All other callers should be expecting a non-None system_ent_recognizer.

mindmeld/models/model.py
```diff
@@ -1286,8 +1287,9 @@ def decode(self, tags_by_example, **kwargs):
         """
         # TODO: support decoding multiple queries at once
         examples = kwargs["examples"]
+        sys_resolver = kwargs.get("sys_resolver") or DucklingRecognizer.get_instance()
```
vijay120, line 1290: sys_resolver should be passed into the model config's constructor instead of hijacking the instance here.

mindmeld/models/tagger_models.py
```diff
@@ -208,11 +208,12 @@ def _convert_params(self, param_grid, y, is_grid=True):
         """
         return param_grid
 
-    def predict(self, examples, dynamic_resource=None):
+    def predict(self, examples, dynamic_resource=None, sys_resolver=None):
```
vijay120, line 211: not needed.

mindmeld/models/tagger_models.py
```diff
@@ -228,16 +229,19 @@ def predict(self, examples, dynamic_resource=None):
...
-            self._label_encoder.decode([example_predicted_tags], examples=[example])[0]
+            self._label_encoder.decode(
+                [example_predicted_tags], examples=[example], sys_resolver=sys_resolver
+            )[0]
             for example_predicted_tags, example in zip(predicted_tags, examples)
         ]
         return labels
 
-    def predict_proba(self, examples, dynamic_resource=None):
+    def predict_proba(self, examples, dynamic_resource=None, sys_resolver=None):
```
vijay120, line 239: not needed

mindmeld/models/taggers/taggers.py
```diff
@@ -277,6 +278,8 @@ def get_entities_from_tags(query, tags):
 
     entities = []
 
+    sys_resolver = sys_resolver or DucklingRecognizer.get_instance()
```
vijay120, line 281: please remove these calls.

mindmeld/components/entity_recognizer.py
minhtuev: Refactored from `ser.py`

mindmeld/query_factory.py
```diff
@@ -30,16 +30,26 @@ class QueryFactory:
...
+        locale=None,
+        language=None,
+        sys_recognizer=None,
     ):
         self.tokenizer = tokenizer
         self.preprocessor = preprocessor
         self.stemmer = stemmer
         self.locale = locale
         self.language = language
+        self.sys_recognizer = sys_recognizer or DucklingRecognizer.get_instance()
```
vijay120, line 52: Again, lets decide on exactly one place where `sys_recognizer = sys_recognizer or DucklingRecognizer.get_instance()` pattern is done and all other users should except `sys_recognizer` and not try to access the singleton.

mindmeld/query_factory.py
```diff
@@ -141,6 +151,15 @@ def create_query_factory(
         language, locale = get_language_config(app_path)
         tokenizer = tokenizer or Tokenizer.create_tokenizer()
         stemmer = stemmer or get_language_stemmer(language_code=language)
+        if app_path:
+            sys_recognizer = SystemEntityRecognizer.get_instance(app_path)
```
vijay120, line 155: here as well.

mindmeld/markup.py
minhtuev: the reason why I am doing this pattern is to make it the interface backward compatible; a non-backward compatible interface will necessitate a major release.

mindmeld/models/taggers/taggers.py
minhtuev: this is needed here since it is an independent function

mindmeld/ser.py
```diff
@@ -12,36 +12,11 @@
...
-    QUANTITY = "quantity"
-    TEMPERATURE = "temperature"
-    VOLUME = "volume"
-    EMAIL = "email"
-    PHONE_NUMBER = "phone-number"
-    URL = "url"
-    TIME = "time"
-
 
 def get_candidates(
```
vijay120, line 21: Why cant `get_candidates ` be a class method of `DucklingRecognizer ` so that we can just access the recognizer through `self` instead of instantiate the singleton again here?

mindmeld/system_entity_recognizer.py
vijay120: isnt there only one URL per `SystemEntityRecognizer ` class? If that is the case, changing it to string or single object makes sense.

mindmeld/system_entity_recognizer.py
vijay120: What does it mean for the caller to pass None when it just uses the `DEFAULT_DUCKLING_URL ` in that case?

mindmeld/markup.py
vijay120: fair point, but `_process_annotations ` is a private function by definition, so its not a public API.

mindmeld/components/entity_recognizer.py
minhtuev: done

mindmeld/components/entity_recognizer.py
minhtuev: done

mindmeld/components/entity_recognizer.py
minhtuev: done

mindmeld/components/entity_recognizer.py
minhtuev: done

mindmeld/ser.py
minhtuev: get_candidates is already refactored into DucklingRecognizer. This is maintained here as a deprecated method.

mindmeld/query_factory.py
minhtuev: `app_path` is an optional argument for `create_query_factory`, so we need a default system entity recognizer.

mindmeld/query_factory.py
minhtuev: We can't because `QueryFactory` can be used independently from the application.

mindmeld/components/nlp.py
```diff
@@ -415,7 +415,7 @@ def __init__(self, app_path, resource_loader=None, config=None):
         self._app_path = app_path
 
         # initialize the system entity recognizer singleton
-        SystemEntityRecognizer.get_instance(app_path)
+        self.system_entity_recognizer = SystemEntityRecognizer.get_instance(app_path)
```
vijay120, line 418: Why is this needed here?

mindmeld/converter/rasa.py
```diff
@@ -162,20 +162,16 @@ def _read_templates(self):
 
     def create_entity_files(self, mm_entry):
         entity_value, entity = mm_entry.strip("{}").split("|")
-        gazetteer_location = (
-            self.mindmeld_project_directory
-            + "/entities/"
-            + entity
-            + "/gazetteer.txt"
+        gazetteer_location = os.path.join(
+            self.mindmeld_project_directory, "entities", entity, "gazetteer.txt"
```
vijay120, line 166: nice

mindmeld/components/nlp.py
```diff
@@ -999,7 +999,8 @@ def __init__(self, app_path, domain, intent, resource_loader=None):
         self.domain = domain
         self.name = intent
 
-        self.entity_recognizer = EntityRecognizer(self.resource_loader, domain, intent)
+        self.entity_recognizer = EntityRecognizer(self.resource_loader, domain, intent,)
```
vijay120, line 1002: comma not needed here.

mindmeld/markup.py
```diff
@@ -260,7 +259,9 @@ def process_markup(markup, query_factory, query_options):
     try:
         raw_text, annotations = _parse_tokens(_tokenize_markup(markup))
         query = query_factory.create_query(raw_text, **query_options)
-        entities = _process_annotations(query, annotations)
+        entities = _process_annotations(
+            query, annotations, sys_resolver=query_factory.sys_recognizer
```
vijay120, line 263: can we rename from `sys_recognizer ` to `system_entity_recognizer ` everywhere? I dont think `sys_recognizer` is clear in it's meaning.

mindmeld/models/taggers/taggers.py
```diff
@@ -269,6 +269,7 @@ def get_entities_from_tags(query, tags):
         query (Query): Any query instance.
         tags (list of str): Joint app and system tags, like those
             created by get_tags_from_entities.
+        sys_resolver (SystemEntityRecognizer): Default to Duckling
```
vijay120, line 272: Can we name this `sys_entity_recognizer`? Right now its named `sys_resolver ` while the component is a recognizer.

mindmeld/query_factory.py
```diff
@@ -141,6 +151,15 @@ def create_query_factory(
...
+        else:
+            sys_recognizer = DucklingRecognizer.get_instance()
```
vijay120, line 157: Shouldnt this be `sys_recognizer = SystemEntityRecognizer.get_instance()`

mindmeld/ser.py
vijay120: Can you annotated these functions as deprecated to that we can remove them in the near future?

mindmeld/system_entity_recognizer.py
```diff
@@ -41,61 +50,194 @@ class SystemEntityError(Exception):
...
+            text (str): The text to examine
+            entity_types (list of str): The entity types to consider
+
+        Returns:
+            list of dict: The system entities found in the text
+        """
+        pass
+
+
+class DummySystemEntityRecognizer(SystemEntityRecognizer):
```
vijay120, line 160: Can we name this is `NoOpSystemEntityRecognizer`? It better reflects the function of this class.

mindmeld/system_entity_recognizer.py
```diff
@@ -41,61 +50,194 @@ class SystemEntityError(Exception):
...
+
+    @staticmethod
+    def get_instance():
+        if not DummySystemEntityRecognizer._instance:
+            DummySystemEntityRecognizer()
+
+        return DummySystemEntityRecognizer._instance
+
+    def parse(self, sentence, **kwargs):
+        return [], NO_RESPONSE_CODE
```
vijay120, line 181: I think the response code should be 200 here.

vijay120: I dont think `NO_RESPONSE_CODE ` is needed.

mindmeld/system_entity_recognizer.py
vijay120: Can we remove this?

mindmeld/system_entity_recognizer.py
minhtuev: The caller might read the URL from some config sources which return None

mindmeld/system_entity_recognizer.py
minhtuev: This is following the existing behavior. We can update this behavior to return 200 instead of -1, but we do need to return the Response Code

vijay120: Sure, I am saying to return 200.

mindmeld/markup.py
minhtuev: done

mindmeld/models/taggers/taggers.py
minhtuev: it was named `sy_resolver` since it was used for entity resolution, not recognition

mindmeld/system_entity_recognizer.py
minhtuev: done

mindmeld/system_entity_recognizer.py
minhtuev: Ok

mindmeld/system_entity_recognizer.py
minhtuev: Done

mindmeld/ser.py
```diff
@@ -63,33 +38,17 @@ def get_candidates(
...
-    language = language or query.language
-    time_zone = time_zone or query.time_zone
-    timestamp = timestamp or query.timestamp
-    response, response_code = parse_numerics(
-        query.text,
-        dimensions=dims,
+    logger.warning(
+        "get_candidates is deprecated in favor of DucklingRecognizer.get_candidates."
+    )
+    DucklingRecognizer.get_instance().get_candidates(
```
vijay120, line 44: Why does this file only use `DucklingRecognizer ` while the other files can use `NoOp`, `Duckling` or anything else?

vijay120: Nmv looks like this is used just for deprecation. Do we have to do `set_system_entity_recognizer` before this call though?

minhtuev: Since the deprecated method uses Duckling, `Duckling.get_instance() `should return a `DucklingRecognizer` even if `SystemEntityRecognizer.get_instance()` returns a `NoOp`

PR #93, opened 2019-12-11
Gracefully handle missing tensorflow extra

jjjacksn:
When the tensorflow extra is not installed users will see an error if they try to use the LstmModel which requires tensorflow:
```
TaggerModel: Classifier type 'lstm' dependencies not found. Install the mindmeld[tensorflow] extra to use this classifier type.
```

Also added a test for this.

### DIFF
--- a/mindmeld/models/tagger_models.py
+++ b/mindmeld/models/tagger_models.py
@@ -28,8 +28,13 @@ from .helpers import (
 )
 from .model import EntityModelEvaluation, EvaluatedExample, Model, ModelConfig
 from .taggers.crf import ConditionalRandomFields
-from .taggers.lstm import LstmModel
 from .taggers.memm import MemmModel
+
+try:
+    from .taggers.lstm import LstmModel
+except ImportError:
+    LstmModel = None
+
 
 logger = logging.getLogger(__name__)
 
@@ -324,6 +329,12 @@ class TaggerModel(Model):
         """Returns the python class of the actual underlying model"""
         classifier_type = self.config.model_settings["classifier_type"]
         try:
+            if classifier_type == LSTM_TYPE and LstmModel is None:
+                msg = (
+                    "{}: Classifier type {!r} dependencies not found. Install the "
+                    "mindmeld[tensorflow] extra to use this classifier type."
+                )
+                raise ValueError(msg.format(self.__class__.__name__, classifier_type))
             return {
                 MEMM_TYPE: MemmModel,
                 CRF_TYPE: ConditionalRandomFields,
--- a/mindmeld/models/taggers/lstm.py
+++ b/mindmeld/models/taggers/lstm.py
@@ -16,9 +16,9 @@ import os
 import re
 
 import numpy as np
-import tensorflow as tf
 from sklearn.externals import joblib
 from sklearn.preprocessing import LabelBinarizer
+import tensorflow as tf
 
 from .embeddings import CharacterSequenceEmbedding, WordSequenceEmbedding
 from .taggers import Tagger, extract_sequence_features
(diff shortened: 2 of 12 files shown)


minhtuev:
Hmm it is not kicking off Travis CI - maybe something is not right with the yaml file.

jjjacksn:
Hmm now I get a lint error related to bot extra not being installed

minhtuev:
Oh that's annoying.

jjjacksn:
Now i'm see a BUNCH of lint errors on the lstm model 🤔

mindmeld/models/taggers/lstm.py
```diff
@@ -210,6 +210,7 @@ def extract_features(self, examples, config, resources, y=None, fit=True):
         Returns:
             (sequence_embeddings, encoded_labels, groups): features for the LSTM network
         """
+        del fit  # unused
```
jjjacksn: @vijay120 is it correct that we aren't using this? Should we make some other change to actually respect the value of fit?

mindmeld/models/taggers/lstm.py
```diff
@@ -447,6 +449,7 @@ def _define_optimizer_and_cost(self):
             log_likelihood, transition_params = tf.contrib.crf.crf_log_likelihood(
                 self.lstm_output_tf, flattened_labels, self.batch_sequence_lengths_tf
             )
+            del transition_params  # unused
```
jjjacksn: @vijay120, again is it correct that we aren't using these params?

vijay120: The issue is with the line above it, do this:
```
            flattened_labels = tf.cast(tf.argmax(self.label_tf, axis=2), tf.int32)
            log_likelihood, _ = tf.contrib.crf.crf_log_likelihood(
                self.lstm_output_tf, flattened_labels, self.batch_sequence_lengths_tf
            )
```

jjjacksn: Yeah, I intentionally didn't use underscore here until it was clear that we shouldn't be using them somewhere else

mindmeld/models/taggers/lstm.py
vijay120: The reason why the `fit` param is not being used here is because in traditional ML models, the fit param is used to scale the input data and featurize it. This is not done in the LSTM model since it directly uses the word embeddings as features without the scaling, hence no such fitting is needed. Hence, the param is not needed.

vijay120: We could technically just do this to solve the issue:
```
        if fit:
            # Train time
            self.resources = resources
            padded_y = self._pad_labels(y, DEFAULT_LABEL)
            y_flat = [item for sublist in padded_y for item in sublist]
            encoded_labels_flat = self.label_encoder.fit_transform(y_flat)
            encoded_labels = []
```

vijay120: So we can solve this issue like this:
```
if fit:
            # Train time
            self.resources = resources
            padded_y = self._pad_labels(y, DEFAULT_LABEL)
            y_flat = [item for sublist in padded_y for item in sublist]
            encoded_labels_flat = self.label_encoder.fit_transform(y_flat)
            encoded_labels = []
```

However, we need to update these lines as well:
Screen Shot 2019-12-12 at 5 14 00 PM

jjjacksn: For now I'll just add a comment explaining that we use the value of y to decide whether to train or not

mindmeld/models/tagger_models.py
```diff
@@ -28,9 +28,14 @@
 )
 from .model import EntityModelEvaluation, EvaluatedExample, Model, ModelConfig
 from .taggers.crf import ConditionalRandomFields
-from .taggers.lstm import LstmModel
 from .taggers.memm import MemmModel
 
+try:
+    from .taggers.lstm import LstmModel
+except ImportError:
+    LstmModel = None
```
vijay120: Could we add a comment here explaining why we are doing this, ie, the tensorflow lib the LSTM model deps on is optional?

PR #95, opened 2019-12-16
Small fixes

minhtuev:
1. Small refactor to convert `get_mm_version` into a public function
2. Small doc updates

### DIFF
--- a/mindmeld/_version.py
+++ b/mindmeld/_version.py
@@ -17,5 +17,5 @@ from __future__ import absolute_import, unicode_literals
 current = "4.2.3"
 
 
-def _get_mm_version():
+def get_mm_version():
     return current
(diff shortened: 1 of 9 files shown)


mindmeld/_version.py
```diff
@@ -17,5 +17,5 @@
 current = "4.2.3"
 
 
-def _get_mm_version():
+def get_mm_version():
```
serapio: Should have a docstring now that it's public

mindmeld/ser.py
```diff
@@ -135,6 +135,7 @@ def parse_numerics(
     timestamp=None,
 ):
     """Calls System Entity Recognizer service API to extract numerical entities from a sentence.
+
     Args:
         sentence (str): A raw sentence.
         dimensions (None or list of str): The list of types (e.g. volume, \
```
serapio: on the next line, let's add a period at the end.

minhtuev:
Build failures are not related to this PR

PR #98, opened 2020-01-03
Added language and locale params to NLP config

vijay120:
## What does this PR do?
This PR adds language and locale as configuration constants to a MindMeld application, so setting these values will make sure system entities are detecting during train and inspect time. For inference, the developer can still choose to set the locale and language, but if not, the application's language configuration is used. It's useful for the developer to choose the language config during inference time if they want to resolve certain entities differently (ie `11/12/2010` mean different dates in `en_GB` and `en_US`).

## Proposed design change
The following config can be added in the `_config.py` file:
```
LANGUAGE_CONFIG = {
    'language': 'en',
    'locale': 'en_CA'
}
```
The PR works for all NLP apis.

## Limitations
1. The PR currently only supports the abbreviated version of locale codes, ie `en_US` and not the extended version ie `en_XX_US`.
2. This change would mean each Mindmeld application is only trained on one language


tests/components/test_dialogue.py
```diff
@@ -291,8 +291,10 @@ def test_convo_params_are_cleared(kwik_e_mart_nlp, kwik_e_mart_app_path):
     [
         ("en", "en_GB", {"lang": "EN", "latent": True, "locale": "en_GB"}),
         ("es", "en_US", {"latent": True, "locale": "en_US"}),
-        (None, None, {"latent": True, "locale": "en_US", "lang": "EN"}),
+        (None, None, {"latent": True, "locale": "en_CA", "lang": "EN"}),
```
minhtuev: why change?

tests/components/test_nlp.py
```diff
@@ -68,6 +69,38 @@ def test_process(kwik_e_mart_nlp):
...
+    assert query.language == 'en'
+    assert query.locale == 'en_CA'
+
+
+def test_process_contains_language_information(kwik_e_mart_nlp):
+    # Timestamp is set to Jan 2nd 2020 so that the resolve thanksgiving date will be in 2020
+    result = kwik_e_mart_nlp.process('Is the Main Street location open for Thanksgiving?',
+                                     timestamp=1578025558)
+    for entity in result['entities']:
+        if entity['text'] == 'Thanksgiving':
```
minhtuev: will this test fail next year (2021)?

tests/components/test_dialogue.py
vijay120: I have updated the kwik-e-mart test application to use the `en_CA` locale

tests/components/test_nlp.py
vijay120: No it wont since I have set the timestamp to the year 2020 (line 86)

mindmeld/query_factory.py
```diff
@@ -130,7 +137,8 @@ def create_query_factory(
         Returns:
             QueryFactory: A QueryFactory object that is used to create Query objects.
         """
-        del app_path
         tokenizer = tokenizer or Tokenizer.create_tokenizer()
         stemmer = stemmer or EnglishNLTKStemmer()
```
minhtuev: Do we pick the stemmer based on the language & locale? Should we continue to use the English Stemmer for non-English languages? @serapio

serapio: This does seem weird now. It should only default to the English stemmer if the language is English, and otherwise the default stemmer should be a dummy (no op) stemmer. We should probably also make the default tokenizer less opinionated (in future work). For other languages, the app should generally be defining it's own tokenizer and stemmer, but we shouldn't use the English stemmer if they fail to do so.

tests/components/test_nlp.py
```diff
@@ -68,6 +69,38 @@ def test_process(kwik_e_mart_nlp):
...
+            # American thanksgiving is 2020-11-26T00:00:00.000-08:00,
+            # which is different from Canada's.
+            assert entity['value'][0]['value'] == '2020-10-12T00:00:00.000-07:00'
```
ktick: Clever way to test whether duckling indeed works correctly with en-CA 👍

mindmeld/query_factory.py
ktick: @serapio What are your thoughts on using the Snowball stemmer?

```
In [1]: from nltk.stem.snowball import SnowballStemmer

In [2]: SnowballStemmer.languages
Out[2]: 
('arabic',
 'danish',
 'dutch',
 'english',
 'finnish',
 'french',
 'german',
 'hungarian',
 'italian',
 'norwegian',
 'porter',
 'portuguese',
 'romanian',
 'russian',
 'spanish',
 'swedish')
```

mindmeld/stemmers.py
```diff
@@ -52,15 +62,43 @@ def stem_word(self, word):
...
+        return EnglishNLTKStemmer()
+
+    language = None
+    if len(language_code) == 2:
+        language = pycountry.languages.get(alpha_2=language_code.upper())
+    elif len(language_code) == 3:
+        language = pycountry.languages.get(alpha_3=language_code.upper())
+
+    if not language:
+        logger.error('Language code {} is not supported for stemming. '
```
minhtuev: Maybe this should be a warning

mindmeld/query_factory.py
vijay120: Implemented the above suggestion

mindmeld/stemmers.py
vijay120: done

mindmeld/stemmers.py
```diff
@@ -52,15 +62,56 @@ def stem_word(self, word):
...
+        return EnglishNLTKStemmer()
+
+    language = None
+    if len(language_code) == 2:
+        language = pycountry.languages.get(alpha_2=language_code)
+    elif len(language_code) == 3:
+        language = pycountry.languages.get(alpha_3=language_code)
+
+    if not language:
+        logger.warning('Language code %s is not supported for stemming. '
```
serapio: `Language code "%" is ...`

mindmeld/stemmers.py
```diff
@@ -52,15 +62,56 @@ def stem_word(self, word):
...
+                       'Consider disabling the stemmer in config.py.', language_code)
+        return NoOpStemmer()
+
+    language_name = language.name.lower()
+    if language_name in nltk.stem.SnowballStemmer.languages:
+        return SnowballNLTKStemmer(language_name)
+
+    logger.warning('Language code %s is not supported for stemming. '
```
serapio: `Language code "%s" is not supported...`

mindmeld/stemmers.py
```diff
@@ -52,15 +62,56 @@ def stem_word(self, word):
...
+                   'Consider disabling the stemmer in config.py.', language_code)
+    return NoOpStemmer()
```
serapio: This is great. Is disabling the stemmer a different result than getting the NoOpStemmer?

mindmeld/stemmers.py
vijay120: done

mindmeld/stemmers.py
vijay120: done

mindmeld/stemmers.py
vijay120: This is a good point. So Mindmeld stems all queries but uses the stemmed word as a feature only when stemming is enabled. I will make this clear in the warning message.

source/userguide/internationalization.rst
```diff
@@ -0,0 +1,87 @@
...
+5. French
+6. German
+7. Hungarian
+8. Italian
+9. Norwegian
+10. Portuguese
+11. Romanian
+12. Russian
+13. Spanish
+14. Swedish
```
serapio: I think we should make the messaging here more nuanced. This makes it sound like only these languages will work, but these are just the languages where stemming works well, which is not important for many languages. e.g. the Vietnamese app seems to work pretty well and Vietnamese wouldn't benefit from stemming. I see it could be more complicated to the user but it would be more accurate to break it down by resource, something like:
* The current tokenizer works well for languages like English where words are delimited by spaces, but may be confused by punctuation marks not found in English. If the language you are dealing with does not use spaces between words or has many non-English-like punctuation marks, you should preprocess your data to remove punctuation and add spaces between words.
 * The stemmer supports the following languages [listed above]
If you are working with a language that is not supported by the stemmer, we recommend disabling the stemmer, increasing the use of character ngram features, and expect you will need more training data.
* The default system entity recognizer supports the following languages [listed]
If you are working with a language that is not supported by the default system entity recognizer, we recommend using custom entities and resolving the entities with custom logic in the dialogue manager.

vijay120: Addressed your comment

source/userguide/internationalization.rst
```diff
@@ -0,0 +1,97 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+For :ref:`system entity resolution <system-entities>`, the following ISO 639-1 language codes are currently supported: [AR, BG, BN, CS, DA,
+DE, EL, EN, ES, ET, FI, FR, GA, HE, HI, HR, HU, ID, IS, IT, JA, KA, KN, KM, KO, LO, ML, MN, MY, NB, NE, NL, PL, PT, RO,
+RU, SV, SW, TA, TH, TR, UK, VI, ZH].
+
+Moreover, the following ISO 3166-2 locale codes are supported per language:
+
+1. EN: [AU, BZ, CA, GB, IN, IE, JM, NZ, PH, ZA, TT, US]
+2. NL: [BE, R.NL]
```
serapio: `NL: [BE, NL]`

source/userguide/internationalization.rst
minhtuev: I feel that apart from this PR we should do an evaluation of the effectiveness of these stemmers to provide the user with a reference of when/why they should be using stemmers.

source/userguide/internationalization.rst
```diff
@@ -0,0 +1,97 @@
...
+        'locale': 'en_US'
+    }
+
+
+Language stemming
+^^^^^^^^^^^^^^^^^
+
+Stemming is an important, language-dependent NLP process that transforms a word to an approximation of its root form. Stemming can be
+useful for some languages like English but not for others like Vietnamese. Mindmeld supports the following ISO 639-1 language codes for stemming:
+[EN, DA, NL, AR, FR, DE, HU, IT, NO, PT, RU, RO, ES, SV, FI].
```
minhtuev: Should we expand the actual names of the languages instead of the codes?

source/userguide/internationalization.rst
```diff
@@ -0,0 +1,97 @@
+Internationalization support
+============================
+
+Mindmeld supports most languages that can be tokenized like English. If the language does not use spaces between
+words or has many non-English-like punctuation marks, pre-process the data to remove punctuations and add spaces between words.
```
minhtuev: @serapio : can we give an example of the languages that have many of these punctuation marks?

source/userguide/internationalization.rst
vijay120: resolved

source/userguide/internationalization.rst
serapio: This is something I had heard was a problem generally with using \W regex across languages, but the ones I tried in German, Arabic and Chinese were all handled correctly.

Issue #106, opened 2020-02-03
Application fails because there is no config file


minhtuev:
Currently, the app requires config.py file even if the user does not want to tune the configuration

minhtuev:
PR is merged

PR #108, opened 2020-02-07
Added try except block for language config

vijay120:
This PR solves https://github.com/cisco/mindmeld/issues/106 by catching invalid path configs.

### DIFF
--- a/mindmeld/components/_config.py
+++ b/mindmeld/components/_config.py
@@ -466,17 +466,21 @@ DEFAULT_NLP_CONFIG = {
 
 
 def get_language_config(app_path):
+
     if not app_path:
         return ENGLISH_LANGUAGE_CODE, ENGLISH_US_LOCALE
 
-    language_config = getattr(_get_config_module(app_path),
-                              'LANGUAGE_CONFIG',
-                              DEFAULT_LANGUAGE_CONFIG)
-
-    locale = language_config.get('locale')
-    language = language_config.get('language')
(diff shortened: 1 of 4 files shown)


mindmeld/components/_config.py
```diff
@@ -466,18 +466,18 @@
 
 
 def get_language_config(app_path):
-    if not app_path:
```
minhtuev: i think we should keep this check, it is easy to follow

tests/conftest.py
```diff
@@ -36,9 +36,6 @@
 )
 AENEID_FILE = "aeneid.txt"
 AENEID_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)), AENEID_FILE)
-HOME_ASSISTANT_APP_PATH = os.path.join(
```
minhtuev: i assume we are removing this because it is not used?

tests/test_path.py
```diff
@@ -58,3 +60,20 @@ def test_get_indexes():
...
+    "path, language, locale",
+    [
+        (APP_PATH, "en", "en_CA"),
+        ("{}/../tests/{}".format(
+            os.path.dirname(
+                os.path.abspath(__file__)),
+            APP_NAME),
+         "en", "en_CA"),
+        (".", "en", "en_US"),
+        (None, "en", "en_US"),
```
minhtuev: should add a test for invalid app path as well

PR #113, opened 2020-02-11
Feature/mindmeld converter

vijay120:
This PR changes the DF converter in the following ways:
1. Renames generated MM intents and entities without extraneous tokens like `entries` and `user_says`.
2. Adds DF converter test
3. Generates two commands, "app_specific" and "unrelated". All unspecified intents go into the "unrelated" domain.
4. Adds a new config.py file that contains parameters
5. Replaces the existing DF sample project with a Banking blueprint from DF.


tests/test_converter.py
```diff
@@ -0,0 +1,10 @@
+from mindmeld.test import TestConversation
```
minhtuev: move this file to https://github.com/cisco/mindmeld/blob/master/tests/converter/d

tests/conftest.py
```diff
@@ -42,9 +43,16 @@
...
 )
-DIALOGFLOW_PROJECT_PATH = os.path.join(
+DF_PROJECT_PATH = os.path.join(
     os.path.dirname(os.path.abspath(__file__)), "converter/dialogflow_sample_project"
 )
+# This is the dialogflow app converted to mindmeld app
+MM_DF_PATH = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)), "converter/mm_df_converted_project"
+)
+df_init = DialogflowConverter(DF_PROJECT_PATH, MM_DF_PATH)
```
minhtuev: this should be inside a test

mindmeld/converter/dialogflow.py
```diff
@@ -182,22 +186,24 @@ def _create_intents_directories(self, intents):
                 mindmeld_intent_directory_name = self.clean_check(
                     sub, self.intents_list
                 )
+
+                domain = 'unrelated' if \
+                    'default' in mindmeld_intent_directory_name else 'app_specific'
```
serapio: if it has to split the line, we should just use a regular `if` block

mindmeld/converter/dialogflow.py
```diff
@@ -225,12 +232,12 @@ def _create_intent_file(
...
                                 df_meta[1:],
                             )
+                            entity_type = self.clean_name(mm_meta) + "_" + language
 
-                        entity_type = self.clean_name(mm_meta) + "_entries_" + language
                         part = "{" + df_text + "|" + entity_type + "}"
                     else:
                         entity_type = (
-                            self.clean_name(df_meta[1:]) + "_entries_" + language
+                            self.clean_name(df_meta[1:]) + "_" + language
```
serapio: are you guaranteed that `language` is a string?

mindmeld/converter/dialogflow.py
```diff
@@ -379,26 +386,15 @@ def create_mindmeld_init(self):
...
-                                    ]
+                                function_name = "renameMe" + str(i) + "_" + language
+                                handles = [
+                                    "intent="
+                                    + "'"
+                                    + self.clean_name(datastore["name"])
+                                    + "_"
+                                    + language
+                                    + "'"
+                                ]
```
serapio: f-string please

tests/conftest.py
```diff
@@ -96,8 +104,17 @@ def mindmeld_project_path2():
...
+    return MM_DF_PATH
+
+
+@pytest.fixture(scope="session")
+def mm_df_nlp(mm_df_app_path):
+    """Provides a built processor instance"""
+    nlp = NaturalLanguageProcessor(app_path=mm_df_app_path)
+    nlp.build()
+    nlp.dump()
+    return nlp
```
serapio: do these files get cleaned up after the test?

tests/converter/dialogflow_sample_project/intents/account.spending.check - context_ spending - date.json
```diff
@@ -0,0 +1,71 @@
+{
```
serapio: Are the spaces in the file name intentional?

mindmeld/converter/dialogflow.py
vijay120: done

mindmeld/converter/dialogflow.py
vijay120: Resolved this issue by making sure the function calling this function always has the language string set.

mindmeld/converter/dialogflow.py
vijay120: done

tests/test_converter.py
vijay120: done

mindmeld/converter/generic_config.txt
```diff
@@ -22,10 +22,52 @@ ENTITY_RECOGNIZER_CONFIG = {
...
+        },
+    },
+    "features": {
+        "bag-of-words": {"lengths": [1, 2, 3]},
+        "freq": {"bins": 5},
+        "enable-stemming": True,
+        "sys-candidates": {},
+    },
+}
+
```
arushir: How was this config selected and why is it included? If it's meant to be the best default parameters for all new applications, we should update the default configs in the MindMeld platform instead

tests/conftest.py
vijay120: done

tests/conftest.py
vijay120: done

tests/converter/dialogflow_sample_project/intents/account.spending.check - context_ spending - date.json
vijay120: Yes, since this is the file name that is generated from DF

mindmeld/converter/generic_config.txt
vijay120: removed this file

mindmeld/converter/converter.py
```diff
@@ -62,20 +62,6 @@ def create_directory(directory):
             except OSError:
                 logger.error("Cannot create directory at %s", directory)
 
-    @staticmethod
```
minhtuev: We should keep this function since it is still being used by the Rasa converter.

mindmeld/converter/dialogflow.py
```diff
@@ -182,22 +186,27 @@ def _create_intents_directories(self, intents):
                 mindmeld_intent_directory_name = self.clean_check(
                     sub, self.intents_list
                 )
+
+                if 'default' in mindmeld_intent_directory_name:
```
minhtuev: we should add documentation here

tests/converter/dialogflow_sample_project/stream.py
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env python
```
minhtuev: Is this part of the DF flow sample project?

tests/converter/test_converter.py
```diff
@@ -0,0 +1,38 @@
+import os
```
minhtuev: we should rename this file to `test_df_converter`

tests/converter/test_converter.py
```diff
@@ -0,0 +1,38 @@
...
+    mm_df_app = importlib.import_module('mm_df_converted_project').app
+    mm_df_app.lazy_init(mm_df_nlp)
+
+    conv = TestConversation(app=mm_df_app)
+    conv.process("what is my balance")
+    conv.assert_text("Here's your latest balance:")
+    conv.assert_domain("app_specific")
+    conv.assert_intent("accountbalancecheck_en")
+    conv.assert_frame({})
+
```
minhtuev: we should also check in the nlp objects that:
- all domains, intents, entities are converted correctly
- the default domain is converted correctly

mindmeld/converter/test.py
```diff
@@ -0,0 +1,7 @@
+from mindmeld.converter.dialogflow import DialogflowConverter
```
minhtuev: we should consolidate this into `test_df_converter` to make sure that this function is always invoked before we can load and test the NLP app

mindmeld/converter/generic_config.txt
```diff
@@ -1,31 +0,0 @@
-ENTITY_RECOGNIZER_CONFIG = {
```
minhtuev: we should keep this since the rasa converter is also using this

mindmeld/converter/dialogflow.py
vijay120: done

tests/converter/dialogflow_sample_project/stream.py
vijay120: good point, deleted.

tests/converter/test_converter.py
vijay120: done

mindmeld/converter/dialogflow.py
minhtuev: i don't think we can use fstring here since it is only available on python 3.6 and above while mindmeld still supports py3.5

tests/converter/test_converter.py
minhtuev: to clarify, for #3 we don't need to check all responses, but add a few more to make sure that the edge cases are addressed.

vijay120: done with the first three points, point 4 is not supported in this PR

mindmeld/converter/generic_config.txt
vijay120: We should be using the default settings. The only difference between this and the default is stemming being enabled, which is not necessary for RASA or DF.

mindmeld/converter/test.py
vijay120: Deleted this file since the sample app has changed.

minhtuev: should this file be deleted?

arushir:
One more thought - since we are working on a MindMeld banking blueprint, maybe we could use the DF Hotel Booking or Car example applications instead? Just to give more diversity in examples.

PR #115, opened 2020-02-11
Upgraded flask

vijay120:
This PR upgrades the Flask version.


tests/test_request.py
```diff
@@ -5,63 +5,63 @@
 
 
 @pytest.fixture
-def request():
+def test_request():
```
minhtuev: I think we should avoid naming fixtures with a `test` prefix since that's how tests are named.
```suggestion
def sample_request():
```

vijay120: done

PR #116, opened 2020-02-19
Implemented slot filling code generation logic

vijay120:
This PR implements slot filling and slot re-prompting logic for the DF converter

### DIFF
--- a/mindmeld/converter/code_generator.py
+++ b/mindmeld/converter/code_generator.py
@@ -1,0 +1,109 @@
+class CodeGenerator:
+    """
+    This class contains primitive python functions to generate
+    python code
+    """
+
+    def begin(self, tab="\t"):
+        self.code = []
+        self.tab = tab
+        self.level = 0
+
+    def end(self):
+        return "".join(self.code)
+
+    def write(self, string, new_lines=1):
+        self.code.append(self.tab * self.level + string + "\n" * new_lines)
+
+    def indent(self):
+        self.level = self.level + 1
+
+    def dedent(self):
+        if self.level == 0:
+            raise SyntaxError("internal error in code generator")
+        self.level = self.level - 1
+
+
+class MindmeldCodeGenerator(CodeGenerator):
+    """
+    This class generates Mindmeld specific python code blocks
+    """
+
+    def generate_handle(self, params):
+        self.write("@app.handle(%s)" % params)
+
+    def generate_header(self, function_name):
+        self.write("def %s(request, responder):" % function_name)
+
+    def generate_function(self, handle, function_name, replies):
+        self.generate_handle(handle)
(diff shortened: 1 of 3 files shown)


mindmeld/converter/code_generator.py
```diff
@@ -0,0 +1,100 @@
+class CodeGenerator:
```
minhtuev: docstring?

tests/converter/test_df_converter.py
```diff
@@ -85,5 +85,14 @@ def test_df_converter():
     conv.assert_intent("accountspendingcheck_en")
     conv.assert_frame({})
 
+    conv.process("transfer money")
+    conv.assert_text("Sure. Transfer from which account?")
+
+    conv.process("checking account")
+    conv.assert_text("To which account?")
+
+    conv.process("transfer to savings account")
```
minhtuev: nice! does the follow up logic work from here when you specify the amount that you want to transfer?

vijay120: No, since there is not enough NLP data for amount role classification. We will not fix the NLP issue in this PR.

mindmeld/converter/code_generator.py
vijay120: done

mindmeld/converter/dialogflow.py
```diff
@@ -360,16 +343,9 @@ def create_mindmeld_init(self):
...
-                "# -*- coding: utf-8 -*-",
-                '"""This module contains the MindMeld application"""',
-                "from mindmeld import Application",
-                "app = Application(__name__)",
-                "__all__ = ['app']",
-            ]
 
-            for info, spacing in zip(begin_info, [1, 2, 1, 1, 0]):
-                target.write(info + "\n" * spacing)
+            self.code_gen.begin(tab="    ")
```
minhtuev: nice!

tests/converter/test_df_converter.py
vijay120: Fixed this issue, looked like it was a bug in the annotation generator.

mindmeld/converter/dialogflow.py
```diff
@@ -385,31 +360,74 @@ def create_mindmeld_init(self):
...
-                                    % (self.clean_name(datastore["name"]), language)
-                                ]
-
-                                target.write(
-                                    "\n\n\n"
-                                    + self.create_function(
-                                        handles=handles,
-                                        function_name=function_name,
-                                        replies=replies,
+                        message = response["messages"][0]
```
minhtuev: consider refactoring this code into a separate function/class? MessageConverter?

vijay120: done

PR #117, opened 2020-02-20
Augmented data

vijay120:
This PR does the following:
1. Removes the test set so that all the limited data can be used for training and evaluation is done through cross-validation.
2. If the training set size is less than 10, the data is duplicated.
3. For the unrelated domain, negative training data is added.

### DIFF
--- a/mindmeld/converter/dialogflow.py
+++ b/mindmeld/converter/dialogflow.py
@@ -22,6 +22,7 @@ import re
 from sklearn.model_selection import train_test_split
 
 from mindmeld.converter.converter import Converter
+from mindmeld.converter.code_generator import MindmeldCodeGenerator
 
 logger = logging.getLogger(__name__)
 
@@ -97,6 +98,7 @@ class DialogflowConverter(Converter):
             self.directory = os.path.dirname(os.path.realpath(__file__))
             self.entities_list = set()
             self.intents_list = set()
+            self.code_gen = MindmeldCodeGenerator()
         else:
             msg = "`{dialogflow_project_directory}` does not exist. Please verify."
             msg = msg.format(dialogflow_project_directory=dialogflow_project_directory)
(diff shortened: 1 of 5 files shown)


tests/converter/test_df_converter.py
```diff
@@ -93,5 +93,25 @@ def test_df_converter():
         "All right. So, you're transferring $200 from your checking to a savings. Is that right?"
     )
 
+    conv.process("hello!")
+    conv.assert_text(
```
minhtuev: is it easier to assert domain/intent here?

vijay120: I think its fine to assert the whole NLR since we want to test if the output behavior is consistent for a given DF app

mindmeld/converter/dialogflow.py
```diff
@@ -259,13 +260,20 @@ def _create_intent_file(
...
+            if key in mindmeld_intent_directory:
+                with open(os.path.join(package_dir, default_intent_to_training_file[key])) as fp:
+                    for line in fp:
+                        all_text.append(line.strip())
 
-        target_test.write("\n".join(test))
-        target_train.write("\n".join(train))
+        # Double the size of the training set if there are less than 10 training examples.
+        # This is needed since the k-fold cross validation parameter is set to 10 for
+        # intent classification.
```
serapio: We could reasonably set `k` to 3 or 5 for this converter. `k=10` is good for when the training set N>50 or so, but using the minimum k is better than duplicating data. If the data is still lower than k, then yeah we need to duplicate.

mindmeld/converter/dialogflow.py
```diff
@@ -259,13 +260,20 @@ def _create_intent_file(
                 sentence += part
             all_text.append(sentence)
 
-        train, test = train_test_split(all_text, test_size=0.2)
+        for key in default_intent_to_training_file:
+            if key in mindmeld_intent_directory:
+                with open(os.path.join(package_dir, default_intent_to_training_file[key])) as fp:
+                    for line in fp:
+                        all_text.append(line.strip())
```
serapio: Why do we combine welcome and unrelated data?

vijay120: We are not combining the data from the two intents.

mindmeld/converter/dialogflow.py
vijay120: Understood, I will address this in a followup PR.

Issue #119, opened 2020-02-20
Make duckling request configurable


vijay120:
Currently, the request to duckling is set to 1sec. This often times outs on long queries: https://github.com/cisco/mindmeld/blob/master/mindmeld/system_entity_recognizer.py#L86

PR #120, opened 2020-02-21
Added request timeout configuration

vijay120:
This PR makes the request call to the system entity recognition service to have a configurable timeout. Refer to https://github.com/cisco/mindmeld/issues/119.

### DIFF
--- a/mindmeld/system_entity_recognizer.py
+++ b/mindmeld/system_entity_recognizer.py
@@ -13,15 +13,22 @@
 import json
 import logging
 import sys
-
+import os
 import requests
 
 from mindmeld.components._config import (
     get_system_entity_url_config,
     is_duckling_configured,
 )
+from mindmeld.exceptions import MindMeldError
 
 NO_RESPONSE_CODE = -1
+SYS_ENTITY_REQUEST_TIMEOUT = os.environ.get("MM_SYS_ENTITY_REQUEST_TIMEOUT", 1.0)
+try:
+    if float(SYS_ENTITY_REQUEST_TIMEOUT) <= 0.0:
+        raise MindMeldError("MM_SYS_ENTITY_REQUEST_TIMEOUT env var has to be > 0.0 seconds.")
+except ValueError:
+    raise MindMeldError("MM_SYS_ENTITY_REQUEST_TIMEOUT env var has to be a float value.")
 
 logger = logging.getLogger(__name__)
 
@@ -83,7 +90,7 @@ class SystemEntityRecognizer:
         url = get_system_entity_url_config(app_path=self.app_path)
 
         try:
-            response = requests.request("POST", url, data=data, timeout=1)
+            response = requests.request("POST", url, data=data, timeout=SYS_ENTITY_REQUEST_TIMEOUT)
 
             if response.status_code == requests.codes["ok"]:
                 response_json = response.json()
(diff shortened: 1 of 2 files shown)


mindmeld/system_entity_recognizer.py
```diff
@@ -22,6 +22,7 @@
 )
 
 NO_RESPONSE_CODE = -1
+SYS_ENTITY_REQUEST_TIMEOUT = float(os.environ.get("MM_SYS_ENTITY_REQUEST_TIMEOUT", 1.0))
```
davidmfrey: We should probably enforce a > 0.0 value?

davidmfrey: For reference, here's what happens when you throw a negative timeout into requests
`ValueError: Attempted to set connect timeout to -1.0, but the timeout cannot be set to a value less than or equal to 0.
`
It's better to warn the user if we see an error rather than defer it to later when we call duckling.

vijay120: Addressed this issue.

jre21: If the environment variable doesn't parse as a float, do we want to catch the resulting `ValueError` and reraise it as a `MindMeldError`?

minhtuev: Probably we need to do some validation here.

mindmeld/system_entity_recognizer.py
```diff
@@ -13,15 +13,19 @@
...
 from mindmeld.components._config import (
     get_system_entity_url_config,
     is_duckling_configured,
 )
+from mindmeld.exceptions import MindMeldError
 
 NO_RESPONSE_CODE = -1
+SYS_ENTITY_REQUEST_TIMEOUT = float(os.environ.get("MM_SYS_ENTITY_REQUEST_TIMEOUT", 1.0))
+if SYS_ENTITY_REQUEST_TIMEOUT <= 0.0:
+    raise MindMeldError("SYS_ENTITY_REQUEST_TIMEOUT value has to be greater than 0.0 seconds")
```
minhtuev:
```suggestion
    raise MindMeldError("SYS_ENTITY_REQUEST_TIMEOUT value has to be greater than 0 seconds")
```

mindmeld/system_entity_recognizer.py
```diff
@@ -83,7 +87,7 @@ def get_response(self, data):
         url = get_system_entity_url_config(app_path=self.app_path)
```
minhtuev: ``def get_response(self, data, timeout=None): 
        if not timeout:
            timeout = SYS_ENTITY_REQUEST_TIMEOUT
``

mindmeld/system_entity_recognizer.py
davidmfrey: I think representing the error as a double is more clear to the end user

mindmeld/system_entity_recognizer.py
vijay120: resolved

mindmeld/system_entity_recognizer.py
vijay120: I am unclear why this needs to be done.

mindmeld/system_entity_recognizer.py
```diff
@@ -13,15 +13,22 @@
...
 )
+from mindmeld.exceptions import MindMeldError
 
 NO_RESPONSE_CODE = -1
+SYS_ENTITY_REQUEST_TIMEOUT = os.environ.get("MM_SYS_ENTITY_REQUEST_TIMEOUT", 1.0)
+try:
+    if float(SYS_ENTITY_REQUEST_TIMEOUT) <= 0.0:
+        raise MindMeldError("MM_SYS_ENTITY_REQUEST_TIMEOUT env var has to be > 0.0 seconds.")
+except ValueError:
+    raise MindMeldError("MM_SYS_ENTITY_REQUEST_TIMEOUT env var has to be a float value.")
```
jre21: Can we change this to the following?
```
except ValueError as e:
    raise MindMeldError(...) from e
```
It's a subtle distinction, but it makes the error handling a bit less confusing because it causes python to treat the later exception as a result of the earlier one, rather than an unrelated bug that happened to trigger in the error-handling code.

PR #121, opened 2020-02-21
Made the number of queries for df conversion configurable

vijay120:
The PR makes the minimum number of queries for the DF to MM conversion script configurable.


mindmeld/converter/dialogflow.py
```diff
@@ -25,6 +25,8 @@
 logger = logging.getLogger(__name__)
 package_dir = os.path.dirname(os.path.abspath(__file__))
 
+MIN_NUM_QUERIES = int(os.environ.get("MM_DF_CONVERTER_MIN_NUM_QUERIES", 5))
```
serapio: I think what we actually want is an env variable that links the number of folds in the generic config file to the number of queries here

minhtuev: Can you elaborate?

vijay120: resolved this issue

PR #124, opened 2020-02-27
Remove Language param from process

vijay120:
This PR does the following:
1. Removes the use of Language param (but does not deprecate) from the process apis since this cannot be changed for each Mindmeld application
2. Adds app-level language and locale fields to the Processor class so that we don't check against the config.py file every-time we want to read the language/locale for a query.

### DIFF
--- a/mindmeld/components/nlp.py
+++ b/mindmeld/components/nlp.py
@@ -392,6 +392,7 @@ class NaturalLanguageProcessor(Processor):
         self.name = app_path
         self._load_custom_features()
         self.domain_classifier = DomainClassifier(self.resource_loader)
+        self.app_language, self.app_locale = get_language_config(app_path)
 
         for domain in path.get_domains(self._app_path):
             self._children[domain] = DomainProcessor(
@@ -692,7 +693,7 @@ class NaturalLanguageProcessor(Processor):
             allowed_nlp_classes = self.extract_allowed_intents(allowed_intents)
 
         if not language and not locale:
-            language, locale = get_language_config(self._app_path)
+            language, locale = self.app_language, self.app_locale
 
         return super().process(
             query_text,
--- a/mindmeld/components/nlp.py
+++ b/mindmeld/components/nlp.py
@@ -27,13 +27,21 @@ from multiprocessing import cpu_count
 
 from .. import path
 from ..core import Bunch, ProcessedQuery
-from ..exceptions import AllowedNlpClassesKeyError, MindMeldImportError, ProcessorError
+from ..exceptions import (
+    AllowedNlpClassesKeyError,
+    MindMeldImportError,
+    ProcessorError,
+    MindMeldError,
+)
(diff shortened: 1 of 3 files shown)


mindmeld/components/nlp.py
```diff
@@ -1077,12 +1099,15 @@ def process(
             (ProcessedQuery): A processed query object that contains the prediction results from \
                 applying the hierarchy of natural language processing models to the input text.
         """
+        del language
```
minhtuev: probably we should add a TODO here so that we can clean up this API in a future version of MindMeld

mindmeld/components/nlp.py
```diff
@@ -824,12 +843,14 @@ def process(
...
```
minhtuev: add TODO

mindmeld/components/nlp.py
```diff
@@ -108,6 +116,7 @@ def __init__(self, app_path, resource_loader=None, config=None):
         self.resource_loader = resource_loader or ResourceLoader.create_resource_loader(
             app_path
         )
+        self.language, self.locale = get_language_config(app_path)
```
minhtuev: This class is also subclassed by DomainProcessor, IntentProcessor, EntityProcessor. Is it intended that each of these classes also need to know about language and locale?

vijay120: Yes, this needs to be done since the client is directly invoke the process api for the DomainProcessor, IntentProcessor, EntityProcessor.

mindmeld/components/nlp.py
vijay120: done

mindmeld/components/nlp.py
vijay120: done

mindmeld/components/nlp.py
```diff
@@ -253,9 +262,12 @@ def process(
                  applying the full hierarchy of natural language processing models to the input \
                  query.
         """
+        # TODO: Deprecate language argument
+        del language
```
serapio: Should note in the doc string that the param is ignored

mindmeld/components/nlp.py
```diff
@@ -684,22 +706,21 @@ def process(
                 applying the full hierarchy of natural language processing models to the input \
                 query.
         """
+        # TODO: Deprecate language argument
+        del language
```
serapio: doc string should indicate param is ignored

vijay120:
@serapio You raise a good point, but here are some reasons the change wouldn't break things:
1. If the developer has a `fr` app and uses the `en` language code during `process`, the app is already broken. Similarly, if the developer has a `fr` app and uses the `en_US` locale code, in the past, it would fail silently, while now, we raise an exception.
2. If the developer has a `fr` app and uses the `fr` language code during `process`, the app will work as expected.

mindmeld/components/nlp.py
vijay120: done

PR #133, opened 2020-03-31
relaxed locale validation

vijay120:
We relax the validation check on the locale such that if the develop or the app doesnt specify it, we just return None.

### DIFF
--- a/mindmeld/components/nlp.py
+++ b/mindmeld/components/nlp.py
@@ -340,6 +340,11 @@ class Processor(ABC):
     def _validate_locale(self, locale=None):
         """This function makes sure the locale is consistent with the app's language code"""
         locale = locale or self.locale
+
+        # if the developer or app doesnt specify the locale, we just use the default locale
+        if not locale:
+            return
+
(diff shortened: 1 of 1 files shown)


mindmeld/components/nlp.py
```diff
@@ -340,6 +340,11 @@ def _process_list(self, items, func, *args, **kwargs):
     def _validate_locale(self, locale=None):
         """This function makes sure the locale is consistent with the app's language code"""
         locale = locale or self.locale
+
+        # if the developer or app doesnt specify the locale, we just use the default locale
```
minhtuev:
```suggestion
        # if the developer or app doesn't specify the locale, we just use the default locale
```

minhtuev: what is the default locale in this context? the default locale of the language?

vijay120:
@angelRev yes.

mindmeld/components/nlp.py
vijay120: Whatever Duckling uses when the locale passed-in in None

PR #134, opened 2020-04-02
Feature/slot filling

ritvikshrivastava:
This PR adds the slot-filling feature to MindMeld.

Usage:
```
from mindmeld.core import FormEntity

 form_transfermoney_en = {
    'entities':[
        FormEntity(
            entity='account_en',
            role='account_from',
            responses=['Sure. Transfer from which account?'],
            custom_eval=test_func # validates this entity using this custom function
            ),
        FormEntity(
            entity='account_en',
            role='account_to',
            responses=['To which account?']
            hints=['checking', 'checkings', ....] # can be only from this list
            ),
        FormEntity(
            entity='sys_amount-of-money',
            responses=['And, how much do you want to transfer?'])
        ],
    'max_retries': 1,
    'exit_keys': [<list of keywords>],
    'exit_msg': 'custom exit message'
    }
```
```
@app.auto_fill(intent='transfermoney_en', form=form_transfermoney_en)
def transfermoney_en_handler(request, responder):
    <DM handler logic>
```
Removes the need to provide `@app.handle` if `@app.auto_fill` is being called. 
The handler functionality will be called by the slot filling class once that process is completed.

- ``form`` is a dictionary containing the following entries:
  - ``entities`` (list, required): List of ``FormEntity`` objects with each defined for one entity slot to be filled.
  - ``max_retries`` (int, optional, default 1): maximum number of retries allowed per entity or slot if user response is invalid.
  - ``exit_msg`` (str, optional): If slot filling is exited abruptly without completion, define custom message to display.
  - ``exit_keys`` (list, optional): List of exit hints for the slot filling flow. If these words or phrases are said by the user, the slot filling logic exits. Default: ['cancel', 'restart', 'exit', 'reset'].

   <br />

- ``FormEntity`` is a class that allows creation of entity objects for slot filling and comprises of the following attributes:
  - ``entity`` (str, required): Entity name
  - ``role`` (str, optional): The role of the entity
  - ``responses`` (list or str, optional): message for prompting the user for missing entities
  - ``retry_response`` (str, optional): message for re-prompting users. If not provided, defaults to ``responses``.
  - ``value`` (str, optional): The resolved value of the entity
  - ``default_eval`` (bool, optional): Use system validation (default: True)
  - ``hints`` (list, optional): Developer defined list of keywords to verify the``user`` input against
  - ``custom_eval`` (func, optional): custom validation function (should return bool:
    validated or not)

The validation types (default (sys and gaz validation), hints (user list validation), custom validation (developer defined validation)) can be used in any combination or standalone.


minhtuev:
A few questions:
1) Is `entity_form` a required field in the frame for slot filing?
2) If there is a validation issue, is there a validation exception? For custom validation 
3) Is `form_fill` function only used for validation?
4) What is the usage of `constraint: closed`? It is not mentioned anywhere in the code?

Alternatively, we can do something similar to this:
```
nlr_form = {
    'intent': 'transfermoney_en',
    'entity': [
        {'entity': 'account_en',
         'role': 'account_from',
         'response': 'Sure. Transfer from which account?',
         'value': None},
        {'entity': 'account_en',
         'role': 'account_to',
         'response': 'To which account?',
         'value': None},
        {'entity': 'sys_amount-of-money',
         'response': 'And, how much do you want to transfer?',
         'value': None}
    ],
    ‘restart’: [‘restart’, ‘exit’]
}

@app.form_fill(nlr_form)
def xyz_handler(request, responder):
    <rest of DM handler logic>
```

ritvikshrivastava:
1. I was considering making `entity_form` optional or instead passing it as kwargs, something consistent with the input style you have suggested.
2. Currently failed validation handling is only done for default (duckling, gazetteer) and user list validation. Shouldn't the erroneous input handling for a developer defined validation function be a feature of their custom function in itself? I don't think we need to make checks for custom functions.
3.  Currently, yes. Either defining a validation function or calling different validation functions in that space. I've kept it clean for now since that function acts as a generic function for all dialog state handlers.
4. Yeah I didn't feel the need to include it (have kept it as a placeholder in the form). Do you think there is a use for making open vs closed distinctions for entity constraints?

About the alternate design:
The issue I see with that design is defining different NLR/entity forms for each intent outside the handler will cause the generation of too many variable name forms for supporting form filling for each intent.

arushir:
For the `entity_form` I think it would be better to define an object with the expected attributes, so it becomes a list of objects rather than a list of dictionaries

arushir:
Regarding (4) - I don't think this is needed here. This terminology is referring to 'closed' entities being entities with a fixed number of possible values and 'open' being entities which can have a wide variety of values, correct? I don't think we've made that terminology clear to external developers, and it doesn't seem like it's being used right now in the code. If anything, we can add an optional parameter to the entity form object which developer can populate with their own terminology and use in their validation logic.

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,187 @@ def _get_dialogue_handler(self, dialogue_state):
...
+        """ 
+        payload format for entity feature extractors: 
+            tuple(query (Query Object), list of entities, entity index)
+
+        For slot filling, the user query will consist of the prompted entity. 
+        Hence `entity = [query]` and `entity_index = 0`.
+
+        """
+        formatted_payload = (query, [query], 0)
+        return formatted_payload
```
arushir, line 748: We shouldn't need to re-run this right? Instead pass in the extracted entities from the request object?

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,187 @@ def _get_dialogue_handler(self, dialogue_state):
...
+
+        Returns:
+            (bool): Boolean whether the user input fulfills the slot requirements
+        """
+        formatted_payload = self._extract_query_features(text)
+
+        if 'sys_' in entity_type:
+            # duckling validation
+            extracted_feature = (
+                entity_features.extract_numeric_candidate_features()(formatted_payload, {})
```
arushir, line 765: this just checks the presence of any time or interval entity, right? i think we should be checking if the entity type is matching the expected type (e.g. date if expecting date, etc.). beyond duckling, we can also check that types like string, float, int are matching the expected type

ritvikshrivastava: Fixed, now it makes a check for any system entity and whether the user input is matching or not.

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,201 @@ def _get_dialogue_handler(self, dialogue_state):
...
+
+        """ Extracting the Query object associated with the user input here.
+
+        process_markup args:
+            text, query factory, query options (optional dict) {language, time_zone and time_stamp}
+
+        returns: 
+            tuple(raw_text (str), Query object, list of entities)
+        """
+        query = process_markup(text, resource_loader.query_factory, {})[1]
```
vijay120, line 742: why the index `1`? Maybe assign `index = 1` and explain what index is doing

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,201 @@ def _get_dialogue_handler(self, dialogue_state):
...
+            text (Request.text): Text in the query
+            entity_type: Type of entity
+
+        Returns:
+            (bool): Boolean whether the user input fulfills the slot requirements
+        """
+        formatted_payload = self._extract_query_features(text)
+        query = formatted_payload[0]
+
+        if 'sys_' in entity_type:
```
vijay120, line 769: We should be enumerating all the duckling sys entities here instead

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,201 @@ def _get_dialogue_handler(self, dialogue_state):
...
+            # system entity validation
+            # checks whether required system entity is present in query or not.
+            extracted_feature = dict(
+                query_features.extract_sys_candidates(entities=entity_type)(query, {})
```
vijay120, line 773: why `{}`?

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,201 @@ def _get_dialogue_handler(self, dialogue_state):
...
+            )
+        else:
+            if self._validation_type == 'ulist':
+                # validation using user defined list
+                return True if text in self._user_list else False
+            else:
+                # gazetteer validation
+                gazetter = (
```
vijay120, line 781: gazetteer

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,201 @@ def _get_dialogue_handler(self, dialogue_state):
...
+                    {'gazetteers': 
+                        {entity_type: 
+                        self.app.app_manager.nlp.resource_loader.get_gazetteer(entity_type)}
+                        }
+                )
+                extracted_feature = (
+                entity_features.extract_in_gaz_features()(formatted_payload, gazetter))
+
+        return extracted_feature != {}
```
vijay120, line 790:
```
empty_dict = {}
return extracted_feature != empty_dict
```

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,201 @@ def _get_dialogue_handler(self, dialogue_state):
...
+        query = formatted_payload[0]
+
+        if 'sys_' in entity_type:
+            # system entity validation
+            # checks whether required system entity is present in query or not.
+            extracted_feature = dict(
+                query_features.extract_sys_candidates(entities=entity_type)(query, {})
+            )
+        else:
+            if self._validation_type == 'ulist':
```
vijay120, line 776: I think `hints` might be a better terminology here than `ulist`, since this is a list of `hints` provided by the client to match against, similar to Google Speech ASR system?

mindmeld/components/dialogue.py
arushir: @vijay120 It is noted in the docstring above that the return value of process_markup is `tuple(raw_text (str), Query object, list of entities)`, so index 1 is getting the Query object. Do you think it needs additional documentation?

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,201 @@ def _get_dialogue_handler(self, dialogue_state):
...
+        """ 
+        payload format for entity feature extractors: 
+            tuple(query (Query Object), list of entities, entity index)
+
+        For slot filling, the user query will consist of the prompted entity. 
+        Hence `entity = [query]` and `entity_index = 0`.
+
+        """
+        formatted_payload = (query, [query], 0)
+        return formatted_payload
```
arushir, line 754: why not return just the query?

mindmeld/components/dialogue.py
arushir: It's a slightly different concept for us, since Google ASR `hints` are expected words but not a comprehensive list of what can be recognized. (Just a thought - not sure what the best terminology is here)

mindmeld/components/dialogue.py
arushir: can we add a check here so that it only defaults to gazetteer validation if a non-empty gazetteer exists for that entity? for a lot of app entities developers may have empty gazetteer files

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,201 @@ def _get_dialogue_handler(self, dialogue_state):
...
+
+    def _validate(self, text, entity_type):
```
arushir, line 756: It might be worth separating out the duckling, ulist, and gazetteer validations into three separate public methods that the developer could leverage if they decide to define their own validation function and want to use one of those as part of it. We can then call those methods from here as well

mindmeld/models/query_features.py
```diff
@@ -529,7 +529,7 @@ def _extractor(query, resources):
         for entity in system_entities:
             for i in entity.token_span:
                 for j in start_positions:
-                    if 0 < i - j < len(feat_seq):
+                    if 0 <= i - j < len(feat_seq):
```
arushir, line 532: why this change?

mindmeld/components/dialogue.py
ritvikshrivastava: Added another line in the docstring to clarify this.

minhtuev:
Nice! Can you add tests for this logic?

One idea is to reimplement this autofill logic for asynchronous kwik-e-mart. Here, instead of using dialogueflow, we can use the new autofill function
https://github.com/cisco/mindmeld/blob/67a9195392c409864fbd0debde3a7393dce727b2/tests/kwik_e_mart/app_async.py#L71

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,194 @@ def _get_dialogue_handler(self, dialogue_state):
...
+    Attributes:
+        app (Application): The application that initializes this flow.        
+    """
+
+    logger = mod_logger.getChild("AutoEntityFilling")
+    """Class logger."""
+
+    def __init__(self, entrance_handler, entity_form, app, **kwargs):
+        self.app = app
+        self.name = entrance_handler.__name__
```
minhtuev, line 709:
```suggestion
        self._name = entrance_handler.__name__
```

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,194 @@ def _get_dialogue_handler(self, dialogue_state):
...
+
+    Attributes:
+        app (Application): The application that initializes this flow.        
+    """
+
+    logger = mod_logger.getChild("AutoEntityFilling")
+    """Class logger."""
+
+    def __init__(self, entrance_handler, entity_form, app, **kwargs):
+        self.app = app
```
minhtuev, line 708: we should make all of these into private variables
```suggestion
        self._app = app
```

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,194 @@ def _get_dialogue_handler(self, dialogue_state):
...
+        self.name = entrance_handler.__name__
+        self.entrance_handler = entrance_handler
```
minhtuev, line 710:
```suggestion
        self._entrance_handler = entrance_handler
```

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,194 @@ def _get_dialogue_handler(self, dialogue_state):
...
+        self._entity_form = entity_form
+
+    def _set_target_state(self):
+        """Set target dialogue state to current flow"""
+        responder.params.target_dialogue_state = self.entrance_handler+'auto_fill'
```
minhtuev, line 715:
```suggestion
        responder.params.target_dialogue_state = self.entrance_handler + 'auto_fill'
```

mindmeld/components/dialogue.py
ritvikshrivastava: I was keeping this behavior consistent with dialogueflow class, can make it a hidden variable if that's ideal

mindmeld/models/query_features.py
minhtuev: probably because of VJ's previous change. we should merge to master here?

mindmeld/models/query_features.py
```diff
@@ -670,15 +670,15 @@ def _extractor(query, resources):
...
+def extract_sys_candidates(**args):
     """
     Return an extractor for features based on a heuristic guess of numeric \
         candidates in the current query.
 
     Returns:
             (function) The feature extractor.
      """
-    del args
+    entities = [args['entities']] if 'entities' in args else None
```
minhtuev, line 681: I assume this is to fix another bug in `extract_sys_candidates`? if so in the interest of expediency, should we split this off into its own PR?

ritvikshrivastava: It isn't, it's a modification to the existing extractor. The change additionally enables passing a specific system entity to obtain only corresponding features. The default was to return features for all system entities in the query. This change allows for validating presence of that particular system entity.

mindmeld/models/query_features.py
ritvikshrivastava: Yeah, I'll rebase this PR

mindmeld/models/query_features.py
ritvikshrivastava: Reverted back to original, made changes in the way slot filling was calling this logic to support original behavior of this function.

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,189 @@ def _get_dialogue_handler(self, dialogue_state):
...
+    requirements for entity slots.
+
+    Attributes:
+        app (Application): The application that initializes this flow.
+    """
+
+    logger = mod_logger.getChild("AutoEntityFilling")
+    """Class logger."""
+
+    def __init__(self, entrance_handler, entity_form, retry_attempts, app, **kwargs):
```
minhtuev, line 707: We should set a default value to `retry_attempts` and validate `entity_form`, at least when `entity_form` is None we should throw an error

```suggestion
    def __init__(self, entrance_handler, entity_form, retry_attempts=1, app, **kwargs):
```

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,189 @@ def _get_dialogue_handler(self, dialogue_state):
...
+        self._app = app
+        self._entrance_handler = entrance_handler
+        self._entity_form = entity_form
+        self._retry_attempts = retry_attempts
+        self._local_form = None
+        self._slot_not_prompted = None
+
+    def _set_target_state(self, responder):
+        """Set target dialogue state to current flow"""
```
minhtuev, line 716:
```suggestion
        """Set target dialogue state to the entrance handler's name"""
```

mindmeld/core.py
```diff
@@ -747,6 +747,38 @@ def __repr__(self):
...
+        value (str): The resolved value of the entity
+        default_eval(bool): Use system validation (default: True)
+        hints(list): Developer defined list of keywords to verify the user input against
+        custom_eval(func): custom validation function
+    """
+    def __init__(
+        self,
+        entity,
+        role=None,
+        responses=None,
```
minhtuev, line 767: We should validate:
1) entity name is not None
2) there is at least one Response

Alternatively, for 2) we can also set a default value for Response as well

mindmeld/components/dialogue.py
minhtuev: Should we add the language/timezone/time_stamp as well?

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,189 @@ def _get_dialogue_handler(self, dialogue_state):
...
+            else:
+                # gazetteer validation
+                gazetteer = (
+                    {'gazetteers':
+                        {entity_type:
+                            self._app.app_manager.nlp.resource_loader.get_gazetteer(entity_type)
+                        }
+                    }
+                )
+                extracted_feature = (
```
minhtuev, line 791: here extracted feature is a tuple, not a dict like other cases?

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,189 @@ def _get_dialogue_handler(self, dialogue_state):
...
+        """
+        payload format for entity feature extractors:
+            tuple(query (Query Object), list of entities, entity index)
+
+        For slot filling, the user query will consist of the prompted entity.
+        Hence `entity = [query]` and `entity_index = 0`.
+
+        """
+        formatted_payload = (query, [query], 0)
+
```
minhtuev, line 774: probably we should define ```extracted_feature = {}``` here

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,189 @@ def _get_dialogue_handler(self, dialogue_state):
...
+        """
+        for entity in request.entities:
+            entity_type = entity['type']
+            role = entity['role']
+            value = entity['text']
+
+            for slot in self._local_form:
+                if entity_type == slot.entity:
+                    if slot.role == None or (role == slot.role):
+                        slot.value = value
```
minhtuev, line 817: we should break the loop here since we only expect one slot in the local_form for each entity

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,189 @@ def _get_dialogue_handler(self, dialogue_state):
...
+            self._slot_not_prompted = True
+            # Fill the form with the entities in the first query
+            self._local_form = copy.deepcopy(self._entity_form)
+            self._initial_fill(request)
+
+        for slot in self._local_form:
+            entity_ = slot.entity
+            value = slot.value
+            nlr = (slot.responses if slot.responses
+                else ["Please provide value for: {}".format(entity_)]
```
minhtuev, line 842: Should we move this into validation or some earlier steps?

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,189 @@ def _get_dialogue_handler(self, dialogue_state):
...
+                        self._slot_not_prompted = True
+                    else:
+                        # retry logic
+                        if 'retry_count' in responder.frame:
+                            if responder.frame['retry_count'] < self._retry_attempts:
+                                responder.frame['retry_count'] += 1
+                            else:
+                                # max attempts exceeded, reset counter, exit flow.
+                                del responder.frame['retry_count']
+                                responder.reply(['Invalid response. How can I help you?'])
```
minhtuev, line 863: We should have a way to customize this response as well

mindmeld/components/dialogue.py
ritvikshrivastava: I've set this default value in app.py, at the point of object initialization

mindmeld/components/dialogue.py
ritvikshrivastava: It is a dict.

mindmeld/core.py
ritvikshrivastava: 2 is already done (default response).

1 is handled currently by exiting the turn.

mindmeld/components/dialogue.py
ritvikshrivastava: I can think of usecases where the same entity might fill two slots?

mindmeld/components/dialogue.py
ritvikshrivastava: This is just setting the default nlr at an object level. This is the first time the object is accessed from the form, I don't see where it can go earlier. Validation is a later step. This is just a default user prompt for the slot value, validation is done on the value they return.

mindmeld/components/dialogue.py
ritvikshrivastava: Removed this response altogether, now it reprocesses the current query with no target dialogue state once it's reached the retry attempt limit - smoother transition to new flow/handler.

mindmeld/components/dialogue.py
ritvikshrivastava: updated.

mindmeld/app.py
```diff
@@ -171,6 +172,20 @@ def _decorator(func):
 
         return _decorator
 
+    def auto_fill(self, **kwargs):
+        """Creates a flow to fill missing entities"""
+
+        def _decorator(func):
+            name = kwargs.pop("name", func.__name__)
+            entity_form = kwargs.pop("entity_form", None)
+            assert entity_form not in (None, ''), "Entity Form cannot be empty."
```
minhtuev, line 181:
```suggestion
            assert entity_form, "Entity Form cannot be empty."
```

mindmeld/core.py
```diff
@@ -747,6 +747,44 @@ def __repr__(self):
...
+        entity=None,
+        role=None,
+        responses=None,
+        value=None,
+        default_eval=True,
+        hints=None,
+        custom_eval=None
+    ):
+        self.entity = entity
+        assert self.entity not in (None, ''), 'Entity cannot be empty.'
```
minhtuev, line 776:
```suggestion
        assert entity_form, "Entity cannot be empty."
```

mindmeld/components/dialogue.py
minhtuev: did we decide to break here?

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,182 @@ def _get_dialogue_handler(self, dialogue_state):
...
+    _logger = mod_logger.getChild("AutoEntityFilling")
+    """Class logger."""
+
+    def __init__(self, entrance_handler, entity_form, app, max_retries=1):
+        self._app = app
+        self._entrance_handler = entrance_handler
+        self._entity_form = entity_form
+        self._max_retries = max_retries  # default 1
+        self._local_form = None
+        self._slot_not_prompted = None
```
vijay120, line 712: what does the variable mean? It could be renamed better.

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,182 @@ def _get_dialogue_handler(self, dialogue_state):
...
+                gaz = self._app.app_manager.nlp.resource_loader.get_gazetteer(entity_type)
+
+                if len(gaz) > 0:
+                    gazetteer = {'gazetteers': {entity_type: gaz}}
+                    extracted_feature = (
+                        entity_features.extract_in_gaz_features()(formatted_payload, gazetteer))
+
+        if slot.hints:
+            # hints / user-list validation
+            extracted_feature = {'hint_validated_entity': text} if text in slot.hints else {}
```
vijay120, line 787: So I think we can extend `slot.hints` with `slot.default_eval` by doing this:
```
if extracted_feature == {}:
    extracted_feature = {'hint_validated_entity': text} if text in slot.hints else {}
else:
    extracted_feature.update({'hint_validated_entity': text} if text in slot.hints else {})
```

mindmeld/components/dialogue.py
```diff
@@ -689,6 +691,182 @@ def _get_dialogue_handler(self, dialogue_state):
...
+
+        if slot.custom_eval:
+            # Custom validation using function provided by developer. Should return True/False
+            # for validation status. If true, then continue, else fail overall validation.
+            if slot.custom_eval(text) not in (True, False) or slot.custom_eval(text) is False:
```
vijay120, line 792: Again, I think this condition can be met along with the two above.

mindmeld/components/dialogue.py
vijay120: I agree with ritvik, we should not break here since we should fill as many slots as possible

mindmeld/app.py
```diff
@@ -171,6 +172,20 @@ def _decorator(func):
 
         return _decorator
 
+    def auto_fill(self, **kwargs):
+        """Creates a flow to fill missing entities"""
+
```
davidmfrey, line 177: Can we add a `functools.wraps()` here to keep the signature and docstrings of the wrapped function?

tmehlinger: Following from that, unless you need to pass keywords through without modification, it's better to provide explicitly named parameters with defaults rather than `kwargs`.

```python
def auto_fill(self, name=None, entity_form=None, max_retries=1):
    ...
```

mindmeld/app.py
```diff
@@ -171,6 +172,20 @@ def _decorator(func):
...
+        def _decorator(func):
+            name = kwargs.pop("name", func.__name__)
```
tmehlinger, line 179: If you go with explicitly-named parameters, here's a common pattern for assigning defaults from a `None` value:
```python
entity_form = name or func.__name__
```

mindmeld/app.py
```diff
@@ -171,6 +172,20 @@ def _decorator(func):
...
+            entity_form = kwargs.pop("entity_form", None)
+            assert entity_form, "Entity Form cannot be empty."
```
tmehlinger, line 181: If you need to verify the type of some variable, it's better to raise a `TypeError` rather than an assertion. You may also want to use `isinstance` to see what the parameter actually is, though I don't know exactly what you expect here.
```python
if not isinstance(entity_form, (tuple, list, MyType)):
    raise TypeError(...)
```
The reason is twofold. One is that a TypeError simply conveys more explicit information ("hey dude, the type is wrong!") and, though this doesn't happen often, if Python is invoked with the `-O` option to optimize bytecode, the interpreter strips assertions.

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,215 @@ def _get_dialogue_handler(self, dialogue_state):
...
+        for slot in self._local_form:
+            nlr = slot.responses
+
+            if not slot.value:
+                # check if user has been prompted for this entity slot
+                if self._prompt_user:
+                    return self._prompt_slot(responder, nlr)
+                else:
+                    # If prompted, validate the user response and retry if invalid response
+                    _isValid, _resolved_value = self._validate(request, slot)
```
tmehlinger, line 887: `_isValid` should be snake-cased.

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,215 @@ def _get_dialogue_handler(self, dialogue_state):
...
+                        slot.value = Entity(
+                            text=request.text,
+                            entity_type=slot.entity,
+                            role=slot.role,
+                            value=_resolved_value).to_dict()
+
+                        self._prompt_user = True
+                    else:
+                        # retry logic
+                        return self._retry_logic(responder, nlr)
```
tmehlinger, line 898: Tip: if you have `return` in an `if` block, you _never_ need `else`.

You can also invert some of these conditions to reduce indentation and make this easier to read:

```python
if self._prompt_user:
    return self._prompt_slot(responder, nlr)

_is_valid, _resolved_value = self._validate(request, slot)

if not _is_valid:
    return return self._retry_logic(responder, nlr)

slot.value = Entity(
    text=request.text,
    entity_type=slot.entity,
    role=slot.role,
    value=_resolved_value).to_dict()

self._prompt_user = True
```

mindmeld/app.py
```diff
@@ -171,6 +172,20 @@ def _decorator(func):
 
         return _decorator
 
+    def auto_fill(self, **kwargs):
+        """Creates a flow to fill missing entities"""
+
+        def _decorator(func):
+            name = kwargs.pop("name", func.__name__)
+            entity_form = kwargs.pop("entity_form", None)
```
davidmfrey, line 180: You can move this argument to be explicitly required in the decorator signature.  This will enforce the check at parse time rather than runtime as well.

`def autofill(self, max_retries=1, *, entity_form=None, **kwargs):
`

Disclaimer: I didn't syntax check this ^^, but something along these lines should do it.

mindmeld/app.py
davidmfrey: I think you mean `name = name or func.__name__`

mindmeld/app.py
davidmfrey: Following from that, you may want something that looks like

`
def auto_fill(self, name=None, max_retries=1, *, entity_form, **kwargs):
    ...`

- You need `**kwargs` to pass it into `add_dialogue_rule`
- Making `entity_form` a required keyword-only arg performs the assertion in line 181 at parse time rather than runtime.  Hence you can drop that assertion.

mindmeld/core.py
```diff
@@ -747,6 +747,44 @@ def __repr__(self):
...
+        assert self.entity, 'Entity cannot be empty.'
+        self.role = role
+        self.responses = (
+            responses or ["Please provide value for: {}".format(self.entity)])
+        self.value = value
+        self.default_eval = default_eval
+        self.hints = hints
+        self.custom_eval = custom_eval
+        if self.custom_eval:
+            assert callable(custom_eval) is True, 'Invalid custom validation function.'
```
minhtuev, line 785: Use exception here as well

mindmeld/app.py
ritvikshrivastava: The idea here was to ensure that the form isn't empty. Do you think TypeError goes for that as well?

mindmeld/app.py
ritvikshrivastava: I guess since this is from the developer point of view, shouldn't the assertion be performed at parse time and fail the app from loading? If that assertion is removed, the app is allowed to load even with a empty form and throw an error at runtime (user-facing).

mindmeld/app.py
ritvikshrivastava: `name = name or func.__name__` will result in a referenced before assignment error here I think.

Also, this behavior was just to keep it consistent with the @app.dialogue_flow decorator. Can change it if you think that's better?

mindmeld/app.py
minhtuev: Let's keep the assertion but change the function signature here to be more explicit.

mindmeld/app.py
minhtuev: Once `name` is defined above in the function signature it will be ok.

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,249 @@ def _get_dialogue_handler(self, dialogue_state):
...
+        self._exit_response = (
+            self._form['exit_msg'] if 'exit_msg' in self._form else 'How may I help you?')
+        self._exit_intent = (
+            self._form['exit_intent'] if 'exit_intent' in self._form else None)
+        self._exit_keys = (
+            self._form['exit_keys'] if 'exit_keys' in self._form else
+            ['cancel', 'restart', 'exit', 'reset'])
+
+        if not isinstance(self._max_retries, int):
+                    raise TypeError("'max_retries' should be of type: int.")
```
arushir, line 731: (minor) indentation looks off here - did the changes pass lint checks?

ritvikshrivastava: Strange, it did pass the lint checks.

Updated it anyhow!

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,249 @@ def _get_dialogue_handler(self, dialogue_state):
...
+        if self._exit_intent and not isinstance(self._exit_intent, list):
+                    raise TypeError("'exit_intent' should be of type: list.")
+        if not isinstance(self._exit_keys, list):
+            raise TypeError("'exit_keys' should be of type: list.")
+
+        self._exit_keys = list(map(str.lower, self._exit_keys))
+
+    def _set_target_state(self, request, responder):
+        """Set target dialogue state to the entrance handler's name"""
+        responder.params.allowed_intents = tuple([''.join([request.domain, ".", request.intent])])
```
vijay120, line 743: Nit:` tuple('{}.{}'.format(request.domain, request.intent))`

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,249 @@ def _get_dialogue_handler(self, dialogue_state):
...
+        responder.params.target_dialogue_state = self._entrance_handler.__name__
+
+    def _exit_flow(self, responder, intents=None):
+        """Exits this flow and clears the related parameter for re-usability"""
+        self._prompt_user = None
+        self._local_form = None
+        responder.params.allowed_intents = tuple(intents) if intents else tuple()
```
vijay120, line 750: should this be `tuple('{}.{}'.format(request.domain, request.intent))` as well?

ritvikshrivastava: No, this is only populated when the dev has provided the domain.intent list in exit intents (in the docs above).

source/userguide/dm.rst
```diff
@@ -567,6 +567,96 @@ There are three ways to exit a Dialogue Flow:
...
+
+This decorator replaces the need to define the ``@app.handle`` decorator. MindMeld will prompt the user for missing entities before applying the handler functionality by itself. The arguments required for this decorator are all the rules that would apply for that dialogue state (such as domain, intent, entities etc.), with the addition of a ``form`` argument.
+
+- ``form`` is a dictionary containing the following entries:
+
+  - ``entities`` (list, required): List of ``FormEntity`` objects with each defined for one entity slot to be filled.
+  - ``max_retries`` (int, optional, default 1): maximum number of retries allowed per entity or slot if user response is invalid.
+  - ``exit_msg`` (str, optional): If slot filling is exited abruptly without completion, define custom message to display.
+  - ``exit_intent`` (list, optional): If slot filling is exited abruptly without completion, the developer can define which particular intents (format ``['domain.intent', 'domain.intent'``) should the following query be redirected to. Default: None.
```
vijay120, line 584: Could we explain a bit further here that the format for exit intent is `domain.intent` or `domain.*`?

source/userguide/dm.rst
```diff
@@ -567,6 +567,96 @@ There are three ways to exit a Dialogue Flow:
...
+Automatic Slot Filling for Entities
+-----------------------------------
+
+MindMeld provides a useful functionality for automatically prompting the user for missing entities or slots required to fulfill an intent. This is done via applying the ``@app.auto_fill`` decorator for the dialogue state handler requiring entities to be obtained prior to applying the functionality defined within.
+
+This decorator replaces the need to define the ``@app.handle`` decorator. MindMeld will prompt the user for missing entities before applying the handler functionality by itself. The arguments required for this decorator are all the rules that would apply for that dialogue state (such as domain, intent, entities etc.), with the addition of a ``form`` argument.
+
+- ``form`` is a dictionary containing the following entries:
+
+  - ``entities`` (list, required): List of ``FormEntity`` objects with each defined for one entity slot to be filled.
```
vijay120, line 581: Could you provide a description about how order matters in filling slots using this list?

mindmeld/components/dialogue.py
minhtuev, line 750: This is setting the next allowed_intents to be the exit intents, regardless of what the user is saying, which is not what we want. We should remove the exit_intents for now.

```suggestion
        responder.params.allowed_intents = tuple(intents) if intents else tuple()
```

source/userguide/dm.rst
```diff
@@ -567,6 +567,96 @@ There are three ways to exit a Dialogue Flow:
...
+  - ``max_retries`` (int, optional, default 1): maximum number of retries allowed per entity or slot if user response is invalid.
+  - ``exit_msg`` (str, optional): If slot filling is exited abruptly without completion, define custom message to display.
+  - ``exit_intent`` (list, optional): If slot filling is exited abruptly without completion, the developer can define which particular intents (format ``['domain.intent', 'domain.intent'``) should the following query be redirected to. Default: None.
+  - ``exit_hints`` (list, optional): List of exit hints for the slot filling flow. If these words or phrases are said by the user, the slot filling logic exits. Default: ['cancel', 'restart', 'exit', 'reset'].
```
minhtuev, line 585:
```suggestion
  - ``exit_keys`` (list, optional): List of exit hints for the slot filling flow. If these words or phrases are said by the user, the slot filling logic exits. Default: ['cancel', 'restart', 'exit', 'reset'].
```

source/userguide/dm.rst
ritvikshrivastava: Removed this feature after speaking with Minh, seems like it could add confusion.

mindmeld/components/dialogue.py
```diff
@@ -689,6 +692,249 @@ def _get_dialogue_handler(self, dialogue_state):
...
+                slot.value = Entity(
+                    text=request.text,
+                    entity_type=slot.entity,
+                    role=slot.role,
+                    value=_resolved_value).to_dict()
+
+                self._prompt_user = True
+
+        # Finish slot-filling and return to handler
+        return self._end_slot_fill(request, responder)
```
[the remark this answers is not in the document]
ritvikshrivastava: Can't do it here, it's actually returning the handler at this point. Changed the rest of them.

mindmeld/app.py
```diff
@@ -63,6 +63,7 @@ def __init__(
         self.request_class = request_class or Request
         self.responder_class = responder_class or DialogueResponder
         self.preprocessor = preprocessor
+        self.autofill = AutoEntityFilling
```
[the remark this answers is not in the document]
ritvikshrivastava: Fixed.

source/userguide/dm.rst
ritvikshrivastava: Ah, good idea! Done.

mindmeld/components/dialogue.py
ritvikshrivastava: Removed this feature.

mindmeld/core.py
```diff
@@ -747,6 +747,46 @@ def __repr__(self):
...
+        role=None,
+        responses=None,
+        retry_response=None,
+        value=None,
+        default_eval=True,
+        hints=None,
+        custom_eval=None
+    ):
+        self.entity = entity
+        assert self.entity, 'Entity cannot be empty.'
```
minhtuev, line 777: Change this into an Exception too

source/userguide/dm.rst
```diff
@@ -567,6 +567,98 @@ There are three ways to exit a Dialogue Flow:
...
+
+MindMeld provides a useful functionality for automatically prompting the user for missing entities or slots required to fulfill an intent. This is done via applying the ``@app.auto_fill`` decorator for the dialogue state handler requiring entities to be obtained prior to applying the functionality defined within.
+
+This decorator replaces the need to define the ``@app.handle`` decorator. MindMeld will prompt the user for missing entities before applying the handler functionality by itself. The arguments required for this decorator are all the rules that would apply for that dialogue state (such as domain, intent, entities etc.), with the addition of a ``form`` argument.
+
+- ``form`` is a dictionary containing the following entries:
+
+  - ``entities`` (list, required): List of ``FormEntity`` objects with each defined for one entity slot to be filled. The order of entities provided in this list is important as the slots will be prompted in that order. So the order should follow the flow of the dialogue intended for slot filling.
+  - ``max_retries`` (int, optional, default 1): maximum number of retries allowed per entity or slot if user response is invalid.
```
minhtuev, line 582:
```suggestion
  - ``max_retries`` (int, optional, default 1): Maximum number of retries allowed per entity or slot if user response is invalid.
```

source/userguide/dm.rst
```diff
@@ -567,6 +567,98 @@ There are three ways to exit a Dialogue Flow:
...
+
+.. |br| raw:: html
+
+   <br />
+
+- ``FormEntity`` is a class that allows creation of entity objects for slot filling and comprises of the following attributes:
+  
+  - ``entity`` (str, required): Entity name
+  - ``role`` (str, optional): The role of the entity
+  - ``responses`` (list or str, optional): message for prompting the user for missing entities
```
minhtuev, line 594:
```suggestion
  - ``responses`` (list or str, optional): Message for prompting the user for missing entities
```

source/userguide/dm.rst
```diff
@@ -567,6 +567,98 @@ There are three ways to exit a Dialogue Flow:
...
+  - ``retry_response`` (str, optional): message for re-prompting users. If not provided, defaults to ``responses``.
```
minhtuev, line 595:
```suggestion
  - ``retry_response`` (str, optional): Message for re-prompting users. If not provided, defaults to ``responses``.
```

source/userguide/dm.rst
```diff
@@ -567,6 +567,98 @@ There are three ways to exit a Dialogue Flow:
...
+  - ``value`` (str, optional): The resolved value of the entity
+  - ``default_eval`` (bool, optional): Use system validation (default: True)
+  - ``hints`` (list, optional): Developer defined list of keywords to verify the user input against
+  - ``custom_eval`` (func, optional): custom validation function (should return bool:
```
minhtuev, line 599:
```suggestion
  - ``custom_eval`` (func, optional): Custom validation function (should return bool:
```

source/userguide/dm.rst
```diff
@@ -567,6 +567,98 @@ There are three ways to exit a Dialogue Flow:
...
+              responses=['To which account?']
+              hints=['checking', 'checkings', ....] # can be only from this list
+              ),
+          FormEntity(
+              entity='sys_amount-of-money',
+              responses=['And, how much do you want to transfer?']
+              )
+          ],
+      'max_retries': 1,
+      'exit_keys': [<list of keywords>],
```
minhtuev, line 641: We should give a real list here

```suggestion
      'exit_keys': [<list of keywords>],
```

source/userguide/dm.rst
```diff
@@ -567,6 +567,98 @@ There are three ways to exit a Dialogue Flow:
...
+      'exit_msg': 'custom exit message'
```
minhtuev, line 642: Use real example here
```suggestion
      'exit_msg': 'custom exit message'
```

source/userguide/dm.rst
```diff
@@ -567,6 +567,98 @@ There are three ways to exit a Dialogue Flow:
...
+
+    from mindmeld.core import FormEntity
+
+    form_transfermoney = {
+      'entities':[
+          FormEntity(
+              entity='account_en',
+              role='account_from',
+              responses=['Sure. Transfer from which account?'],
+              custom_eval=developer_func # validates the user-response for this entity
```
minhtuev, line 627: Maybe better to use custom_eval for 'sys_amount-of-money'. It could be something as simple as checking the "$" sign in the string.

```suggestion
              custom_eval=developer_func # validates the user-response for this entity 
```

ritvikshrivastava: Added.

PR #135, opened 2020-04-03
Fixed system entity span detection logic to detect entities with no context

vijay120:
This PR fixes a bug in the entity recognizer that prevents it from detecting entities with no context, even if the numerical parser correctly identifies the entity at a system entity.

For example, the query `2am` would not be tagged as a `sys_time` entity even though the numerical parser correctly recognized `2am` as a sys_time entity.


mindmeld/models/query_features.py
```diff
@@ -529,7 +529,7 @@ def _extractor(query, resources):
         for entity in system_entities:
             for i in entity.token_span:
                 for j in start_positions:
-                    if 0 < i - j < len(feat_seq):
+                    if 0 <= i - j < len(feat_seq):
```
minhtuev, line 532: Should a similar logic apply to the other extreme as well?
```suggestion
                    if 0 <= i - j <= len(feat_seq):
```

serapio:
To rephrase the condition to be sure I understand: When a system entity was at the first token of a query, that token would not get the sys entity features, which can throw off the tagger, but the behavior would be transient, since other active features could be a strong enough signal to get that tag right, and then the rest would follow?

minhtuev:
Thanks @vijay120 for fixing this bug, should we add some direct tests for `extract_sys_candidate_features` as well?

vijay120:
@serapio 
>When a system entity was at the first token of a query, that token would not get the sys entity features, which can throw off the tagger
That is correct

>but the behavior would be transient, since other active features could be a strong enough signal to get that tag right, and then the rest would follow
The behavior is not transient since in the presence of the query `2am`, the only signal is the system entity signal, since the training data doesnt have this exist word, there is no context and no gazetteer as well.

vijay120:
@minhtuev the above test for just `2am` tests that functionality since there is no training data for it.

mindmeld/models/query_features.py
vijay120: No, since the access pattern is `feat_seq[i - j]`, `i-j < len(feat_seq)`

tests/test_query_features.py
```diff
@@ -427,3 +427,25 @@ def test_entity_gaz_query_features(
...
+        ('2am', 'sys_time', 'times_and_dates', 'change_alarm'),
+        ('4:30 pm', 'sys_time', 'times_and_dates', 'change_alarm'),
+        ('tomorrow at noon', 'sys_time', 'times_and_dates', 'change_alarm'),
+        # custom entity in training data should be detected
+        ('bathroom', 'location', 'smart_home', 'set_thermostat'),
+        ('bedroom', 'location', 'smart_home', 'set_thermostat'),
+        ('balcony', 'location', 'smart_home', 'set_thermostat'),
+        # gaz entity should be detected even if not in training data
+        ('powder room', 'location', 'smart_home', 'set_thermostat'),
+        ('living room', 'location', 'smart_home', 'set_thermostat'),
```
jjjacksn, line 445: Were all of these tests failing previously?

vijay120: only the `# system entity should be detected` tests cases:
```
1. `2am`
2. `4:30pm`
```

I added the other tests cases to make sure the other featurizers like `in-gaz-span-seq` work properly as well.

PR #136, opened 2020-04-13
Adding Github Action

minhtuev:
Github Actions are the default Github's support for continuous integration; recently we have seen instability with Travis CI.

- Create pythonpackage.yml


.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,55 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6]
```
vijay120, line 18: 3.7?

.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,55 @@
...
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+        java-version: '9.0.4' # The JDK version to make available on the path.
+        java-package: jdk # (jre, jdk, or jdk+fx) - defaults to jdk
+        architecture: x64 # (x64 or x86) - defaults to x64
+    
+    - name: Install dependencies
+      run: |
+        ./scripts/travis_install.sh
```
vijay120, line 37: why install travis ci when I assume github actions replicates it?

tmehlinger:
I agree with Vijay's observations. 😄 

I also know _zero_ about Github Actions so if it works, go nuts. At a glance, it looks fine to me.

.github/workflows/pythonpackage.yml
minhtuev: This is a poorly named script but basically it installs all of our dependency.

.github/workflows/pythonpackage.yml
minhtuev: We had a previous issue with enabling build for Python 3.7 in Travis CI, let me enable it here to see if it works.

minhtuev:
@tmehlinger @vijay120 : Py3.7 is taking too long to complete, let me remove it from the current build for now.

PR #138, opened 2020-04-15
Parametrize build with MM_EXTRAS

minhtuev:
1. Parametrize github action with environment MM_EXTRAS, which forces build to install extra dependency (tensorflow) and run tensorflow tests.

2. I also disable pytest in Travis - it has been unreliable for the past few months.

3. Add some lint fixes


minhtuev:
Tensorflow test failure is observed here:
https://github.com/cisco/mindmeld/pull/138/checks?check_run_id=635286388

PR #141, opened 2020-04-23
Adding a method to serialize Params and FrozenParams objects

minhtuev:
In order to serialize the Request object, we need to serialize Params and Frozen Params

### DIFF
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -126,9 +126,7 @@ class DialogueStateRule:
                     raise ValueError(msg.format(single, plural))
                 elif single in kwargs and isinstance(kwargs[single], str):
                     resolved[plural] = {kwargs[single]}
-                elif plural in kwargs and (
-                    isinstance(kwargs[plural], list) or isinstance(kwargs[plural], set)
-                ):
+                elif plural in kwargs and isinstance(kwargs[plural], (list, set)):
                     resolved[plural] = set(kwargs[plural])
                 else:
                     msg = "Invalid argument type for {!r}"
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -91,7 +91,7 @@ class DialogueStateRule:
             dialogue_state (str): The name of the dialogue state.
             domain (str): The name of the domain to match against.
             has_entity (str): A single entity type to match.
-            has_entities (str, list, set): A single entity type or a list/set of entity types to match
+            has_entities (list, set): A list/set of entity types to match
                 against.
             intent (str): The name of the intent to match against.
         """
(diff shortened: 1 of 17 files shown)


mindmeld/components/request.py
```diff
@@ -246,6 +246,32 @@ def nlp_params(self):
...
+        fields = [
+            "allowed_intents",
+            "target_dialogue_state",
+            "time_zone",
+            "timestamp",
+            "language",
+            "locale",
+            "dynamic_resource",
+        ]
+        _dic = self.nlp_params()
```
vijay120, line 259: Its unclear why you are using the `nlp_params ` here. The `nlp_params ` method is only used for validation of certain params attributes. Why don't we just iterates through all the object attributes and return them in a dictionary?

minhtuev: I see, I thought that the `nlp_params` is preferred method to retrieve nlp params, so I wanted to reuse that. I want to add this function to support serialization for both Params and FrozenParams object.

mindmeld/components/dialogue.py
```diff
@@ -120,7 +120,7 @@ def __init__(self, dialogue_state, **kwargs):
         resolved = {}
         for keys in key_kwargs:
             if len(keys) == 2:
-                single, plural = keys
+                single, plural = keys  # pylint: disable=unbalanced-tuple-unpacking
```
serapio, line 123: I realize this is incidental to your changes, but I hadn't realized we were doing this. In L130 we are assuming `single` is a string, and L132 assumes `plural` is an iterable of strings, but the doc string says `has_entity` and `has_entities` are both `(str, list, set)`

mindmeld/components/request.py
```diff
@@ -228,9 +228,10 @@ def dm_params(self, handler_map):
             return {"target_dialogue_state": None}
         return {"target_dialogue_state": target_dialogue_state}
 
-    def nlp_params(self):
+    def validate_nlp_params(self):
         """
-        Validate time zone, timestamp, and dynamic resource parameters.
+        Validate time zone, timestamp, and dynamic resource parameters and return the params as a
```
serapio, line 233: ... plus language and locale. Seems there should be a better conceptual description for the parameters grouped together here. And we should keep the method above with a parallel name (=> validate_dm_params)

mindmeld/components/dialogue.py
minhtuev: Ah, good eyes! Update docstring.

mindmeld/components/request.py
minhtuev: @vijay120 : what are the reasons for grouping these together?

mindmeld/components/dialogue.py
```diff
@@ -90,8 +90,8 @@ def __init__(self, dialogue_state, **kwargs):
         Args:
             dialogue_state (str): The name of the dialogue state.
             domain (str): The name of the domain to match against.
-            has_entity (str, list, set): A synonym for the ``has_entities`` param.
-            has_entities (str, list, set): A single entity type or a list of entity types to match
+            has_entity (str): A single entity type to match.
+            has_entities (str, list, set): A single entity type or a list/set of entity types to match
```
serapio, line 94: has_entities shouldn't be a string, because it will turn the string into a set of characters

minhtuev: Done

PR #144, opened 2020-05-05
Fix Allowed Intent Input Type

ritvikshrivastava:
The current docs mention the use of a list whereas internal validation is for tuple. Added list to the accepted type for passing in allowed intents.


minhtuev:
Should we update the logic to use both list and tuple?

ritvikshrivastava:
@minhtuev wasn't the idea to keep it a tuple to ensure immutability? I mean it can be converted internally, yeah.

serapio:
It seems like what we actually want is to leave `list` in the documentation for simplicity, and pass `converter=tuple` when constructing it. 
https://github.com/cisco/mindmeld/blob/2178a92acaa3008f7ef7ce80e0eaf319c3b1da46/mindmeld/components/request.py#L183

tests/components/test_nlp.py
```diff
@@ -617,7 +617,7 @@ def test_sys_entity_feature(kwik_e_mart_nlp):
 
 
 test_data_dyn = [
-    ("kadubeesanahalli", None, "store_info", "", ""),
+    ("kadubeesanahalli", None, "store_info", "get_store_hours", ""),
```
vijay120, line 620: This test change seems unrelated to the code change.

ritvikshrivastava: It's in reference to a failing test for all PRs (refer to Minh's comment above). Should I move it to a separate PR?

PR #145, opened 2020-05-07
Adding CustomAction and custom action handler in Application

minhtuev:
This is a PR to add CustomAction and its handler in a MindMeld application.

Essentially we are allowing a 3rd party to handle a MindMeld dialogue turn.

Example:
```
app.custom_action(intent='fetch_info', action=''fetch_info_action')
```

We also support asynchronous requests as well:
```
app.custom_action(intent='fetch_info', action=''fetch_info_action', async=True)
```

Whenever the user hits `fetch_info` intent, the Request and Responder objects are serialized and sent as payload to the custom action URL. The resulting frame and directives are returned as a property of the Responder object.

We do not need to put full attention on the RasaConverter; this PR includes the RasaConverter to also include the usage of CustomAction.

OpenAPI specification: https://app.apiary.io/mindmeldnlp/

### DIFF
--- a/mindmeld/app.py
+++ b/mindmeld/app.py
@@ -184,25 +184,16 @@ class Application:
                 "`action` or `actions` must be present in arguments."
             )
 
-        if action:
-            self.add_custom_action(
-                action,
-                asynch=kwargs.pop("asynch", False),
-                overwrite=kwargs.pop("overwrite", False),
-                config=kwargs.pop("config", None),
-                **kwargs
-            )
-        else:
-            self.add_custom_action_sequence(
-                actions,
-                asynch=kwargs.pop("asynch", False),
-                overwrite=kwargs.pop("overwrite", False),
-                config=kwargs.pop("config", None),
-                **kwargs
-            )
+        self.add_custom_action_sequence(
+            [action] if action else actions,
+            async_mode=kwargs.pop("async_mode", False),
+            overwrite=kwargs.pop("overwrite", False),
+            config=kwargs.pop("config", None),
+            **kwargs
+        )
 
     def add_custom_action(
-        self, action, asynch=False, overwrite=False, config=None, **kwargs
+        self, action, async_mode=False, overwrite=False, config=None, **kwargs
     ):
         """Adds a custom action handler for the dialogue manager.
 
@@ -211,7 +202,7 @@ class Application:
 
         Args:
             action (str): The name of the custom action
-            asynch (bool): Whether we should invoke this custom action asynchronously
+            async_mode (bool): Whether we should invoke this custom action asynchronously
             overwrite (bool): Whether we should overwrite the Responder with fields from the
                 response, otherwise we will extend the fields (frame, directives) accordingly.
             config (dict): The custom action config, if different from the application's.
@@ -225,13 +216,13 @@ class Application:
 
         custom_action = CustomAction(action, config, overwrite=overwrite)
         state_name = kwargs.pop("name", "custom_action_{}".format(action))
-        if asynch:
+        if async_mode:
             self.add_dialogue_rule(state_name, custom_action.invoke_async, **kwargs)
         else:
             self.add_dialogue_rule(state_name, custom_action.invoke, **kwargs)
 
     def add_custom_action_sequence(
-        self, actions, asynch=False, overwrite=False, config=None, **kwargs
+        self, actions, async_mode=False, overwrite=False, config=None, **kwargs
     ):
         """Adds a custom action sequence handler for the dialogue manager.
 
@@ -240,7 +231,7 @@ class Application:
 
         Args:
             actions (list): A list of custom actions
-            asynch (bool): Whether we should invoke this custom action asynchronously
+            async_mode (bool): Whether we should invoke this custom action asynchronously
             overwrite (bool): Whether we should overwrite the Responder with fields from the
                 response, otherwise we will extend the fields (frame, directives) accordingly.
             config (dict): The custom action config, if different from the application's.
@@ -254,7 +245,7 @@ class Application:
 
         action_seq = CustomActionSequence(actions, config, overwrite=overwrite)
         state_name = kwargs.pop("name", "custom_actions_{}".format(actions))
-        if asynch:
+        if async_mode:
             self.add_dialogue_rule(state_name, action_seq.invoke_async, **kwargs)
         else:
             self.add_dialogue_rule(state_name, action_seq.invoke, **kwargs)
--- a/mindmeld/app.py
+++ b/mindmeld/app.py
@@ -21,7 +21,11 @@ import sys
 from .app_manager import ApplicationManager
 from .cli import app_cli
 from .components._config import get_custom_action_config
-from .components.custom_action import CustomAction, CustomActionException
+from .components.custom_action import (
+    CustomAction,
+    CustomActionException,
+    CustomActionSequence,
+)
 from .components.dialogue import DialogueFlow, DialogueResponder, AutoEntityFilling
 from .components.request import Request
 from .server import MindMeldServer
@@ -172,15 +176,30 @@ class Application:
             app.custom_action(intent='greeting', action='say_greeting')
             app.custom_action(entity='person', action='greet_person')
         """
-        if "action" not in kwargs:
-            raise CustomActionException("`action` is a required argument.")
-        self.add_custom_action(
-            kwargs.pop("action"),
-            asynch=kwargs.pop("asynch", False),
-            overwrite=kwargs.pop("overwrite", False),
-            config=kwargs.pop("config", None),
-            **kwargs
-        )
+        action = kwargs.pop("action", None)
+        actions = kwargs.pop("actions", None)
+
+        if not (action or actions):
+            raise CustomActionException(
+                "`action` or `actions` must be present in arguments."
+            )
+
+        if action:
+            self.add_custom_action(
+                action,
+                asynch=kwargs.pop("asynch", False),
+                overwrite=kwargs.pop("overwrite", False),
+                config=kwargs.pop("config", None),
+                **kwargs
+            )
+        else:
+            self.add_custom_action_sequence(
+                actions,
+                asynch=kwargs.pop("asynch", False),
+                overwrite=kwargs.pop("overwrite", False),
+                config=kwargs.pop("config", None),
+                **kwargs
+            )
 
     def add_custom_action(
         self, action, asynch=False, overwrite=False, config=None, **kwargs
@@ -210,6 +229,35 @@ class Application:
             self.add_dialogue_rule(state_name, custom_action.invoke_async, **kwargs)
         else:
             self.add_dialogue_rule(state_name, custom_action.invoke, **kwargs)
+
+    def add_custom_action_sequence(
+        self, actions, asynch=False, overwrite=False, config=None, **kwargs
+    ):
+        """Adds a custom action sequence handler for the dialogue manager.
+
+        Whenever the user hits this state, we invoke the sequence of custom actions and returns
+            the appropriate responder.
+
+        Args:
+            actions (list): A list of custom actions
+            asynch (bool): Whether we should invoke this custom action asynchronously
+            overwrite (bool): Whether we should overwrite the Responder with fields from the
+                response, otherwise we will extend the fields (frame, directives) accordingly.
+            config (dict): The custom action config, if different from the application's.
+        """
+        if not actions:
+            raise CustomActionException("Argument `actions` should not be empty.")
+
+        config = config or self.custom_action_config
+        if not config:
+            raise CustomActionException("Argument `config` should not be empty.")
+
+        action_seq = CustomActionSequence(actions, config, overwrite=overwrite)
+        state_name = kwargs.pop("name", "custom_actions_{}".format(actions))
+        if asynch:
+            self.add_dialogue_rule(state_name, action_seq.invoke_async, **kwargs)
+        else:
+            self.add_dialogue_rule(state_name, action_seq.invoke, **kwargs)
 
     def dialogue_flow(self, **kwargs):
         """Creates a dialogue flow for the application"""
--- a/mindmeld/app.py
+++ b/mindmeld/app.py
@@ -200,7 +200,7 @@ class Application:
                 "There is no configuration specified for this action."
             )
 
-        actions = [action] or actions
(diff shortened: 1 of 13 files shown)


mindmeld/app.py
```diff
@@ -161,6 +165,36 @@ def add_dialogue_rule(self, name, handler, **kwargs):
...
+        of nlp entities to be more flexible.
+
+        Examples:
+            app.custom_action(intent='greeting', action='say_greeting')
+            app.custom_action(entity='person', action='greet_person')
+        """
+        action = kwargs.get("action")
+        if not action:
+            raise CustomActionException("`action` is a required argument.")
+        self.add_custom_action(action, kwargs.get("asynch", False), **kwargs)
```
arushir, line 179: why not expose only this method and remove `add_custom_action`?

arushir:
@minhtuev The [link](https://app.apiary.io/mindmeldnlp/) is giving me a `404: The API Project you have requested could not be found.`

Can you describe a bit more on the goals for this PR and the plan for future changes?
1. It seems like these changes will only handle on-shot requests. How will support for multi-turn dialogues be added?
2. All of the custom actions are sending requests to the same URL with the full NLP info, so what is the point of allowing users to add multiple custom actions for different domain/intent/entity combinations? Would it not be simpler for that logic to be done in the 3rd party code in the developers language of choice?

mindmeld/app.py
```diff
@@ -161,6 +168,97 @@ def add_dialogue_rule(self, name, handler, **kwargs):
...
+    def custom_action(self, **kwargs):
+        """Adds a custom action, similar to `add_custom_action` but allows the ordering
+        of nlp entities to be more flexible.
+
+        Examples:
+            app.custom_action(intent='greeting', action='say_greeting')
+            app.custom_action(entity='person', action='greet_person')
+        """
+        action = kwargs.pop("action", None)
+        actions = kwargs.pop("actions", None)
```
vijay120, line 180: Why cant we just use one field called actions where it can be a list of 1 or many?

mindmeld/app.py
```diff
@@ -161,6 +168,97 @@ def add_dialogue_rule(self, name, handler, **kwargs):
...
+            self.add_custom_action_sequence(
+                actions,
+                asynch=kwargs.pop("asynch", False),
+                overwrite=kwargs.pop("overwrite", False),
+                config=kwargs.pop("config", None),
+                **kwargs
+            )
+
+    def add_custom_action(
+        self, action, asynch=False, overwrite=False, config=None, **kwargs
```
vijay120, line 205: asynch -> async

mindmeld/app.py
```diff
@@ -161,6 +168,97 @@ def add_dialogue_rule(self, name, handler, **kwargs):
...
+    ):
+        """Adds a custom action handler for the dialogue manager.
+
+        Whenever the user hits this state, we invoke the custom action instead and returns
+            the appropriate responder.
+
+        Args:
+            action (str): The name of the custom action
+            asynch (bool): Whether we should invoke this custom action asynchronously
+            overwrite (bool): Whether we should overwrite the Responder with fields from the
```
vijay120, line 215: So if overwrite is set of False, and there is a key collision, we just use the key in the original application (and not the custom action)?

mindmeld/app.py
```diff
@@ -161,6 +168,97 @@ def add_dialogue_rule(self, name, handler, **kwargs):
...
+
+        custom_action = CustomAction(action, config, overwrite=overwrite)
+        state_name = kwargs.pop("name", "custom_action_{}".format(action))
+        if asynch:
+            self.add_dialogue_rule(state_name, custom_action.invoke_async, **kwargs)
+        else:
+            self.add_dialogue_rule(state_name, custom_action.invoke, **kwargs)
+
+    def add_custom_action_sequence(
+        self, actions, asynch=False, overwrite=False, config=None, **kwargs
```
vijay120, line 234: Same, async

mindmeld/components/custom_action.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def get_json_payload(self, request, responder):
+        request_json = {
+            "text": request.text,
+            "domain": request.domain,
+            "intent": request.intent,
+            "context": dict(request.context),
+            "params": request.params.to_dict(),
+            "frame": dict(request.frame),
+        }
```
vijay120, line 39: Lets just make a `to_dict` function for the `Request` class so that we dont have to do this everytime.

mindmeld/components/custom_action.py
```diff
@@ -0,0 +1,187 @@
...
+import aiohttp
+import requests
+
+from .dialogue import DialogueResponder
+
+
+logger = logging.getLogger(__name__)
+
+
+RESPONSE_FIELDS = ["frame", "directives", "params"]
```
vijay120, line 12: Why not include all fields in `DialogueResponder` like `slots`, `dialogue_state` and `history`?

mindmeld/components/custom_action.py
```diff
@@ -0,0 +1,187 @@
...
+                    logger.warning(
+                        "`%s` not in the response of custom action %s",
+                        field,
+                        self._name,
+                    )
+            if self.overwrite:
+                responder.frame = result_json.get("frame", {})
+                responder.directives = result_json.get("directives", [])
+            else:
+                responder.frame.update(result_json.get("frame", {}))
```
vijay120, line 114: update overwrites the existing key. I dont think this is the intention when supplying overwrite=False.

vijay120: There is no population of the params field here.

mindmeld/components/custom_action.py
```diff
@@ -0,0 +1,187 @@
...
+    def invoke(self, request, responder):
+        for action in self.actions:
+            result = action.invoke(request, responder)
+            if not result:
+                logger.warning("Failed to invoke action %s.", action)
+                return False
+        return True
+
+    async def invoke_async(self, request, responder):
+        for action in self.actions:
```
vijay120, line 166: You can do this:
```
    await asyncio.gather(
        [action.invoke_async(request, responder) for action in self.actions]
    )
```

mindmeld/models/model.py
```diff
@@ -1295,9 +1295,10 @@ def decode(self, tags_by_example, **kwargs):
             list: A list of decoded labels per query
         """
         # TODO: support decoding multiple queries at once
+        system_entity_recognizer = SystemEntityRecognizer.get_instance()
         examples = kwargs["examples"]
         labels = [
-            get_entities_from_tags(examples[idx], tags, self.system_entity_recognizer)
+            get_entities_from_tags(examples[idx], tags, system_entity_recognizer)
```
vijay120, line 1301: Why are we doing this? Looks like `self.system_entity_recognizer` is defined.

mindmeld/app.py
minhtuev: `async` is a Python keyword so you can't use it; it can be either `asynch` or `async_mode`. I chose `asynch` because it is shorter.

mindmeld/app.py
minhtuev: same as above

mindmeld/components/custom_action.py
minhtuev: Because I don't want the requests to be too large.

mindmeld/app.py
serapio: elsewhere in the code we have been using `async_mode`. I think it's better to be consistent with that.

minhtuev: Done

mindmeld/app.py
minhtuev: It is not grammatically correct since `action` is singular. It is also consistent with how we use the `handle` function (`entity` vs `entities`)

mindmeld/components/custom_action.py
minhtuev: Done

mindmeld/components/custom_action.py
minhtuev: You can't log the warning if the action fails to execute

mindmeld/app.py
minhtuev: Mainly, the difference between `overwrite=True` and `overwrite=False` was initially conceived for the directives. Suppose the 3rd party server returns four directives, while the initial responder already has two. With `overwite=True`, we will overwrite the responder's directive field with four new directives. With `overwrite=False`, we will simply append the four new directives to the responder, so we will have a total of six directives.

For the frame, currently, it is a similar idea, when `overwrite=False`, we will "update" the dictionary with new keys and values. If there is a collision, we assume that it is intended by the user since we also send them that `frame` as well.

mindmeld/components/custom_action.py
minhtuev: We assume that if the user changes the key value in the frame that we send them, they intend for that change to effect. The `overwrite` field distinguishes the case where the user wants the returned frame to be a "hard" change (as it is) or a soft change (merging and updating with any existing key/value).

mindmeld/components/custom_action.py
vijay120: We should atleast include `slots ` and `dialogue_state `, as discussed in the design meeting where we concluded that these fields are important for slot filling and multi-turn dialogue, two features important for the client requesting this.

mindmeld/components/custom_action.py
vijay120: What about the common case where the user wants to only update keys that he/she changed, and not any other keys? We could accomodate this by doing the following:
1. If overwrite is True and user provides keys to change, we overwrite those keys that the user wants to change. If they want to overwrite ALL the keys (as you mentioned), they can also do that by supplying all the keys to overwrite.
2. If overwrite is False and user provides keys whose values collide with existing keys, those collision keys are NOT updated.

I believe the above behavior is the usual definition of what is means to "override" a key or not.

mindmeld/models/model.py
vijay120: This issue is still not addressed.

mindmeld/app.py
vijay120: If the directives were passed to the 3rd party service (as we concluded in the design meeting), then the user can simply append/mutate their directives to the existing set of directive, set  `overwrite=True` and expect the DM to pick up these changes. If  `overwrite=False`, I wouldn't expect the `directive` key in the DM to change by the 3rd party service.

mindmeld/components/custom_action.py
minhtuev: I see, sure. Make sure that the API doc is reviewed since it is the source of truth.

minhtuev:
@vijay120 : so the behavior is as followed:
1. When `merge=True`, we will merge the fields (directives, frame, slots, params) in this manner:
- If there is an existing key, we will update it with a new value from the response.
- If there is not an existing key, we will add it to the dictionary.
2. When `merge=False`, we will use the fields (directives, frame, slots, params) as it is returned.

minhtuev:
@tmehlinger @mhuang2173 : if you can take a look at the OpenAPI doc and the section for configuring SSL support, it will be great.

mindmeld/app.py
```diff
@@ -168,82 +167,41 @@ def add_dialogue_rule(self, name, handler, **kwargs):
...
 
         config = config or self.custom_action_config
         if not config:
-            raise CustomActionException("Argument `config` should not be empty.")
+            raise CustomActionException(
+                "There is no configuration specified for this action."
+            )
 
-        action_seq = CustomActionSequence(actions, config, overwrite=overwrite)
+        actions = [action] or actions
```
serapio, line 203: this doesn't do what you meant

mindmeld/components/_config.py
```diff
@@ -472,6 +472,19 @@ class NlpConfigError(Exception):
...
+def get_custom_action_config(app_path):
+    if not app_path:
+        return None
+    try:
+        custom_action_config = getattr(
+            _get_config_module(app_path), "CUSTOM_ACTION_CONFIG", None
+        )
+        return custom_action_config
+    except (OSError, IOError):
+        logger.info("No app configuration file found.")
```
vijay120, line 484: `logger.error`

mindmeld/components/custom_action.py
```diff
@@ -0,0 +1,202 @@
...
+    This class allows the client to send Request and Responder to another server and return the
+      the directives and frame in the response.
+    """
+
+    def __init__(self, name: str, config: dict, merge: bool = True):
+        self._name = name
+        self._config = config or {}
+        self.url = self._config.get("url")
+        self._cert = self._config.get("cert")
+        self._public_key = self._config.get("public_key")
```
vijay120, line 32: Fantastic

mindmeld/components/custom_action.py
```diff
@@ -0,0 +1,202 @@
...
+                    )
+            if self.merge:
+                responder.frame.update(result_json.get("frame", {}))
+                responder.directives.extend(result_json.get("directives", []))
+                responder.slots.update(result_json.get("slots", {}))
+            else:
+                responder.frame = result_json.get("frame", {})
+                responder.directives = result_json.get("directives", [])
+                responder.slots = result_json.get("slots", {})
+            responder.params = Params(**result_json.get("params", {}))
```
vijay120, line 116: Why is `responder.params = Params(**result_json.get("params", {}))` done when `self.merge == True`? I would also update the existing params if possible, no?

mindmeld/openapi/custom_action.yaml
```diff
@@ -0,0 +1,187 @@
...
+        type: string
+        description: "The language of the request"
+      locale:
+        type: string
+        description: "The locale of the request"
+      timestamp:
+        type: string
+        description: "The timestamp of the request"
+
+  Responder:
```
vijay120, line 166: Can we please document these fields based on the descriptions listed here: https://www.mindmeld.com/docs/userguide/dm.html?highlight=directive#responder. The current documentation does not explain to the developer what the objects do.

minhtuev: This is the documentation for the API, not the documentation for the Responder so I am not sure if we should include all the descriptions from the Responder documentation. I think we should include the link for the documentation and that might be enough.

mindmeld/components/custom_action.py
minhtuev: Done

mindmeld/app.py
minhtuev: Done

PR #147, opened 2020-05-07
Add option to use deep embedders for QA

arushir:
This PR adds support for the question answering with deep semantic embedding vectors.

**Note: This PR also updates MindMeld to ES 7**

If a developer would like to use a built in embedder (currently provide implementation for [sentence bert](https://arxiv.org/abs/1908.10084) and [glove](https://nlp.stanford.edu/projects/glove/)), they can use the new question answerer config to specify the details. 

```
QUESTION_ANSWERER_CONFIG = {
    "model_type": "embedder",
    "model_settings": {
        "embedder_type": "bert",
        "trained_data": "bert-base-nli-mean-tokens",
        "embedding_fields": {"wxa_faq": ['question', 'answer']}  # list of fields to generate embeddings for each index
    }
}
```

When loading the kb they will need to pass in the `app-path` to access the config. At load time, the embedding fields will be ingested into ElasticSearch as dense vectors. 

```
mindmeld load-kb faq_embedding wxa_faq data/wxa/faq_data.jsonl --app_path ./
```

If a developer would like to use their own embedder, they can use the provided abstract class and implement the `load` and `encode` methods. The `model_settings` dict in the config will be passed to the load method, so any needed parameters can be specified there. The class can be used after registering via the `register_embedder` method.
```
from mindmeld.models import Embedder, register_embedder

class MyEmbedder(Embedder):
    def load(self, **kwargs):
        # return the loaded model
    def encode(self, text_list):
        # encodes each query in the list 

register_embedder('my_embedder_name', MyEmbedder)
```
```
QUESTION_ANSWERER_CONFIG = {
    "model_type": "embedder",
    "model_settings": {
        "embedder_type": "my_embedder_name",
        "some_model_param": "my_model_param",
        "embedding_fields": {"wxa_faq": ['question', 'answer']}
    }
}
```

Developers can use the `query_type = embedder` parameter with the question answerer, and it will retrieve the documents with the closest cosine similarity to the embedding. 

```
answers = qa.get(index=index_name, query_type='embedder', 
                 question=query)
```

Developers can also do the search using multiple embedding fields or even using a combination of embedder and text based search. 

```
    answers = qa.get(index=index_name, query_type='embedder_text',
                     question=query,
                     answer=query)
``` 

See this [doc](https://cisco-my.sharepoint.com/:w:/p/arushi/EVnPxSKkDSdEnkj1sbhxBQ4BzSjyAcyjdSlwOU8287yF3A?e=66K6Ou) for more details.


arushir:
@mhuang2173 

1. The `embedding_fields` will always be configured in the `model_settings` for both built in and custom embedders. For custom embedders, developers can optionally pass additional key, value pairs in `model_settings` to be passed as kwargs to the custom load() method. Do you think this makes sense, or would you recommend updating the config in some way?

2. Right now we need the `_embedding` suffix, since we give users the option to do a text search, an embedding search, or both. When searching with the `question` field it does the usual text search and when searching with the `question_embedding` field it does the embedding search. That way they can use signals from both text and embedding in a single search. The other option would be to introduce a fourth query type so we would have `keyword`, `text`, `embedding`, `text_and_embedding` and then we wouldn't need the `_embedding` suffix. Would that make more sense?

mindmeld/components/dialogue.py
```diff
@@ -129,7 +129,9 @@ def __init__(self, dialogue_state, **kwargs):
                     raise ValueError(msg.format(single, plural))
                 elif single in kwargs and isinstance(kwargs[single], str):
                     resolved[plural] = {kwargs[single]}
-                elif plural in kwargs and isinstance(kwargs[plural], (list, set, tuple)):
+                elif plural in kwargs and isinstance(
```
minhtuev, line 132: I think this file is just applying black right?

mindmeld/models/embedder_models.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def load(self, **kwargs):
+        DEFAULT_BERT = "bert-base-nli-mean-tokens"
+        if "trained_data" in kwargs:
+            return SentenceTransformer(kwargs["trained_data"])
+        else:
+            logger.warning("No bert model specifications passed, using default.")
+            return SentenceTransformer(DEFAULT_BERT)
+
+    def encode(self, text_list):
```
minhtuev, line 122: is there any requirement for the list of texts? any length limit?

minhtuev: on another note can this model run in parallel?

setup.py
```diff
@@ -12,7 +12,7 @@
     "Click~=6.0",
     "click-log==0.1.8",
     "distro~=1.3",
-    "elasticsearch5~=5.5",
+    "elasticsearch>=7.0",
```
minhtuev, line 15: does this change make our previous version of ES backward-incompatible?

mindmeld/models/embedder_models.py
```diff
@@ -0,0 +1,127 @@
...
+        with open(self.cache_path, "wb") as fp:
+            pickle.dump(self.cache, fp)
+
+
+class BertEmbedder(Embedder):
+    """
+    Encoder class for bert models as described here: https://github.com/UKPLab/sentence-transformers
+    """
+
+    def load(self, **kwargs):
```
minhtuev, line 114: maybe we should add a way for the developers to use their own custom BERT model?

```
# Use BERT for mapping tokens to embeddings
word_embedding_model = models.Transformer('path/to/your/BERT/model')

# Apply mean pooling to get one fixed sized sentence vector
pooling_model = models.Pooling(word_embedding_model.get_word_embedding_dimension(),
                               pooling_mode_mean_tokens=True,
                               pooling_mode_cls_token=False,
                               pooling_mode_max_tokens=False)

model = SentenceTransformer(modules=[word_embedding_model, pooling_model])

File: mindmeld/models/embedder_models.py [hunk 5]
```diff
@@ -0,0 +1,127 @@
...
+        DEFAULT_BERT = "bert-base-nli-mean-tokens"
+        if "trained_data" in kwargs:
```
minhtuev, line 116: `model_name` might be a clearer name?

```suggestion
        if "model_name" in kwargs:
```

mindmeld/components/question_answerer.py
```diff
@@ -304,18 +367,41 @@ def transform(doc):
                     app_namespace,
                 )
 
+        def _generate_mapping_data(embedder_model, embedding_fields):
+            # generates a dictionary with any metadata needed to create the mapping"
+            MAX_ES_VECTOR_LEN = 2048
```
minhtuev, line 372: is this an ES hard limit?

mindmeld/components/question_answerer.py
```diff
@@ -304,18 +367,41 @@ def transform(doc):
...
+            embedding_properties = []
+            mapping_data = {"embedding_properties": embedding_properties}
+
+            dims = len(embedder_model.get_encodings(["encoding"])[0])
+            if dims > MAX_ES_VECTOR_LEN:
```
minhtuev, line 377: is there a limit on the size of the vector itself (in terms of memory)?

mindmeld/components/question_answerer.py
```diff
@@ -100,9 +121,11 @@ def get(self, index, size=10, query_type="keyword", **kwargs):
         Returns:
             list: A list of matching documents.
         """
-
         doc_id = kwargs.get("id")
 
+        if not query_type:
+            query_type = self._query_type
```
minhtuev, line 127: We can follow this pattern for concision :)

```suggestion
            query_type = query_type or self._query_type
```

mindmeld/components/question_answerer.py
```diff
@@ -256,7 +283,26 @@ def load_kb(
                 connection to the Elasticsearch host.
             clean (bool): Set to true if you want to delete an existing index
                 and reindex it
+            app_path (str): The path to the directory containing the app's data
         """
+        embedder_model = None
```
minhtuev, line 288: maybe we should refactor this into a `load_emebedder_model` function

minhtuev:
I have some comments, but at the high level is this change backward compatible with ES5 and below?

mindmeld/components/dialogue.py
arushir: That's right, fixing lint errors that weren't caught before

mindmeld/models/embedder_models.py
arushir: I haven't explicitly set any length limits, but the max that will be passed will be the number of fields to generate embeddings for, so I don't think one is needed. The encode() method will pass all provided fields through the network at once so it generates the embeddings in parallel rather than looping over them.

setup.py
arushir: If an app developer has an existing index in an older version of ES, we should recommend that they do a clean build to recreate those indices with ES7. Otherwise, all exposed ES functionality via MindMeld methods will continue to work the same way.

mindmeld/models/embedder_models.py
arushir: Developers can do this by implementing their own embedder inheriting from the abstract class:

from mindmeld.models import Embedder, register_embedder

class MyEmbedder(Embedder):
    def load(self, **kwargs):
        # return the loaded model
    def encode(self, text_list):
        # encodes each query in the list 

register_embedder('my_embedder_name', MyEmbedder)

I think it's cleaner to have them use this functionality rather than supporting too much variation in the built in version. But let me know if you think otherwise.

mindmeld/components/question_answerer.py
arushir: I believe so: https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html

mindmeld/components/question_answerer.py
arushir: Not sure I understand - how are memory and length limits different here?

mindmeld/components/question_answerer.py
arushir: will take a look

arushir:
@minhtuev If an app developer has an existing index in an older version of ES, they would need to do a clean build to recreate those indices with ES7. Otherwise, all exposed ES functionality via MindMeld methods will continue to work the same way.

mindmeld/models/embedder_models.py
arushir: Yeah I wasn't sure what the best naming would be. I had initially decided against using `model_` since that is already being used for `model_type` and `model_settings` which is talking about the embedder model rather than the specific trained Bert model. Now looking at it again, I think we should have a `model_name` parameter in the config so users can name the different versions of the same model for the cache lookup, but trying to decide if that should be reused here or be a more generic parameter.

mindmeld/components/question_answerer.py
arushir: fixed

mindmeld/components/question_answerer.py
arushir: fixed

mindmeld/models/embedder_models.py
arushir: fixed

mindmeld/models/embedder_models.py
minhtuev: Ah got it, so the developer does not need to explicit set the device (CPU/GPU/TPU etc)?

arushir: Oh yeah, we are not explicitly setting the device here. Based on the developers system and settings, if CUDA is available GPU should be used by default

mindmeld/components/_config.py
```diff
@@ -102,97 +103,91 @@
...
     "language": ENGLISH_LANGUAGE_CODE,
     "locale": ENGLISH_US_LOCALE,
 }
 
-DOC_TYPE = "document"
 
 # ElasticSearch mapping to define text analysis settings for text fields.
 # It defines specific index configuration for synonym indices. The common index configuration
 # is in default index template.
 DEFAULT_ES_SYNONYM_MAPPING = {
```
vijay120, line 119: Is it possible to prevent deprecation by having one such mapping for <ES 7.0 and one for 7.0?

mindmeld/components/_elasticsearch_helpers.py
```diff
@@ -223,13 +223,29 @@ def delete_index(
         raise KnowledgeBaseError
 
 
+def create_index_mapping(base_mapping, mapping_data):
```
vijay120, line 226: this function is not being used anywhere

mindmeld/components/_elasticsearch_helpers.py
```diff
@@ -15,14 +15,14 @@
 import logging
 import os
 
-from elasticsearch5 import ConnectionError as EsConnectionError
-from elasticsearch5 import (
+from elasticsearch import ConnectionError as EsConnectionError
```
vijay120, line 18: Is the `elasticsearch ` lib for ES 7 or does it work across all ES versions?

mindmeld/components/_elasticsearch_helpers.py
```diff
@@ -107,7 +107,7 @@ def get_field_names(
             )
 
         res = es_client.indices.get(index=scoped_index_name)
-        all_field_info = res[scoped_index_name]["mappings"]["document"]["properties"]
+        all_field_info = res[scoped_index_name]["mappings"]["properties"]
```
vijay120, line 110: Is this specific for ES7 or for all ES versions?

mindmeld/models/embedder_models.py
```diff
@@ -0,0 +1,185 @@
...
+        query = query.translate(str.maketrans("", "", string.punctuation))
+        tokens = query.split()
+        tokens = [t.strip() for t in tokens]
+        return tokens
+
+    def encode(self, text_list):
+        token_list = [self.get_query_tokens(text) for text in text_list]
+        vector_list = [self.model.encode_sequence_of_tokens(tl) for tl in token_list]
+        encoded_vecs = []
+        for vl in vector_list:
```
vembar, line 168: Is this check necessary? Wont the axis make sure it leaves a list with a single list as is?

source/userguide/kb.rst
```diff
@@ -813,3 +813,86 @@ We can perform the same search using the :meth:`query()` API as well.
 	To delete and reindex your data, follow the steps mentioned :ref:`here <import_kb>`.
 
 
+.. _semantic_embeddings:
+
+Leveraging semantic embeddings
+------------------------------
+Up until now, the described Question Answering relies on text based retrieval. Deep learning based dense embeddings (character, word, or sentence) are in many cases better at capturing semantic information than traditional sparse vectors. Pretrained or fine-tuned embeddings can be used to find the best match in the Knowledge Base even if the search token wasn’t present in the uploaded data.
+
+To leverage semantic embeddings in search, the first step is to store the embeddings in your Knowledge Base. You can decide to use one of the provided embedders, or you can use your own. If you app consists of mainly standard English vocabulary, one of provided embedders may work well enough, but if the text you are searching against has quite a bit of domain specific vocabulary, you may benefit from training or fine tuning your own embedder on your data.
```
vembar, line 822: Typo in the 3rd sentence: "If your app .."

source/userguide/kb.rst
```diff
@@ -813,3 +813,86 @@ We can perform the same search using the :meth:`query()` API as well.
...
+          "trained_data": "bert-base-nli-mean-tokens",
+          "embedding_fields": ['question', 'answer']
+      }
+  }
+
+Here is the full list of included models. 
+
+# TODO: add table.
+
+One your config file is specified, you can load the Knowledge Base with the embeddings by passing in the flag ``--generate-embeddings`` or ``-g`` to the ``load-kb`` CLI command. Or the parameter ``generate_embeddings=True`` to the ``load_kb`` method. This will generate embeddings for each of the config specified fields and adds them to your json file before loading.
```
vembar, line 841: "Once your config file is specified..."

mindmeld/components/_config.py
arushir: Yeah good point, will see how we can make this backwards compatible.

mindmeld/components/_elasticsearch_helpers.py
arushir: It's being used in question_answerer.py

mindmeld/components/_elasticsearch_helpers.py
arushir: It works across all versions, but you need to have your python client match versions with your ES client: https://elasticsearch-py.readthedocs.io/en/master/ 

I think we should recommend that everyone installs ES7 and the matching client in the documentation and requirements, but need to think more about backwards compatibility.

mindmeld/components/_elasticsearch_helpers.py
arushir: It's needed for a breaking change that was made in ES7. I need to check if this version will also work with previous versions of ES

mindmeld/models/embedder_models.py
arushir: It was converting a single list to a single numeric value rather than keeping the vector as is. Is there a way to force it to leave it as is?

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
 	To delete and reindex your data, follow the steps mentioned :ref:`here <import_kb>`.
 
 
+.. _semantic_embeddings:
+
+Leveraging semantic embeddings
+------------------------------
+Up until now, the described Question Answering relies on text based retrieval. Deep learning based dense embeddings (character, word, or sentence) are in many cases better at capturing semantic information than traditional sparse vectors. Pretrained or fine-tuned embeddings can be used to find the best match in the Knowledge Base even if the search token wasn’t present in the uploaded data.
+
+To leverage semantic embeddings in search, the first step is to generate the embeddings for your desired Knowledge Base fields. You can decide to use one of the provided embedders, or you can use your own. If you app consists of mainly standard English vocabulary, one of provided embedders may work well enough, but if the text you are searching against has quite a bit of domain specific vocabulary, you may benefit from training or fine tuning your own embedder on your data.
```
ktick, line 822:
```suggestion
To leverage semantic embeddings in search, the first step is to generate the embeddings for your desired knowledge base fields. You can use one of the provided embedders or use your own. If your app mainly consists of standard English vocabulary, one of provided embedders may work well enough, but if the text you are searching against has quite a bit of domain specific vocabulary, you may benefit from training or fine tuning your own embedder on your data.
```

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+
+To use one of the built in embedders, you can simply specify an emedder ``model_type`` in your Question Answering config file, specify which model you would like to use ``model_settings``, and specify which fields you would like to generate embeddings for in ``embedding_fields``. The embedding fields parameter takes a dictionary where the key is the name of your index, and the value is a list of field names or regexes to match against the field names for that index. Using the HR Assistant blueprint as an example, here is what the config would look like.
```
ktick, line 824:
```suggestion
The settings for semantic embeddings are part of the ``QUESTION_ANSWERER_CONFIG`` in the app configuration file, ``config.py``. To use semantic embeddings, you need to specify a supported ``model_type``,  the ``model_settings``, and the fields you would like to generate embeddings for in ``embedding_fields``. The ``embedding_fields`` parameter takes a dictionary where the key is the name of your index, and the value is a list of field names or regexes to match against the field names for that index.

Using the HR Assistant blueprint as an example, here is what the question answerer config could look like.
```

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+
+  +-----------------------+------------------------------------------------------------------------------------------------------------+
+  | Model type            | Description                                                                                                |
+  +=======================+============================================================================================================+
+  | embedder              | For leveraging deep semantic embedders. This option allows for using deep semantic embedders like          |
+  |                       | sentence bert for doing vector-based retrieval                                                             |
+  +-----------------------+------------------------------------------------------------------------------------------------------------+
+  | embedder_keyword      | For leveraging deep semantic embedders and text signals for structured search on keywords or short spans   |
+  |                       | of text. Glove as an additional signal has shown improvements in this use case.                            |
```
ktick, line 847:
```suggestion
  |                       | of text. GloVe as an additional signal has shown improvements in this use case.                            |
```

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+          "embedding_fields": {
+              "faq_data": ["question", "answer"]
+          }
+      }
+  }
+
+There are three available model types which leverage semantic embedders. The specified ``model_type`` is also the default ``query_type`` for all question answering calls for your application, but you can always pass in a different query type to your ``qa.get()`` command as desired.
+
+  +-----------------------+------------------------------------------------------------------------------------------------------------+
+  | Model type            | Description                                                                                                |
```
ktick, line 841: Could you change the descriptions to:

Only leverage deep semantic embedders. This option allows for using deep semantic embedders like Sentence-BERT or GloVe for doing vector-based retrieval.

Leverage a combination of deep semantic embedders and text signals in a way that's optimized for search on keywords or short spans of text. GloVe is preferable for this use case.

Leverage a combination of deep semantic embedders and text signals in a way that's optimized for search on larger paragraphs or passages of unstructured text. Sentence-BERT is preferable for this use case.

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+  +-----------------------+------------------------------------------------------------------------------------------------------------+
+  | embedder_keyword      | For leveraging deep semantic embedders and text signals for structured search on keywords or short spans   |
+  |                       | of text. Glove as an additional signal has shown improvements in this use case.                            |
+  +-----------------------+------------------------------------------------------------------------------------------------------------+
+  | embedder_text         | For leveraging deep semantic embedders and text signals for unstructured search on larger paragraphs or    |
+  |                       | passages. BERT as an additional signal has shown improvements in this use case.                            |
+  +-----------------------+------------------------------------------------------------------------------------------------------------+
+
+
+The two included embedder types are `Sentence BERT <https://github.com/UKPLab/sentence-transformers>`_ and `GloVe <https://nlp.stanford.edu/projects/glove/>`_.
```
ktick, line 854:
```suggestion
The two included embedder types are `Sentence-BERT <https://github.com/UKPLab/sentence-transformers>`_ and `GloVe <https://nlp.stanford.edu/projects/glove/>`_. 
```

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+  |                       | of text. Glove as an additional signal has shown improvements in this use case.                            |
+  +-----------------------+------------------------------------------------------------------------------------------------------------+
+  | embedder_text         | For leveraging deep semantic embedders and text signals for unstructured search on larger paragraphs or    |
+  |                       | passages. BERT as an additional signal has shown improvements in this use case.                            |
+  +-----------------------+------------------------------------------------------------------------------------------------------------+
+
+
+The two included embedder types are `Sentence BERT <https://github.com/UKPLab/sentence-transformers>`_ and `GloVe <https://nlp.stanford.edu/projects/glove/>`_. 
+
+Sentence BERT is a modification of the pretrained BERT network that use siamese and triplet network structures to derive semantically meaningful sentence embeddings that can be compared using cosine-similarity. To use this embedder type, you can specify ``bert`` as the ``embedder_type`` in the ``model_settings``.
```
ktick, line 856:
```suggestion
Sentence-BERT is a modification of the pretrained BERT network that use siamese and triplet network structures to derive semantically meaningful sentence embeddings that can be compared using cosine-similarity. To use this embedder type, you can specify ``bert`` as the ``embedder_type`` in the ``model_settings``.
```

ktick: @arushir - Is there a chance that we might ever support the vanilla BERT as a built-in option? If there's a chance for that, I'd prefer calling this ``embedder_type`` as 'sentence-bert' (instead of just 'bert'). But if we're fairly certain that this adaption of BERT is the only one we'll natively support for question answering, I'm OK with calling it 'bert' for simplicity.

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+
+Multiple models fine-tuned on BERT / RoBERTa / DistilBERT / ALBERT / XLNet are provided. The available models are specified `here <https://github.com/UKPLab/sentence-transformers#english-pre-trained-models>`_, and the specific trained model can be specified via the ``model_name`` parameter in the ``model_settings`` section of the config. The default is ``bert-base-nli-mean-tokens`` which is a BERT model trained on the `SNLI <https://nlp.stanford.edu/projects/snli/>`_ and `MultiNLI <https://cims.nyu.edu/~sbowman/multinli/>`_ datasets with mean-tokens pooling.
```
ktick, line 858:
```suggestion
Multiple models fine-tuned on BERT / RoBERTa / DistilBERT / ALBERT / XLNet are provided. You can view the full list  `here <https://github.com/UKPLab/sentence-transformers#english-pre-trained-models>`_, and specify your choice of pre-trained model via the ``model_name`` parameter in the ``model_settings`` section of the config. The default is ``bert-base-nli-mean-tokens`` which is a BERT model trained on the `SNLI <https://nlp.stanford.edu/projects/snli/>`_ and `MultiNLI <https://cims.nyu.edu/~sbowman/multinli/>`_ datasets with mean-tokens pooling.
```

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+
+The provided GloVe embeddings are word vectors trained on the `Wikipedia <https://dumps.wikimedia.org/>`_ and `Gigaword 5 <https://catalog.ldc.upenn.edu/LDC2011T07>`_ datasets. To use this embedder type, you can specify ``glove`` as the ``embedder_type`` in the ``model_settings``.
```
ktick, line 860:
```suggestion
The provided GloVe embeddings are word vectors trained on the `Wikipedia <https://dumps.wikimedia.org/>`_ and `Gigaword 5 <https://catalog.ldc.upenn.edu/LDC2011T07>`_ datasets. To use this embedder type, specify ``glove`` as the ``embedder_type`` in the ``model_settings``.
```

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+
+The 50, 100, 200, and 300 dimension word embeddings are provided. The desired dimension can be specified via the ``token_embedding_dimension`` in the ``model_settings`` section of the config. For fields with more than one token, word vector averaging is used.
```
ktick, line 862:
```suggestion
The 50, 100, 200, and 300 dimension word embeddings are supported. The desired dimension can be specified via the ``token_embedding_dimension`` in the ``model_settings`` section of the config. For fields with more than one token, word vector averaging is used. 
```

ktick, line 862: @arushir Do we choose a default dimension size if ``token_embedding_dimension`` is not provided by the user? Is so, mention it here.

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+The provided GloVe embeddings are word vectors trained on the `Wikipedia <https://dumps.wikimedia.org/>`_ and `Gigaword 5 <https://catalog.ldc.upenn.edu/LDC2011T07>`_ datasets. To use this embedder type, you can specify ``glove`` as the ``embedder_type`` in the ``model_settings``.
+
+The 50, 100, 200, and 300 dimension word embeddings are provided. The desired dimension can be specified via the ``token_embedding_dimension`` in the ``model_settings`` section of the config. For fields with more than one token, word vector averaging is used. 
+
+Once your config file is specified, you can load the Knowledge Base and the embeddings will automatically be generated as specified in the config and stored in your index as dense vectors. Note that you must pass the application path in order for the config to be processed. This can be done via the CLI command with the ``--app-path`` option or via the ``app_path`` parameter for the ``load_kb`` method.
+
+
+.. code-block:: console
+
+  mindmeld load-kb hr_assistant faq_emb data/hr_faq_data.jsonl --app_path ./
```
ktick, line 869: @arushir Does ``--app_path .`` (i.e. without the `/`) not work?

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+
+All of the vectors generated at load time will be cached for faster retrieval at inference time and for future loads. It is stored with other generated data in the generated folder under the provided model name. It’s important to update the mode name when updating the model settings to maintain consistency with the cache.
```
ktick, line 871:
```suggestion
All of the vectors generated at load time will be cached for faster retrieval at inference time and for future loads. It is stored with other cached data in the ``.generated`` folder under the provided model name. It’s important to update the model name when updating the model settings to maintain consistency with the cache. 
```

ktick: @arushir I suppose that this warning of "It’s important to update the mode name when updating the model settings to maintain consistency with the cache." is only important for developers who define their own custom embeddings? If so, it may be good to move this sentence to a note box, following your explanation of custom-defined embeddings.

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+
+  mindmeld load-kb hr_assistant faq_emb data/hr_faq_data.jsonl --app_path ./
+
+All of the vectors generated at load time will be cached for faster retrieval at inference time and for future loads. It is stored with other generated data in the generated folder under the provided model name. It’s important to update the mode name when updating the model settings to maintain consistency with the cache. 
+
+.. code-block:: console
+
+  .generated/indexes/<model_name>_cache.pkl
+
+If instead you would like to use your own embedder, you can use the provided ``Embedder`` abstract class. You need to implement two methods: ``load`` and ``encode``. The load method will load and return your embedder model. The encode method will take a list of text strings and return a list of numpy vectors. You can register your class for use with MindMeld via the ``register_embedder`` method as shown below.
```
ktick, line 877:
```suggestion
If our built-in embedders don't fit your use case and you would like to use your own embedder, you can use the provided ``Embedder`` abstract class. You need to implement two methods: ``load`` and ``encode``. The load method will load and return your embedder model. The encode method will take a list of text strings and return a list of numpy vectors. You can register your class for use with MindMeld via the ``register_embedder`` method as shown below.
```

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+  class MyEmbedder(Embedder): 
+      def load(self, **kwargs): 
+          # return the loaded model 
+      def encode(self, text_list): 
+          # encodes each query in the list  
+
+  register_embedder('my_embedder_name', MyEmbedder) 
+
+
+In many cases, you may like to provide some parameters to your load method to initialize your model in a certain way. The ``model_settings`` dictionary in the config will be passed to the load method as kwargs, so any needed parameters can be specified there.
```
ktick, line 892:
```suggestion
In many cases, you may like to provide some parameters to your ``load`` method to initialize your model in a certain way. The ``model_settings`` dictionary in the config will be passed to the load method as ``kwargs``, so any needed parameters can be specified there.
```

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
...
+      "model_type": "embedder", 
+      "model_settings": { 
+          "embedder_type": "my_embedder_name", 
+          "some_model_param": "my_model_param", 
+          "embedding_fields": {"faq": ['question', 'answer']} 
+      } 
+  } 
+
+
+Once your KB has been created, to search the knowledge base leveraging vector similarity, we will use the :meth:`get()` API as before with one small modification. We set the ``query_type`` parameter to ``embedder``, ``embedder_keyword``, or ``embedder_text``. This will automatically find results for which the embedded search fields are close in cosine similarity to the embedded query.
```
ktick, line 907:
```suggestion
Once your knowledge base has been created, to search it while leveraging vector similarity, we will use the :meth:`get()` API as before with one small modification. We set the ``query_type`` parameter to ``embedder``, ``embedder_keyword``, or ``embedder_text``. This will automatically find results for which the embedded search fields are close in cosine similarity to the embedded query.
```

source/userguide/kb.rst
```diff
@@ -813,3 +813,122 @@ We can perform the same search using the :meth:`query()` API as well.
 	To delete and reindex your data, follow the steps mentioned :ref:`here <import_kb>`.
 
 
+.. _semantic_embeddings:
+
+Leveraging semantic embeddings
+------------------------------
+Up until now, the described Question Answering relies on text based retrieval. Deep learning based dense embeddings (character, word, or sentence) are in many cases better at capturing semantic information than traditional sparse vectors. Pretrained or fine-tuned embeddings can be used to find the best match in the Knowledge Base even if the search token wasn’t present in the uploaded data.
```
ktick, line 820:
```suggestion
The question answerer capabilities described so far rely purely on text-based retrieval. Deep learning based dense embeddings (character, word, or sentence) are in many cases better at capturing semantic information than traditional sparse vectors. Pretrained or fine-tuned embeddings can be used to find the best match in the knowledge base even if the search token wasn’t present in the uploaded data.
```

source/userguide/kb.rst
arushir: Fixed

source/userguide/kb.rst
arushir: Default is 300, will add that

source/userguide/kb.rst
arushir: That works as well, can update it

source/userguide/kb.rst
arushir: This is actually needed for the glove embeddings as well in some cases. The built in implementation (adapted from entity recognition) allows the user to provide a file path location for the embeddings along with the dimensionality of the embeddings. If they have different embeddings stored in different file path, for example, they would have to update the model name accordingly to match.

mindmeld/components/_config.py
arushir: Note: the code has been updated so it is fully backwards compatible.

mindmeld/components/_elasticsearch_helpers.py
arushir: Verified that it works across all versions for the methods we are using

arushir:
Note: force pushed to squash commits from the test experiments we are now running here: https://github.com/cisco/mindmeld/pull/160

mindmeld/models/embedder_models.py
```diff
@@ -0,0 +1,184 @@
...
+    def __init__(self, app_path, **kwargs):
+        """Initializes an embedder.
+        """
+        self.model_name = kwargs.get("model_name", "default")
+        self.cache_path = path.get_embedder_cache_file_path(
+            app_path, kwargs.get("embedder_type", "default"), self.model_name
+        )
+
+        folder = os.path.dirname(self.cache_path)
+        if not os.path.isdir(folder):
```
tmehlinger, line 54: Will there ever be a case that multiple `Embedder` instances will initalize concurrently in different threads or processes? If so, you'll want some additional safety around file operations here--there's a potential race condition here where two threads could simultaneously fail the check for the directory, then both try to create it. The same idea applies to other file system operations throughout this class.

If there will only ever be one instance, this will work fine the way it is.

arushir: In the current code, there will only ever be one instance at a time (see `create_embedder_model` function). I can still add some additional safety if it's cleaner that way - do you have any docs/pointers on that?

mindmeld/components/_elasticsearch_helpers.py
```diff
@@ -272,26 +341,28 @@ def load_index(
...
             if not okay:
                 logger.error("Failed to %s document %s: %r", action, doc_id, result)
             else:
                 count += 1
             pbar.update(1)
+
         # close the progress bar and flush all output
         pbar.close()
+        es_client.indices.flush(index=scoped_index_name)
+        es_client.indices.refresh(index=scoped_index_name)
```
vijay120, line 365: Could you add some documentation here on why we do this?

vijay120, line 365: @mhuang2173 Could you review this?

mhuang2173: i wonder do we need both? from their release note documentation looks like we would only need `refresh()` to make the indexed docs searchable?

arushir: Flushing makes sure data is permanently stored in the Lucene index: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-flush.html#flush-api-all-ex
Refreshing makes all operations performed on the index available for search: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html

Can add to the code as well.

arushir: the tests were failing without doing both, and in the latest release refresh is no longer part of flush: https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#flush-force-merge-no-longer-refresh

mindmeld/components/question_answerer.py
```diff
@@ -304,18 +404,54 @@ def transform(doc):
...
+                )
+
+            return mapping_data
+
+        es_client = es_client or create_es_client(es_host)
+        if is_es_version_7(es_client):
+            mapping_data = _generate_mapping_data(embedder_model, embedding_fields)
+            qa_mapping = create_index_mapping(DEFAULT_ES_QA_MAPPING, mapping_data)
+        else:
+            if embedder_model:
```
minhtuev, line 433: Should we throw an error here instead of logging it?

mindmeld/models/embedder_models.py
```diff
@@ -0,0 +1,184 @@
...
+        cache_miss_indices = [i for i, vec in enumerate(encoded) if vec is None]
+        text_to_encode = [text_list[i] for i in cache_miss_indices]
+        model_encoded_text = self.encode(text_to_encode)
+
+        for i, v in enumerate(cache_miss_indices):
+            encoded[v] = model_encoded_text[i]
+            self.cache[text_to_encode[i]] = model_encoded_text[i]
+        return encoded
+
+    def dump(self):
```
minhtuev, line 109: should we pass in an optional path where you can dump your model elsewhere outside of the cache?

mindmeld/models/embedder_models.py
```diff
@@ -0,0 +1,184 @@
...
+    return tokens
+
+
+class GloveEmbedder(Embedder):
+    """
+    Encoder class for GloVe embeddings as described here: https://nlp.stanford.edu/projects/glove/
+    """
+
+    def load(self, **kwargs):
+        DEFAULT_EMBEDDING_DIM = 300
```
minhtuev, line 149: Put this as a class constant so we can access the default embedding dim

arushir: Why would you want to access this outside of the load method? These classes are not exposed to app developers to use directly

mindmeld/models/embedder_models.py
arushir: We don't do this for the dump methods of our other components, so this seems more consistent?

minhtuev: I think the other models were pretty small and standard; if the user wants to inspect or analyze the embedding in some ways it might be convenient to allow them to dump to a custom location.

mindmeld/models/embedder_models.py
minhtuev: I think in the case if the developer wants to ask "what is the default embedding dim" it is easier to look up class constants instead of inspecting the code.

mindmeld/components/_elasticsearch_helpers.py
vijay120: @arushir Can we just use the `refresh` here and remove `flush`? Looks like bulk upload does an implicit `flush`

mindmeld/components/_elasticsearch_helpers.py
```diff
@@ -60,6 +62,26 @@ def create_es_client(es_host=None, es_user=None, es_pass=None):
...
+            major_version,
+        )
+    if major_version >= 7:
+        return True
+    return False
+
+
+def resolve_es_config_for_version(config, es_client):
+    if not is_es_version_7(es_client):
+        if DOC_TYPE not in config.get("mappings", {}):
```
vijay120, line 79: Could you add a comment on why we are doing this logic specially for ES7?

mindmeld/models/embedder_models.py
vijay120: Shouldnt ` np.average(vl, axis=0) = vl` where len(vl) == 1?

mindmeld/models/embedder_models.py
```diff
@@ -0,0 +1,184 @@
...
+        return SentenceTransformer(bert_model_name)
+
+    def encode(self, text_list):
+        return self.model.encode(text_list)
+
+
+def get_query_tokens(query):
+    """Splits into tokens, removes punctuation, and removes whitespace.
+    """
+    query = query.translate(str.maketrans("", "", string.punctuation))
```
vijay120, line 137: I am not exactly sure what the `str.maketrans` is doing:
```
>>> str.maketrans("", "", string.punctuation)
{33: None, 34: None, 35: None, 36: None, 37: None, 38: None, 39: None, 40: None, 41: None, 42: None, 43: None, 44: None, 45: None, 46: None, 47: None, 58: None, 59: None, 60: None, 61: None, 62: None, 63: None, 64: None, 91: None, 92: None, 93: None, 94: None, 95: None, 96: None, 123: None, 124: None, 125: None, 126: None}
```

pytest.ini
```diff
@@ -6,3 +6,5 @@ markers =
     no_extras: tests which require extras to not be present
     tensorflow: tests which require the tensorflow extra
     no_tensorflow: tests which require the tensorflow extra to not be present
+    bert: tests which require the bert extra
+    no_bert: tests which require the bert extra to not be present
```
minhtuev, line 10: we should also add es7 marker here

mindmeld/models/embedder_models.py
arushir: Not sure I understand, how would it be easier? They would still have to inspect the code, just one line above where it is now, right? If the developer is looking for defaults they would most likely check the documentation?

mindmeld/models/embedder_models.py
arushir: This is just a cache of embeddings, not the model itself, so I don't think it would be useful to inspect/analyze?

mindmeld/components/_elasticsearch_helpers.py
arushir: Interesting, yes I'll try that.

mindmeld/models/embedder_models.py
arushir: This line removes punctuation.

https://docs.python.org/3.3/library/stdtypes.html?highlight=maketrans#str.maketrans

So with maketrans, the first two arguments are mapping empty to empty (no-op), the third argument is mapping all punctuation to None. This passed to query.translate, will give us the query with punctuation removed.

mindmeld/components/_elasticsearch_helpers.py
```diff
@@ -223,13 +252,53 @@ def delete_index(
...
+    for emb in embedding_properties:
+        properties[emb["field"]] = {"type": "dense_vector", "dims": emb["dims"]}
+    if "mappings" not in base_mapping:
+        base_mapping["mappings"] = base_mapping
+    base_mapping["mappings"]["properties"] = properties
+    return base_mapping
+
+
+def version_compatible_streaming_bulk(
+    es_client, docs, index, chunck_size, raise_on_error, doc_type
```
serapio, line 273:
```suggestion
    es_client, docs, index, chunk_size, raise_on_error, doc_type
```

mindmeld/models/embedder_models.py
minhtuev: I see, OK

mindmeld/components/question_answerer.py
```diff
@@ -69,7 +90,14 @@ def _es_client(self):
...
 
-    def get(self, index, size=10, query_type="keyword", **kwargs):
+    @property
+    def _query_type(self):
+        if self._qa_config.get("model_type") in ALL_QUERY_TYPES:
+            return self._qa_config.get("model_type")
+        else:
+            return DEFAULT_QUERY_TYPE
+
+    def get(self, index, size=10, query_type=None, **kwargs):
```
serapio, line 100: the docstring for `query_type` should be updated

mindmeld/models/embedder_models.py
minhtuev: They can use auto-hinting (`GloveEmbedder. DEFAULT_EMBEDDING_DIM`) or look at the generated API reference. For example, you can see the constant `CLF_TYPE` here:
https://www.mindmeld.com/docs/apidoc/mindmeld.components.role_classifier.html#mindmeld.components.role_classifier.RoleClassifier.CLF_TYPE

mindmeld/components/question_answerer.py
```diff
@@ -256,23 +299,74 @@ def load_kb(
...
 
-        def _doc_count(data_file):
             with open(data_file) as data_fp:
                 line = data_fp.readline()
                 data_fp.seek(0)
                 if line.strip() == "[":
                     docs = json.load(data_fp)
-                    return len(docs)
+                    count = len(docs)
+                    doc_example = docs[0]
```
serapio, line 338: This seems like something that could surprise the developer. Suppose they have an inventory with up to 3 reviews, with keys like "product_review_0", "product_review_1" ... The first doc might have only one or none of these, and then the fields would get left out of the embeddings.

mindmeld/components/question_answerer.py
```diff
@@ -256,23 +299,74 @@ def load_kb(
...
                         count += 1
-                    return count
-
-        def _doc_generator(data_file):
-            def transform(doc):
+                    doc_example = json.loads(line)
+                embedding_fields = [
+                    key
+                    for key in doc_example.keys()
+                    if match_regex(key, embedding_fields)
```
serapio, line 347: could we use a different variable name for the string field names vs the regex patterns

mindmeld/components/question_answerer.py
```diff
@@ -495,7 +637,7 @@ def _build_clause(self, clause_type, query_type="keyword", **kwargs):
         else:
             raise Exception("Unknown clause type.")
 
-    def query(self, query_type="keyword", **kwargs):
+    def query(self, query_type=DEFAULT_QUERY_TYPE, **kwargs):
```
serapio, line 640: this docstring should be updated too

mindmeld/models/embedder_models.py
serapio: Should we use our own tokenizer for this, or is this meant to align with BERT preprocessing? Like this would squash "language-dependent", "smeagol@precious-my.webex.com", "absolutely—that" into single tokens.

mindmeld/components/_elasticsearch_helpers.py
arushir: Sure added. It is because of the breaking change in ES7, and because we don't want users to have to re-index existing indices that were created with older versions of MindMeld: https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html#removal-of-types

mindmeld/components/_elasticsearch_helpers.py
arushir: Updated to use only refresh

mindmeld/models/embedder_models.py
arushir: I think the two are similar enough, updated to use the MM tokenizer.

mindmeld/components/question_answerer.py
arushir: Yeah good point, updating

arushir:
@vijay120 @serapio thanks for the feedback, I've finished implementing those changes, let me know if it looks good to merge!

source/userguide/kb.rst
```diff
@@ -813,3 +813,124 @@ We can perform the same search using the :meth:`query()` API as well.
 	To delete and reindex your data, follow the steps mentioned :ref:`here <import_kb>`.
 
 
+.. _semantic_embeddings:
+
+Leveraging semantic embeddings
+------------------------------
+The question answerer capabilities described so far rely purely on text-based retrieval. Deep learning based dense embeddings (character, word, or sentence) are in many cases better at capturing semantic information than traditional sparse vectors. Pretrained or fine-tuned embeddings can be used to find the best match in the knowledge base even if the search token wasn’t present in the uploaded data.
+
+To leverage semantic embeddings in search, the first step is to generate the embeddings for your desired knowledge base fields. You can use one of the provided embedders or use your own. If your app mainly consists of standard English vocabulary, one of provided embedders may work well enough, but if the text you are searching against has quite a bit of domain specific vocabulary, you may benefit from training or fine tuning your own embedder on your data.
```
serapio, line 822:
```suggestion
To leverage semantic embeddings in search, the first step is to generate the embeddings for your desired knowledge base fields. You can use one of the provided embedders or use your own. If your app mainly consists of standard English vocabulary, one of the provided embedders may work well enough, but if the text you are searching against has quite a bit of domain-specific vocabulary, you may benefit from training or fine tuning your own embedder on your data.
```

source/userguide/kb.rst
```diff
@@ -813,3 +813,124 @@ We can perform the same search using the :meth:`query()` API as well.
...
+      }
+  }
+
+There are three available model types which leverage semantic embedders. The specified ``model_type`` is also the default ``query_type`` for all question answering calls for your application, but you can always pass in a different query type to your ``qa.get()`` command as desired.
+
+  +-----------------------+------------------------------------------------------------------------------------------------------------+
+  | Model type            | Description                                                                                                |
+  +=======================+============================================================================================================+
+  | embedder              | Only leverage deep semantic embedders. This option allows for using deep semantic embedders like           |
+  |                       | Sentence-BERT or GloVe for doing vector-based retrieval                                                    |
```
serapio, line 846:
```suggestion
  |                       | Sentence-BERT or GloVe for doing vector-based retrieval.                                                    |
```

mindmeld/components/question_answerer.py
```diff
@@ -256,23 +304,67 @@ def load_kb(
                 connection to the Elasticsearch host.
             clean (bool): Set to true if you want to delete an existing index
                 and reindex it
+            app_path (str): The path to the directory containing the app's data
+            config (dict): The QA config if passed directly rather than loaded from the app config
         """
+        embedder_model = None
+        embedding_fields = []
+        if not app_path and not config:
+            logger.warning(
```
vijay120, line 313: Nit: Shouldnt this warning be localized to when devs use the embedder or not when they use regular ES?

arushir: Not sure I understand - the only way to tell if someone is trying to use the embedder is if they provide an application config. How would we filter cases for which to give this warning?

mindmeld/models/embedder_models.py
vijay120: I believe @tmehlinger 's scenario happens when two MindMeld application pointing to the same app_path are running happen to reach this code path simultaneously. I don't think this scenario is very likely. Moreover, there aren't any adverse effects in clearing the cache apart from longer processing time. @tmehlinger Let us know if these reasons are sufficient to address the issue.

mindmeld/models/taggers/embeddings.py
```diff
@@ -192,6 +192,7 @@ def __init__(
         sequence_padding_length,
         token_embedding_dimension=None,
         token_pretrained_embedding_filepath=None,
+        use_padding=True,
```
vijay120, line 195: Could we combine `sequence_padding_length ` and `use_padding ` by just stating that if `sequence_padding_length = 0`, then `use_padding => False`?

tests/components/test_question_answerer.py
```diff
@@ -63,12 +61,53 @@ def food_ordering_answerer(food_ordering_app_path, es_client):
...
+
+
+@pytest.mark.extras
+@pytest.fixture
+def food_ordering_with_glove(food_ordering_app_path, es_client):
+    glove_qa_config = {
+        "model_type": "embedder",
+        "model_settings": {
+            "embedder_type": "glove",
+            "embedding_fields": {"menu_items_glove": ["name"]},
```
vijay120, line 97: @arushir  Do you think it's intuitive for developers to specify the field name derived from the entity type ie `menu_items` and MindMeld automatically mapping that to the correct field name `menu_items_glove `based on the `embedder_type `? Or do you think it's better for the developer to explicitly specify `menu_items_glove `?

mindmeld/components/question_answerer.py
vijay120: From my understanding, this warning will occur in the most common scenario where the user is just using `{"model_type": "keyword"}` and does not provide `app_path `. In this case, the warning of `You must provide either the application path to upload embeddings as specified  in the app config or directly provide the QA config` doesn't concern him/her since they didnt intend to use an embedding.

arushir: If they did not provide either an app_path or a config, there is no way to know that they have specified `{"model_type": "keyword"}`

tests/components/test_question_answerer.py
arushir: The key in the `embedding_fields` dictionary is the name that the developer gives their index. If the name of their index is `menu_items`, they would specify `menu_items` here. I just used `menu_items_glove` because I created two different indices for the tests (see pytest fixtures above), one with glove embeddings and another for bert embeddings

mindmeld/models/taggers/embeddings.py
arushir: I think it would be counterintuitive to combine them. `sequence_padding_length` is not how much padding we add to the end of each vector, but the total length of the padded sequence. So by that definition, `sequence_padding_length=0` and `use_padding=False` are two different things. In the former you would return a vector of length 0 and the latter you would return a vector of length `token_embedding_dimension * token_sequence_length`

PR #149, opened 2020-05-19
Updates to Slot Filling

ritvikshrivastava:
Changes Added:

1. Custom validation function now receives the ``request`` object in as a replacement to the user input text.
2. Improved gazetteer validation to capture entities even if the user input has extra surrounding words. (e.g. from the banking blueprint 'to my checking account' now works).
3. Added support for asynchronous app
4. Added new tests.


mindmeld/components/dialogue.py
```diff
@@ -720,23 +718,17 @@ def __init__(self, entrance_handler, form, app):
         self._check_attr()
 
     def _check_attr(self):
-        if not ("entities" in self._form and len(self._form["entities"]) > 0):
+        if not ('entities' in self._form and len(self._form['entities']) > 0):
```
minhtuev, line 721: you shoud apply black on this

mindmeld/components/dialogue.py
```diff
@@ -867,29 +862,39 @@ def _initial_fill(self, request):
...
-            params=request.params,
+            params=request.params
         )
 
         self._exit_flow(responder)
+
+        if async_mode:
+            return self._end_slot_fill_sync(request, responder)
+        else:
+            self._end_slot_fill_async(request, responder)
```
minhtuev, line 890: here as well?
```suggestion
            return self._end_slot_fill_async(request, responder)
```

mindmeld/components/dialogue.py
```diff
@@ -906,9 +911,7 @@ def _retry_logic(self, responder, nlr):
             self._app.app_manager.dialogue_manager.reprocess()
 
     def __call__(self, request, responder):
-        """The iterative call to fill missing slots in the entity form till all slots have been
-        filled up or the flow has been exited.
-
+        """
```
minhtuev, line 914: why remove doc?

ritvikshrivastava: Just removed the first half. It's been moved to the individual call functions.

mindmeld/components/dialogue.py
ritvikshrivastava: Done.

mindmeld/components/dialogue.py
```diff
@@ -796,21 +796,22 @@ def _validate(self, request, slot):
...
 
         extracted_feature = {}
         _resolved_value = {}
+        query = None
 
         if slot.default_eval:
             if entity_type in DEFAULT_SYS_ENTITIES:
                 # system entity validation - checks for presence of required system entity
+
+                if request.entities:
```
minhtuev, line 808:
```suggestion
                entity_text = str(request.entities[0]["value"][0]["value"]) if request.entities else text
                query = self._extract_query_features(entity_text)
```

mindmeld/components/dialogue.py
```diff
@@ -820,6 +821,19 @@ def _validate(self, request, slot):
 
             else:
                 # gazetteer validation
+
+                # payload format for entity feature extractors:
+                # tuple(query (Query Object), list of entities, entity index)
+                if request.entities:
```
minhtuev, line 827: is this logic similar to line 808?

ritvikshrivastava: No the follow-up logic is a bit different.

mindmeld/components/dialogue.py
ritvikshrivastava: Done!

source/userguide/dm.rst
```diff
@@ -618,8 +618,8 @@ Transfer money in a banking app:
 
     from mindmeld.core import FormEntity
 
-    def test_for_money(ent):
-      return True if '$' in ent else False
+    def test_for_money(text, request):
```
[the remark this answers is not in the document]
ritvikshrivastava: I believe so, in ``request.text``. This is a redundancy but I thought text explicitly providing the developer with text and request as an additional resource might be easier.

mindmeld/components/dialogue.py
```diff
@@ -820,6 +821,19 @@ def _validate(self, request, slot):
...
+                    query = self._extract_query_features(
+                        request.entities[0]["value"][0]["cname"]
+                    )
+
+                if not query:
+                    query = self._extract_query_features(text)
+
+                formatted_payload = (query, [query], 0)
```
minhtuev, line 835: Tiny nit: does this look neater this way?

```suggestion
                extracted_feature = entity_features.extract_in_gaz_features()(
                    (query, [query], 0), gazetteer
                )
```

mindmeld/components/dialogue.py
```diff
@@ -848,10 +861,7 @@ def _validate(self, request, slot):
             # Custom validation using function provided by developer. Should return True/False
             # for validation status. If true, then continue, else fail overall validation.
 
-            if (
-                slot.custom_eval(text) not in (True, False)
-                or slot.custom_eval(text) is False
-            ):
+            if slot.custom_eval(text, request) is not True:
```
minhtuev, line 864: Nit:
```suggestion
            if not slot.custom_eval(text, request):
```

mindmeld/components/dialogue.py
```diff
@@ -957,7 +977,25 @@ def __call__(self, request, responder):
...
+        self.__call__(request, responder)
+
+    async def call_async(self, request, responder):
+        """The slot-filling call for asynchronous apps
+
+        Args:
+            request (Request): The request object.
+            responder (DialogueResponder): The responder object.
+        """
+        self.__call__(request, responder)
```
minhtuev, line 998: @tmehlinger: my knowledge of async is a bit spotty here but do we need an await statement here as well?
```suggestion
        await self.__call__(request, responder)
```

tests/components/test_auto_fill.py
```diff
@@ -18,9 +18,23 @@ def test_auto_fill_happy_path(kwik_e_mart_app, kwik_e_mart_app_path, qa_kwik_e_m
...
+def test_auto_fill_happy_path_validation(kwik_e_mart_app, kwik_e_mart_app_path, qa_kwik_e_mart):
+    """Tests a happy path for the app with gazetter validation."""
+    convo = Conversation(
+        app=kwik_e_mart_app, app_path=kwik_e_mart_app_path, force_sync=True
+    )
+    directives = convo.process("What's the store phone number?").directives
+    assert_target_dialogue_state(convo, "send_store_phone")
+    assert_reply(directives, "Which store would you like to know about?")
+
+    directives = convo.process("the store on elm street").directives
```
minhtuev, line 31: do we need to make assertions for these directives? I am not sure what we are testing here

mindmeld/components/dialogue.py
```diff
@@ -796,21 +796,22 @@ def _validate(self, request, slot):
...
 
         if slot.default_eval:
             if entity_type in DEFAULT_SYS_ENTITIES:
                 # system entity validation - checks for presence of required system entity
+
+                entity_text = (
+                    str(request.entities[0]["value"][0]["value"])
+                    if request.entities
+                    else text
+                )
```
vijay120, line 812: Does this mean we HAVE to detect the non-contextual entity using the tagger which are common in such slot filling use-cases? From my experience, its extremely hard to detect these non-contextual queries just by the tagger, which is why it made sense to pass the whole query through the `_extract_query_features ` function.

tests/components/test_auto_fill.py
```diff
@@ -52,3 +66,19 @@ def test_auto_fill_exit_flow(kwik_e_mart_app, kwik_e_mart_app_path):
     directives = convo.process("exit").directives
     assert_reply(directives, "Sorry I cannot help you. Please try again.")
     assert_target_dialogue_state(convo, None)
+
+
+@pytest.mark.conversation
+@pytest.mark.asyncio
+async def test_auto_fill_async(async_kwik_e_mart_app, kwik_e_mart_app_path):
+    """Tests auto fill for async apps."""
+    convo = Conversation(app=async_kwik_e_mart_app, app_path=kwik_e_mart_app_path)
```
minhtuev, line 75: I think if you set `app`, you don't need to set `app_path`. Setting `force_sync` equal to true will allow you to make the assertion normally without await.
 
```suggestion
    convo = Conversation(app=async_kwik_e_mart_app, force_sync=True)
```

mindmeld/components/dialogue.py
ritvikshrivastava: Yeah that's exactly what is happening. If the entities are not detected through the tagger, then the whole query is passed through the ``_extract_query_features``.

vijay120, line 812: Nice!

tests/components/test_auto_fill.py
```diff
@@ -18,9 +18,23 @@ def test_auto_fill_happy_path(kwik_e_mart_app, kwik_e_mart_app_path, qa_kwik_e_m
     assert_target_dialogue_state(convo, None)
 
 
+@pytest.mark.conversation
+def test_auto_fill_happy_path_validation(kwik_e_mart_app, kwik_e_mart_app_path, qa_kwik_e_mart):
+    """Tests a happy path for the app with gazetter validation."""
+    convo = Conversation(
+        app=kwik_e_mart_app, app_path=kwik_e_mart_app_path, force_sync=True
```
minhtuev, line 25: I think you don't need the rest of the params
```suggestion
        app=kwik_e_mart_app)
```

mindmeld/components/dialogue.py
ritvikshrivastava: Can't use ``not``. Negation only applies to boolean cases. This is handling any output other than a True value (which could be non boolean too).

tests/components/test_auto_fill.py
```diff
@@ -52,3 +66,19 @@ def test_auto_fill_exit_flow(kwik_e_mart_app, kwik_e_mart_app_path):
...
+    """Tests auto fill for async apps."""
+    convo = Conversation(app=async_kwik_e_mart_app, app_path=kwik_e_mart_app_path)
+    response = await convo.process("What's the store phone number?")
+    directives = response.directives
+    assert_target_dialogue_state(convo, "send_store_phone")
+    assert_reply(directives, "Which store would you like to know about?")
+
+    response = await convo.process("exit")
+    directives = response.directives
+    assert_reply(directives, "Sorry I cannot help you. Please try again.")
```
minhtuev, line 83: We should have a complete flow test here where we fill the slots completely

source/userguide/dm.rst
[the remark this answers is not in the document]
vijay120: Could we provide an example where the request object is also used? It seems puzzling to pass it in without using it.

mindmeld/components/dialogue.py
minhtuev: That is not our problem, in the sense that the custom eval functions are expected to return bool anyway?

mindmeld/components/dialogue.py
vijay120, line 998: Do we have to do `await self.__call__(request, responder)` here?

source/userguide/dm.rst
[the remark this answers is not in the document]
ritvikshrivastava: Good point! I'll add that.

mindmeld/components/dialogue.py
ritvikshrivastava: It is, but then we'd need to do error handling separately. Isn't this a shorter fix?

tmehlinger: The docs (that you created! 😄) say `custom_eval` should return a boolean. I would just make that assumption here. I don't see why there should be any error handling to determine whether we received a boolean value when that's what the docs dictate. If someone misuses it, the behavior is undefined.

In any case, explicit comparisons to `True` or `False` are a style no-no. It implies the expression could be something _other_ than boolean, and mixing other types with boolean is an even bigger no-no.

ritvikshrivastava: Okay! Makes sense :)

source/userguide/dm.rst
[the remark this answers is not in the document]
ritvikshrivastava: Merged the two now. Text and request. The example covers the whole idea.

mindmeld/components/dialogue.py
ritvikshrivastava: Updated the order slightly, but kept the extra variable. Just felt easier to read and explain to anyone who reads the code later.

mindmeld/components/dialogue.py
tmehlinger: You only `await` coroutines, `__call__` is a regular function. An `async` _allows_ you to `await` coroutines and to be `await`ed by something further up the stack. It does not require you to only invoke coroutines, you're free to call synchronous functions however you like.

You also don't generally want to call dunder methods directly because they're just the implementation for the low-level interface. In this case, `__call__` makes instances of this class callable so you can just do that:

```python
async def call_async(self, request, responder):
    self(request, responder)
```

tmehlinger: Expanding on this a bit more... there's really no reason to have this method or `call_sync`. `__call__` isn't asynchronous and can't be made asynchronous so you might as well drop both `call_sync` and `call_async` and just invoke instances of this class as callables.

```python
filler = AutoEntityFilling()
filler(my_request, my_responder)
```

ritvikshrivastava: I think the reason I did the two functions was that the application file is where these were being called and passed into the ``add_dialogue_rule`` method, which requires an asynchronous handler to be passed when app is in async mode. Having a common __call__ doesn't allow for that? What's the best alternative then?

tmehlinger: That makes sense, I'm with you now. Carry on!

mindmeld/components/dialogue.py
```diff
@@ -979,14 +979,14 @@ def __call__(self, request, responder):
         # Finish slot-filling and return to handler
         return self._end_slot_fill(request, responder, self._app.async_mode)
 
-    def call_sync(self, request, responder):
-        """The slot-filling call for synchronous apps
+    # def call_sync(self, request, responder):
```
minhtuev, line 982: remove?

ritvikshrivastava: I had updated this. Strange. Fixed now!

PR #156, opened 2020-06-12
Reprocessing queries for slot-filling

ritvikshrivastava:
This PR allows slot-filling to use ``allowed_intents`` and reprocess a query when the retry logic is exhausted. Also adds a test to check the same.

This comes from the observation that the in-built ``reprocess`` does not work with allowed_intents set from the previous turn and always reprocesses to those specified intents. This fix by-passes the reprocess call and directly calls the ``apply_handler`` functionality in the DM with a repeat ``nlp.process`` call and creation of a new request object that is passed to the DM as a new query turn.

Also, fixes a validation bug.

### DIFF
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -805,11 +805,11 @@ class AutoEntityFilling:
             if entity_type in DEFAULT_SYS_ENTITIES:
                 # system entity validation - checks for presence of required system entity
 
-                entity_text = (
-                    str(request.entities[0]["value"][0]["value"])
-                    if request.entities
-                    else text
-                )
+                try:
+                    entity_text = str(request.entities[0]["value"][0]["value"])
+                except (KeyError, IndexError):
+                    entity_text = text
+
                 query = self._extract_query_features(entity_text)
 
                 resources = {}
@@ -822,12 +822,11 @@ class AutoEntityFilling:
             else:
                 # gazetteer validation
 
-                if request.entities:
+                try:
                     query = self._extract_query_features(
                         request.entities[0]["value"][0]["cname"]
                     )
-
-                if not query:
+                except (KeyError, IndexError):
                     query = self._extract_query_features(text)
 
                 gaz = self._app.app_manager.nlp.resource_loader.get_gazetteer(
@@ -890,6 +889,7 @@ class AutoEntityFilling:
         # Returns filled entity objects as request.entities
         # We pass in the previous turn's responder's params to the current request
         request = self._app.app_manager.request_class(
+            text=request.text,
             entities=[slot.value for slot in self._local_form],
             context=request.context or {},
             history=request.history or [],
@@ -914,7 +914,7 @@ class AutoEntityFilling:
         self._retry_attempts = 0
         self._prompt_turn = False
 
-    def _retry_logic(self, responder, nlr):
+    def _retry_logic(self, request, responder, nlr):
         if self._retry_attempts < self._max_retries:
             self._retry_attempts += 1
             responder.reply(nlr)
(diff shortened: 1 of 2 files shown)


mindmeld/components/dialogue.py
```diff
@@ -914,15 +914,26 @@ def _prompt_slot(self, responder, nlr):
...
         if self._retry_attempts < self._max_retries:
             self._retry_attempts += 1
             responder.reply(nlr)
         else:
             # max attempts exceeded, reset counter, exit auto_fill.
             self._retry_attempts = 0
             self._exit_flow(responder)
-            self._app.app_manager.dialogue_manager.reprocess()
+
+            processed_query = self._app.app_manager.nlp.process(query_text=request.text)
```
minhtuev, line 926: Should `request` already encapsulate all of the fields?

ritvikshrivastava: No it doesn't because the allowed intents were set at the turn before this, hence the request before this step has fixed intent and param definition. nlp.process here is applied after removing the allowed_intents.

mindmeld/components/dialogue.py
```diff
@@ -914,15 +914,26 @@ def _prompt_slot(self, responder, nlr):
...
+
+            request = self._app.app_manager.request_class(
+                context=request.context or {},
+                history=request.history or [],
+                frame=responder.frame or {},
+                params=FrozenParams(**DialogueResponder.to_json(responder.params)),
```
minhtuev, line 932: request.params?

mindmeld/components/dialogue.py
minhtuev: where is that removal being done?

mindmeld/components/dialogue.py
ritvikshrivastava: No, we're replicating the responder params to become the new request params. Basically what Conversation.process would do in the app_manager.

The idea is the replicate the next turn without actually going to the next turn.

mindmeld/components/dialogue.py
ritvikshrivastava: In ``self._exit_flow``, line 924. That function resets both the ``target_dialogue_state`` and ``allowed_intents``

mindmeld/components/dialogue.py
vijay120: `params=FrozenParams(**responder.params.to_dict()),`

mindmeld/components/dialogue.py
minhtuev: I see got it, can you add a comment here about why we need to reprocess the NLP component

tests/components/test_auto_fill.py
```diff
@@ -60,6 +60,24 @@ def test_auto_fill_exit_flow(kwik_e_mart_app, qa_kwik_e_mart):
     assert_target_dialogue_state(convo, None)
 
 
+@pytest.mark.conversation
+def test_auto_fill_switch_flow(kwik_e_mart_app, qa_kwik_e_mart):
+    """Tests slot-fill switch flow logic."""
```
minhtuev, line 65: can you give more description to the switch flow logic?

mindmeld/components/dialogue.py
```diff
@@ -805,11 +805,11 @@ def _validate(self, request, slot):
             if entity_type in DEFAULT_SYS_ENTITIES:
                 # system entity validation - checks for presence of required system entity
 
-                entity_text = (
-                    str(request.entities[0]["value"][0]["value"])
-                    if request.entities
-                    else text
-                )
+                try:
```
minhtuev, line 808: can you add the tests that this try-except is supposed to catch?

ritvikshrivastava: Added!

PR #157, opened 2020-06-17
Updating to scikit-learn v0.23.1

vrdn-23:
This is a PR with changes to upgrade the current scikit-learn version from 0.19.2 to 0.23.1

The major changes are 3-fold:
1) The default solver for LogisticRegression for scikit-learn changed from "liblinear" to "lbfgs" in version 0.22 (documentation [here](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html).) 

2) The dependency for joblib was removed from sklearn.externals and now requires to imported as a new module and installed separately. (See some references [here](https://github.com/scikit-learn/scikit-learn/issues/14177) and [here](https://github.com/EpistasisLab/tpot/issues/869)).

3) The [LabelEncoder class](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.LabelEncoder.html) in scikit-learn has added a new check functionality in the inverse_transform method. The following is a comparative code difference:

Old code in v0.19.2:
```
def inverse_transform(self, y):

        ...
        ...

        check_is_fitted(self, 'classes_')

        diff = np.setdiff1d(y, np.arange(len(self.classes_)))
        if diff:
            raise ValueError("y contains new labels: %s" % str(diff))
        y = np.asarray(y)
        return self.classes_[y]
```
New code in v0.23.1:
```
def inverse_transform(self, y):

        ...
        ...

        check_is_fitted(self)
        y = column_or_1d(y, warn=True)
        # inverse transform of empty array is empty array
        if _num_samples(y) == 0:
            return np.array([])

        diff = np.setdiff1d(y, np.arange(len(self.classes_)))
        if len(diff):
            raise ValueError(
                    "y contains previously unseen labels: %s" % str(diff))
        y = np.asarray(y)
        return self.classes_[y]
```
That new check for column_or_1d causes the current tests to fail unless a 1d array is passed as input to the function. Previously the code would work, even if it was a numpy.int64 input. To circumvent this, the change in [memm.py](mindmeld/models/taggers/memm.py) is introduced.

I think I may have to work on getting the formatting right. I seem to pass the lintme score and pytests reasonably but please let me know if there are any other comments or changes that need to be made! 

The changes have been tested locally on Python 3.6.5


mindmeld/components/_config.py
```diff
@@ -42,11 +42,12 @@
 
 DEFAULT_DOMAIN_CLASSIFIER_CONFIG = {
     "model_type": "text",
-    "model_settings": {"classifier_type": "logreg",},
+    "model_settings": {"classifier_type": "logreg", },
     "param_selection": {
         "type": "k-fold",
         "k": 10,
-        "grid": {"fit_intercept": [True, False], "C": [10, 100, 1000, 10000, 100000]},
+        "grid": {"solver": ["liblinear"],
```
vijay120, line 49: I dont think it makes sense to add a single entry to the grid payload. Instead, we should just default load `liblinear ` into the `logreg` implementation.

mindmeld/models/taggers/memm.py
```diff
@@ -150,7 +150,7 @@ def _predict_proba_example(self, example, config, resources):
             X, _ = self._preprocess_data([features])
             prediction = self._clf.predict_proba(X)[0]
             predicted_tag = np.argmax(prediction)
```
vijay120, line 152: Maybe we could do this:
```
prediction = self._clf.predict_proba(X)
predicted_tag = np.argmax(prediction)
prev_tag = self.class_encoder.inverse_transform(predicted_tag])
```

setup.py
```diff
@@ -24,8 +24,9 @@
     "python-dateutil~=2.6",
     "pytz",  # uses calendar versioning
     "scipy>=0.13.3,<2.0",
-    'scikit-learn>=0.18.1,<0.20; python_version < "3.7"',
-    'scikit-learn>=0.19.2,<0.20; python_version >= "3.7"',
+    'scikit-learn>=0.23.1; python_version >= "3.6"',
+    'scikit-learn~=0.19.2, <20.0; python_version < "3.6"',
```
vijay120, line 28: Maybe this:
`'scikit-learn>=0.18.1,<0.20; python_version < "3.6"',`

Are we supporting 0.23.1 for 3.6 and above and leaving <3.6 to what is there before?

mindmeld/models/tagger_models.py
```diff
@@ -16,7 +16,7 @@
 import os
 import random
 
-from sklearn.externals import joblib
+import joblib
```
minhtuev, line 19: does this work with lower version of sklearn?

maybe we can do this:
```
try: 
    import joblib
except:
    from sklearn.externals import joblib

File: setup.py [hunk 5]
```diff
@@ -55,7 +56,7 @@
     author="Cisco Systems, Inc.",
     author_email="contact@mindmeld.com",
     url="https://github.com/cisco/mindmeld",
-    packages=["mindmeld",],
+    packages=["mindmeld", ],
```
minhtuev, line 59: haha why this extra space?

minhtuev:
I agree with @vijay120 regarding backward compatibility, especially now that we have other Cisco teams building apps and depending on MindMeld.

mindmeld/models/tagger_models.py
JavaiMaster: I'm pretty sure it does. I think the joblib package works same as it's scikit-lean equivalent. And did you mean for the try catch be the other way around? :)

mindmeld/components/_config.py
serapio: What is the reason to keep `liblinear` as our default? We have had our own issues with liblinear and did switch to lbfgs for some apps at that time. I don't remember now if that affected how the other params are specified, but lbfgs is a reasonable default.

mindmeld/models/text_models.py
```diff
@@ -45,13 +46,13 @@
 DECISION_TREE_TYPE = "dtree"
 RANDOM_FOREST_TYPE = "rforest"
 SVM_TYPE = "svm"
+MLP_TYPE = "mlp"
 SUPER_LEARNER_TYPE = "super-learner"
```
serapio, line 50: The rest of the super-learner was stripped out. We should remove this line too.

mindmeld/components/_config.py
JavaiMaster: The old version uses liblinear as default but the new version uses lbfgs. If this is not a hard requirement, then I can remove any places where this is enforced. I initially added it cause I thought this was the solver giving the best result for the default use cases. Please do let me know whether it's needed @vijay120 @serapio

setup.py
JavaiMaster: I think 0.18.1 is almost 5 versions behind the previous one. I kept the versioning for python<3.6 to 19.2, because 0.23.1 doesn't support below python 3.7. I don't see any significant changes (in the changelog) that would impact us if we shift from 0.18.1 to 0.19.2.
Is there any additional tests you would want to confirm that it's safe to move to 0.19.2?

mindmeld/components/_config.py
serapio: after reviewing the sklearn documentation, I see that the choice of liblinear vs lbfgs does interact with the hyper-params `penalty`, `fit_intercept` and `multi_class`, e.g. `penalty=l1` is only valid with liblinear, and `penalty=none` is not valid with liblinear. So existing configs that were implicitly assuming liblinear might degrade performance switching to lbfgs, but retrained models with default params or with params appropriate to lbfgs should be fine (and existing trained models would not be affected). I'm inclined to just warn in the release notes how the change in default solver is relevant to the other hyper-parameters, to keep the configs simpler and consistent with sklearn defaults.

serapio:
A few places we are suppressing warnings like this:
```
    warnings.filterwarnings(
        "module", category=DeprecationWarning, module="sklearn.preprocessing.label"
    )
```
Hopefully that's not necessary anymore.

mindmeld/models/text_models.py
vrdn-23: Removed

setup.py
vrdn-23: Fixed

mindmeld/models/taggers/memm.py
vrdn-23: Done

mindmeld/components/_config.py
vrdn-23: Fixed as discussed. I've let the default go to lbfgs and changed the kwik_e_mart settings to be compatible with the tests.

setup.py
vrdn-23: Hey @vijay120 , let me know what you would like me to do here.

source/userguide/entity_recognizer.rst
```diff
@@ -485,7 +485,7 @@ Here's an example of a ``config.py`` file where custom settings optimized for th
            'k': 5,
            'scoring': 'accuracy',
            'grid': {
-               'penalty': ['l1', 'l2'],
+               'penalty': ['none', 'l2'],
```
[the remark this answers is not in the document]
vrdn-23: Oof. Nice catch!
```python
>>> a = LR(penalty='none', C=0.05)
>>> a
LogisticRegression(C=0.05, penalty='none')
>>> clf = a.fit(X, y)
/Users/vidamoda/.pyenv/versions/mindmeld/lib/python3.6/site-packages/sklearn/linear_model/_logistic.py:1321: UserWarning: Setting penalty='none' will ignore the C and l1_ratio parameters
  "Setting penalty='none' will ignore the C and l1_ratio "
```
It gives a user warning but apart from that the code doesn't break. Would you like me to remove it nonetheless?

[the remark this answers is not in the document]
serapio: Since it's an example, it would be better to just leave it out.

tests/components/test_entity_recognizer.py
```diff
@@ -11,7 +11,7 @@ def test_memm_model(kwik_e_mart_app_path):
         "k": 5,
         "scoring": "accuracy",
         "grid": {
-            "penalty": ["l1", "l2"],
+            "penalty": ["l2"],
```
serapio, line 14: Could remove this since it is the default.

vrdn-23: I tried removing it but it failed on the assert a few lines below so I'm keeping it for the time being.

source/userguide/entity_recognizer.rst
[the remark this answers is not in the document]
vrdn-23: Fixed!

vrdn-23:
**TL;DR: There is no clean way to port saved models from previous scikit-learn versions to the new versions. Makes sense to update the version when we have more substantial changes to go with it. Closing the PR for now and hopefully it can be used for future reference!**

So after much experimentation and a lot (I mean A LOT) of help from @vijay120 , there are a few issues that were discovered on the topic of backward compatibility for the models for previous versions of scikit-learn. 

1. The official policy of the scikit-learn team is that, they do not support backwards compatibility for the models from previous versions. See [here](https://scikit-learn.org/dev/modules/model_persistence.html) and [here](https://github.com/scikit-learn/scikit-learn/issues/14724).

2. The current joblib package available in scikit-learn (0.19.2) deprecates in the latest version. To see if their model persistence page (see first link above), does provide a straightforward solution to this, I tried experimenting by saving models from the current version and tried loading them with the new version. I tried saving them in 2 ways. One using the standard dc.dump() (read dc as nlp.domain_classifier) method provided with mindmeld and the other using the raw joblib package. 

a) When trying to load a model using the latest version of scikit-learn and saved using dc.dump(), we get the following error:
```python
>>> dc.load(model_path='/Users/vidamoda/domain_classifier192.pkl')
/Users/vidamoda/.pyenv/versions/mindmeld/lib/python3.6/site-packages/sklearn/utils/deprecation.py:143: FutureWarning: The sklearn.linear_model.logistic module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.linear_model. Anything that cannot be imported from sklearn.linear_model is now part of the private API.
  warnings.warn(message, FutureWarning)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/vidamoda/MindMeld/my_mm_workspace/mindmeld/mindmeld/components/domain_classifier.py", line 85, in load
    super().load(*args, **kwargs)
  File "/Users/vidamoda/MindMeld/my_mm_workspace/mindmeld/mindmeld/components/classifier.py", line 434, in load
    self._model = joblib.load(model_path)
  File "/Users/vidamoda/.pyenv/versions/mindmeld/lib/python3.6/site-packages/joblib/numpy_pickle.py", line 585, in load
    obj = _unpickle(fobj, filename, mmap_mode)
  File "/Users/vidamoda/.pyenv/versions/mindmeld/lib/python3.6/site-packages/joblib/numpy_pickle.py", line 504, in _unpickle
    obj = unpickler.load()
  File "/Users/vidamoda/.pyenv/versions/3.6.5/lib/python3.6/pickle.py", line 1050, in load
    dispatch[key[0]](self)
  File "/Users/vidamoda/.pyenv/versions/3.6.5/lib/python3.6/pickle.py", line 1338, in load_global
    klass = self.find_class(module, name)
  File "/Users/vidamoda/.pyenv/versions/3.6.5/lib/python3.6/pickle.py", line 1388, in find_class
    __import__(module, level=0)
ModuleNotFoundError: No module named 'sklearn.externals.joblib'
```

This error occurs because of the fact that the module name gets saved in the pickle code when we do a joblib save from the current version of sklearn. So when trying to load the file back, this will look to use that same module, however will not find it because the module is deprecated in the new version. This means that, if we save a model using sklearn.externals.joblib, we will not be able to load it back using the normal joblib package, which effectively means that all models from previous versions will have to be re-trained and re-saved.

b) As an additional check, just to see if I could bypass that save using sklearn.externals.joblib, I also tried saving the same model using normal joblib package and see if that could be loaded and used as a classifier. 
```python
>>> import joblib
>>> dc = joblib.load("/Users/vidamoda/domain_classifier_joblib_192.pkl")
/Users/vidamoda/.pyenv/versions/mindmeld/lib/python3.6/site-packages/sklearn/base.py:334: UserWarning: Trying to unpickle estimator LogisticRegression from version 0.19.2 when using version 0.23.1. This might lead to breaking code or invalid results. Use at your own risk.
  UserWarning)
/Users/vidamoda/.pyenv/versions/mindmeld/lib/python3.6/site-packages/sklearn/utils/deprecation.py:143: FutureWarning: The sklearn.preprocessing.label module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.preprocessing. Anything that cannot be imported from sklearn.preprocessing is now part of the private API.
  warnings.warn(message, FutureWarning)
/Users/vidamoda/.pyenv/versions/mindmeld/lib/python3.6/site-packages/sklearn/base.py:334: UserWarning: Trying to unpickle estimator LabelEncoder from version 0.19.2 when using version 0.23.1. This might lead to breaking code or invalid results. Use at your own risk.
  UserWarning)
/Users/vidamoda/.pyenv/versions/mindmeld/lib/python3.6/site-packages/sklearn/utils/deprecation.py:143: FutureWarning: The sklearn.feature_extraction.dict_vectorizer module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.feature_extraction. Anything that cannot be imported from sklearn.feature_extraction is now part of the private API.
  warnings.warn(message, FutureWarning)
/Users/vidamoda/.pyenv/versions/mindmeld/lib/python3.6/site-packages/sklearn/base.py:334: UserWarning: Trying to unpickle estimator DictVectorizer from version 0.19.2 when using version 0.23.1. This might lead to breaking code or invalid results. Use at your own risk.
  UserWarning)
>>> dc.evaluate()
The tokenizer resource has not been registered to the model. Using default tokenizer.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/vidamoda/MindMeld/my_mm_workspace/mindmeld/mindmeld/components/classifier.py", line 339, in evaluate
    evaluation = self._model.evaluate(queries, labels)
  File "/Users/vidamoda/MindMeld/my_mm_workspace/mindmeld/mindmeld/models/text_models.py", line 123, in evaluate
    predictions = self.predict_proba(examples)
  File "/Users/vidamoda/MindMeld/my_mm_workspace/mindmeld/mindmeld/models/text_models.py", line 212, in predict_proba
    X, _, _ = self.get_feature_matrix(examples, dynamic_resource=dynamic_resource)
  File "/Users/vidamoda/MindMeld/my_mm_workspace/mindmeld/mindmeld/models/text_models.py", line 375, in get_feature_matrix
    self._extract_features(example, dynamic_resource, self.tokenizer)
  File "/Users/vidamoda/MindMeld/my_mm_workspace/mindmeld/mindmeld/models/model.py", line 1093, in _extract_features
    feat_set.update(feat_extractor(example, workspace_resource))
  File "/Users/vidamoda/MindMeld/my_mm_workspace/mindmeld/mindmeld/models/query_features.py", line 927, in _extractor
    for gaz_name, gaz in resources[GAZETTEER_RSC].items():
KeyError: 'gazetteers'
```

While this approach did succeed in loading the model (with a few innocuous warnings), it was not able to successfully evaluate it because the load function Mindmeld provides, also loads the gazetteers and other resources and this does not take care of that. The current resources dictionary is empty and we get a key error for 'gazetteers'. I did try building nlp from scratch and reassigning the domain_classifier object to the loaded object, but this also proved unsuccessful. 

The final conclusion is that, as far as I can tell, there is no straightforward or clean-cut way for us to allow the re-use of models from previous sci-kit learn versions which means that this will be a breaking change. After talking to @vijay120 and others, it seemed to make more sense that we dock this PR for now and probably revisit it when the time comes for an upgrade that will allow for much more features and warrants the need for a breaking change and much newer versions. I've addressed all outstanding review comments and maybe this will help anyone doing an upgrade later down the line. 

Thanks!
@minhtuev @serapio @vijay120 @arushir

PR #164, opened 2020-06-28
Enable user to run 'python -m jukebot converse --verbose' to display intent confidences

owanr:
I couldn't figure out a way to show the confidences (intent probabilities) while conversing, and it would be great if we had that option.

### DIFF
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -1211,7 +1211,7 @@ class Conversation:
         params (FrozenParams): The params returned by the most recent turn.
         force_sync (bool): Force synchronous return for `say()` and `process()` \
             even when app is in async mode.
-	verbose (bool, optional): If True, returns class probabilities along with class \
+        verbose (bool, optional): If True, returns class probabilities along with class \
                 prediction.
     """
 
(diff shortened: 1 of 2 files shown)


minhtuev:
@owanr : can you add in a comment the expected result of your change?

mindmeld/components/dialogue.py
```diff
@@ -1339,6 +1345,7 @@ def process(self, text, params=None, force_sync=False):
             context=self.context,
             frame=self.frame,
             history=self.history,
+            verbose=self.verbose
```
vijay120, line 1348: Change looks good. Can you also add the verbose flag here: https://github.com/cisco/mindmeld/pull/164/files#diff-c21e7ac893d7b0e1d68df2df2ce7ff52R1389

owanr: Sorry, I'm not quite sure which line you're referring to. The link doesn't do anything when I click it. Can you please clarify?

vijay120: I mean line 1377 (old code) and 1384 (new code) of `mindmeld/components/dialogue.py`: 
```
        response = await self._app_manager.parse(
            text,
            params=internal_params,
            context=self.context,
            frame=self.frame,
            history=self.history,
        )
```

owanr: Done! Thank you for the clarification.

vijay120:
@owanr Awesome, thanks for this PR. I am merging this in.

PR #166, opened 2020-07-01
Update md5 hashes for osx, ubuntu16, ubuntu18

alhuang10:
Updating duckling binary md5s. New Duckling includes new Spanish rules from a recent bug fix (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-133623).

Do not merge until Duckling binaries themselves are properly updated.


mindmeld/cli.py
```diff
@@ -37,7 +37,7 @@
 from ._util import blueprint
 from ._version import current as __version__
 from .components import Conversation, QuestionAnswerer
-from .constants import DEVCENTER_URL
+from .constants import DEVCENTER_URL, BINARIES_URL, DUCKLING_VERSION
```
jre21, line 40: No need to keep importing DEVCENTER_URL here, since you just changed the one line in this file that references it.

mindmeld/constants.py
```diff
@@ -12,4 +12,6 @@
 # limitations under the License.
 DEFAULT_TRAIN_SET_REGEX = r"train.*\.txt"
 DEFAULT_TEST_SET_REGEX = r"test.*\.txt"
-DEVCENTER_URL = "https://devcenter.mindmeld.com"
+DEVCENTER_URL = "https://blueprints.mindmeld.com"
```
jre21, line 15: It's probably best to rename this, since "devcenter" isn't a good description of what we're using it for anymore.

mindmeld/cli.py
vijay120: done

mindmeld/constants.py
vijay120: done

jre21:
I took a brief look at the test failures: seems we're getting http connection timeouts when testing embeddings, although I'm not familiar enough with the code to know which url we were trying to hit.

PR #170, opened 2020-07-03
Updated Dockerfiles to use Ubuntu 18.04 and Python 3.6.9 to solve aiohttp Docker build issue

Zozman:
Currently when attempting to build the Docker container build of Mindmeld provided at https://github.com/cisco/mindmeld/blob/master/docker_containers/mindmeld_docker/Dockerfile , the following issue occurs during the Mindmeld package installation step:

mindmeld-build-error

Upon checking the build process it seems that the current form of the container installs Python `3.5.2`, which seems to be the version the `16.04` base image of ubuntu uses.  To solve the issue, this PR updates the base image to `18.04` and makes a couple of adjustments so the build completes, causing Python `3.6.9` to be installed instead, which does not run into the aiohttp issue during the Mindmeld installation step.

### DIFF
--- a/docker_containers/mindmeld_dep_docker/Dockerfile
+++ b/docker_containers/mindmeld_dep_docker/Dockerfile
@@ -8,7 +8,7 @@
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
 # Pull base image.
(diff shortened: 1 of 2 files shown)


minhtuev:
Verified that the builds are working locally.

PR #174, opened 2020-07-08
Fix 4.3 Documentation Errors

ritvikshrivastava:
Documentation issues addressed:

Slot Filling
Banking Assistant
Getting Started with MindMeld


source/userguide/getting_started.rst
```diff
@@ -153,7 +153,6 @@ macOS:
 
    curl https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-6.7.0.tar.gz -o elasticsearch-6.7.0.tar.gz
```
minhtuev, line 154: @arushir : should we recommend ES7 now instead?

vijay120: I think we can stick to 6.7 here.

arushir: yes!

PR #175, opened 2020-07-08
Selected top sys candidate

vijay120:
For queries like "400 hundred", Duckling returns the following entity candidates for amount-of-money:

```
   {
      "body":"four hundred",
      "start":0,
      "value":{
         "value":400,
         "type":"value",
         "unit":"unknown"
      },
      "end":12,
      "dim":"amount-of-money",
      "latent":True
   },
```

and 

```
 {
      "body":"four hundred",
      "start":0,
      "value":{
         "value":5,
         "type":"value",
         "unit":"unknown"
      },
      "end":12,
      "dim":"amount-of-money",
      "latent":True
   },
```

The correct resolve entity is the first one and not the second one. Moreover, Duckling sorts the correct one higher than the incorrect one. 

However, in MindMeld code, we pick the worse candidate since we select the last candidate in the code. We fix this issue in the PR.


mindmeld/system_entity_recognizer.py
```diff
@@ -442,7 +442,11 @@ def resolve_system_entity(self, query, entity_type, span):
             )
 
         if len(entity_type_filtered_candidates) > 0:
-            return entity_type_filtered_candidates[-1]
+            # Duckling sorts most probable entity candidates higher than
+            # the lower probable candidates. So we return the best possible
+            # candidate in this case when multiple duckling candidates are
+            # returned.
+            return entity_type_filtered_candidates[0]
```
serapio, line 449: doesn't this select the wrong candidate for the time grain cases?

vijay120: Do you have an example?

tests/test_markup.py
```diff
@@ -93,21 +93,30 @@ def test_load_markup_error(query_factory, query):
...
 @pytest.mark.load
 @pytest.mark.system
-def test_load_system(query_factory):
+@pytest.mark.parametrize(
+    "query, entity_value, entity_type, entity_start_span, value, unit",
+    [
+        ("show me houses under {600,000 dollars|sys_amount-of-money}",
+         "600,000 dollars", "sys_amount-of-money", 21, 600000, "$"),
+        ("pay me {400|sys_amount-of-money}",
+         "400", "sys_amount-of-money", 7, 400, "unknown"),
```
serapio, line 102: Let's add tests for one or more of these:
join the 2 p.m. meeting
Let's meet at 2 pm on Sunday, August 19 2018
 join the 11:39 meeting```

vijay120: done

PR #185, opened 2020-07-13
Only call ElasticSearch when mapping.json is populated

arushir:
Some developers who are not using the Entity Resolution functionality of MindMeld don't want to get ElasticSearch set up. If an entity exists, ES is by default called, which can be confusing to those who only want to use the Entity Recognition component.

This change checks to see if the mapping.json file is empty when initializing the resolver. If it is, it doesn't make calls to ElasticSearch.

### DIFF
--- a/mindmeld/components/entity_resolver.py
+++ b/mindmeld/components/entity_resolver.py
@@ -79,10 +79,7 @@ class EntityResolver:
         num_canonical_entities = len(
             self._resource_loader.get_entity_map(self.type).get("entities", [])
         )
-        if num_canonical_entities == 0:
-            self._no_canonical_entities = True
-        else:
-            self._no_canonical_entities = False
+        self._no_canonical_entities = num_canonical_entities == 0
 
     @property
     def _es_index_name(self):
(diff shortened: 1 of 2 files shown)


mindmeld/components/entity_resolver.py
```diff
@@ -76,6 +76,14 @@ def __init__(
         self._es_host = es_host
         self._es_config = {"client": es_client, "pid": os.getpid()}
 
+        num_canonical_entities = len(
+            self._resource_loader.get_entity_map(self.type).get("entities", [])
+        )
+        if num_canonical_entities == 0:
```
minhtuev, line 82:
```
self._no_canonical_entities = (num_canonical_entities == 0)

arushir:
@minhtuev I tested it by:
1. stopping ElasticSearch, deleting the mapping.json file for kwik-e-mart, and running
   ```
   python -m kwik_e_mart build
   ```
   and verified that it did not error out due to the lack of ES connection.

2. With the same settings (no ES running and no mapping.json), tested it out in the shell:
   ```
   from mindmeld.components.nlp import NaturalLanguageProcessor
   nlp = NaturalLanguageProcessor(app_path='kwik_e_mart')
   nlp.load()
   er = nlp.domains['store_info'].intents['get_store_hours'].entities['store_name'].entity_resolver
   er.predict(Entity(text='23 elm', entity_type='store_name'))
   ```
   and verified that it returned an empty list

3. re-starting ElasticSearch and verifying it works as expected when the mapping.json does exist.

PR #189, opened 2020-07-14
Update documentation with Optional Dependency

minhtuev:
1. Add a section to manage optional dependency (Duckling, ES)
2. Add instruction to clone sample server

### DIFF
--- a/source/userguide/optional_dependency.rst
+++ b/source/userguide/optional_dependency.rst
@@ -1,0 +1,30 @@
+Managing Optional Dependency
+============================
+
+MindMeld depends on Duckling for numerical parsing and ElasticSearch for Entity Resolution and Question Answering. However if your application does not need these components, you can turn off them by configuring your MindMeld application.
+
+
+Turning off Duckling
+--------------------
+
+To turn off Duckling, specify an empty dictionary for the ``'system_entity_recognizer'`` key:
+
+.. code-block:: python
+
+   NLP_CONFIG = {
+       'system_entity_recognizer': {}
+   }
+
+
+Turning off ElasticSearch for Entity Resolution
+-----------------------------------------------
+
+If you choose not to use Elasticsearch, MindMeld provides a simple baseline version of entity resolution. This Exact Match Model only resolves to an object when the text exactly matches a canonical name or synonym. To use the Exact Match Model, add the following to your app config (``config.py``) located in the top level of your app folder:
+
+.. code-block:: python
+
+    ENTITY_RESOLVER_CONFIG = {
+        'model_type': 'exact_match'
+    }
+
+Similarly when your ``mapping.json`` is empty, the application will not make a call to ElasticSearch.
--- a/source/userguide/optional_dependency.rst
+++ b/source/userguide/optional_dependency.rst
@@ -1,7 +1,7 @@
-Managing Optional Dependency
-============================
+Managing Optional Dependencies
+==============================
 
-MindMeld depends on Duckling for numerical parsing and ElasticSearch for Entity Resolution and Question Answering. However if your application does not need these components, you can turn off them by configuring your MindMeld application.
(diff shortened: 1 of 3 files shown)


source/userguide/optional_dependency.rst
```diff
@@ -0,0 +1,30 @@
...
+
+   NLP_CONFIG = {
+       'system_entity_recognizer': {}
+   }
+
+
+Turning off ElasticSearch for Entity Resolution
+-----------------------------------------------
+
+If you choose not to use Elasticsearch, MindMeld provides a simple baseline version of entity resolution. This Exact Match Model only resolves to an object when the text exactly matches a canonical name or synonym. To use the Exact Match Model, add the following to your app config (``config.py``) located in the top level of your app folder:
```
vijay120, line 22: What happens when one has the exact match model and a populated mapping.json file?

minhtuev: Not sure, @arushir what do you think?

source/userguide/optional_dependency.rst
```diff
@@ -0,0 +1,30 @@
+Managing Optional Dependency
+============================
```
arushir, line 2: Dependency -> Dependencies

source/userguide/optional_dependency.rst
```diff
@@ -0,0 +1,30 @@
...
+
+MindMeld depends on Duckling for numerical parsing and ElasticSearch for Entity Resolution and Question Answering. However if your application does not need these components, you can turn off them by configuring your MindMeld application.
```
arushir, line 4: does not need these components -> does not need the fully optimized version of these components 

you can use Entity Recognition without ES, but it is a very basic resolver

source/userguide/optional_dependency.rst
arushir: With the latest change, you don't need to explicitly 'turn off' ElasticSearch if you are not using the Entity Resolver component. It will not be called if there is no mapping.json

If you do want to do Entity Resolution and do not want to use ElasticSearch, there is a very simple baseline version of entity resolution, as describe in what was pasted above, and also [here](https://www.mindmeld.com/docs/userguide/entity_resolver.html#about-the-exact-match-text-similarity-model). I.e. if there you are using the exact match model and a populated mapping.json `This Exact Match Model only resolves to an object when the text exactly matches a canonical name or synonym.`

arushir: It seems like these distinctions need to be made more explicit or reworded in this doc.

source/userguide/optional_dependency.rst
minhtuev: Done

source/userguide/optional_dependency.rst
```diff
@@ -0,0 +1,30 @@
...
+
+If you choose not to use Elasticsearch, MindMeld provides a simple baseline version of entity resolution. This Exact Match Model only resolves to an object when the text exactly matches a canonical name or synonym. To use the Exact Match Model, add the following to your app config (``config.py``) located in the top level of your app folder:
+
+.. code-block:: python
+
+    ENTITY_RESOLVER_CONFIG = {
+        'model_type': 'exact_match'
+    }
+
+This Exact Match model only resolves to an object when the text exactly matches a canonical name or synonym. Alternatively, when your ``mapping.json`` is empty (i.e. when your application does not include any information for entity resolution), MindMeld will not make a call to Elasticsearch.
```
arushir, line 30: `This Exact Match model only resolves to an object when the text exactly matches a canonical name or synonym.` - This line is duplicated in the paragraph above and can be removed here.

arushir: Alternatively, when your ``mapping.json`` is empty (i.e. when your application does not include any information for entity resolution), MindMeld will not make a call to Elasticsearch. -> Note that if your application is not leveraging the Entity Resolver, i.e. your ``mapping.json`` files are empty, then MindMeld will not make calls to ElasticSearch.

minhtuev: Done

PR #190, opened 2020-07-15
Updates for slot-filling

ritvikshrivastava:
Adds support for directly invoking the slot-filling functionality without a decorator. Can be used by developers in two ways:

1. Just invoking a form directly inside the handler.
```
@app.handle(intent="transfer_money")
def transfer_money_handler(request, responder):
    AutoEntityFilling(handler=transfer_money_followup_handler, form=transfer_form, app=app).invoke(request, responder)

def transfer_money_followup_handler(request, responder):
    <continue logic after completed form>
```

2. Having an initial form through ``@app.auto_fill()``, and a follow up form inside the handler. This can be used for cases where a secondary form might be needed based on the results of the first (such as Lindsey's Spanish Healthcare Assistant).
```
@app.auto_fill(intent="transfer_money", form=form_1)
def transfer_money_handler(request, responder):
    <checks from the first form>
    AutoEntityFilling(handler=transfer_money_followup_handler, form=form_2, app=app).invoke(request, responder)

def transfer_money_followup_handler(request, responder):
    <continue logic after completed form_2>
```

### DIFF
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -706,7 +706,7 @@ class DialogueFlow(DialogueManager):
         return handler
 
 
-class AutoEntityFilling:   # pylint: disable=R0902
+class AutoEntityFilling:
     """A special dialogue flow sublcass to implement Automatic Entitiy (Slot) Filling
     (AEF) that allows developers to prompt users for completing the missing
     requirements for entity slots.
@@ -722,9 +722,8 @@ class AutoEntityFilling:   # pylint: disable=R0902
         self._app = app
         self._handler = handler
         self._form = form
-        self._local_form = None
+        self._local_entity_form = None
         self._prompt_turn = None
-        self._invoked = False
         self._previous_rule = None
         self._check_attr()
 
@@ -766,7 +765,7 @@ class AutoEntityFilling:   # pylint: disable=R0902
     def _exit_flow(self, responder):
         """Exits this flow and clears the related parameter for re-usability"""
         self._prompt_turn = None
-        self._local_form = None
+        self._local_entity_form = None
         responder.params.allowed_intents = tuple()
         responder.exit_flow()
 
@@ -888,7 +887,7 @@ class AutoEntityFilling:   # pylint: disable=R0902
             entity_type = entity["type"]
             role = entity["role"]
 
-            for slot in self._local_form:
+            for slot in self._local_entity_form:
                 if entity_type == slot.entity:
                     if (slot.role is None) or (role == slot.role):
                         slot.value = entity
@@ -902,28 +901,12 @@ class AutoEntityFilling:   # pylint: disable=R0902
             text=request.text,
             domain=request.domain,
             intent=request.intent,
-            entities=[slot.value for slot in self._local_form],
+            entities=[slot.value for slot in self._local_entity_form],
             context=request.context or {},
             history=request.history or [],
             frame=responder.frame or {},
             params=request.params,
         )
-
-        if self._invoked:
-            kwargs = {'targeted_only': False}
-            name = self._handler.__name__
-            if self._previous_rule:
-                # The following is to reset the previously modified dialogue rule back to
-                # its original state. If the the rule cannot be added, the mapping is changed back.
-                # This is done so that the rule conditions map to the original handler instead of
-                # mapping to the auto-fill standalone call at all times.
-                try:
-                    self._app.app_manager.dialogue_manager.add_dialogue_rule(
-                        name, self._previous_rule, **kwargs
-                    )
-                except AssertionError:
-                    self._app.app_manager.dialogue_manager.handler_map[name] = self._previous_rule
-                self._previous_rule = None
 
         self._exit_flow(responder)
 
@@ -982,16 +965,16 @@ class AutoEntityFilling:   # pylint: disable=R0902
 
         self._set_next_turn(request, responder)
 
-        if self._prompt_turn is None or self._local_form is None:
+        if self._prompt_turn is None or self._local_entity_form is None:
             # Entering the flow
             self._prompt_turn = True
-            self._local_form = copy.deepcopy(self._entity_form)
+            self._local_entity_form = copy.deepcopy(self._entity_form)
             self._retry_attempts = 0
 
             # Fill the form with the entities in the first query
             self._initial_fill(request)
 
-        for slot in self._local_form:
+        for slot in self._local_entity_form:
 
             if not slot.value:
                 # check if user has been prompted for this entity slot
@@ -1034,8 +1017,6 @@ class AutoEntityFilling:   # pylint: disable=R0902
         """
         Invoke slot-filling as a direct call without requiring a decorator.
         """
-        self._invoked = True
-
         # ensures that the slot-filling function is targeted.
         kwargs = {'targeted_only': True}
 
@@ -1055,6 +1036,9 @@ class AutoEntityFilling:   # pylint: disable=R0902
             return
 
         try:
+            # sets a dialogue rule for the handler passed in this invoke call to iteratively call
+            # the slot-filling flow till completion or exit. This rule is added temporarily for this
+            # flow and reset for the handler with every new invoke call.
             self._app.app_manager.dialogue_manager.add_dialogue_rule(name, self.__call__, **kwargs)
         except AssertionError:
             self._app.app_manager.dialogue_manager.handler_map[name] = self.__call__
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -720,7 +720,7 @@ class AutoEntityFilling:   # pylint: disable=R0902
 
     def __init__(self, handler, form, app):
         self._app = app
-        self._entrance_handler = handler
+        self._handler = handler
         self._form = form
         self._local_form = None
         self._prompt_turn = None
@@ -761,7 +761,7 @@ class AutoEntityFilling:   # pylint: disable=R0902
         responder.params.allowed_intents = tuple(
             ["{}.{}".format(request.domain, request.intent)]
         )
-        responder.params.target_dialogue_state = self._entrance_handler.__name__
+        responder.params.target_dialogue_state = self._handler.__name__
 
     def _exit_flow(self, responder):
         """Exits this flow and clears the related parameter for re-usability"""
@@ -910,9 +910,13 @@ class AutoEntityFilling:   # pylint: disable=R0902
         )
 
         if self._invoked:
-            kwargs = {'domain': request.domain, 'intent': request.intent}
-            name = self._entrance_handler.__name__
+            kwargs = {'targeted_only': False}
+            name = self._handler.__name__
             if self._previous_rule:
+                # The following is to reset the previously modified dialogue rule back to
+                # its original state. If the the rule cannot be added, the mapping is changed back.
+                # This is done so that the rule conditions map to the original handler instead of
+                # mapping to the auto-fill standalone call at all times.
                 try:
                     self._app.app_manager.dialogue_manager.add_dialogue_rule(
                         name, self._previous_rule, **kwargs
@@ -928,10 +932,10 @@ class AutoEntityFilling:   # pylint: disable=R0902
         return self._end_slot_fill_sync(request, responder)
 
     def _end_slot_fill_sync(self, request, responder):
-        return self._entrance_handler(request, responder)
+        return self._handler(request, responder)
 
     async def _end_slot_fill_async(self, request, responder):
-        return await self._entrance_handler(request, responder)
+        return await self._handler(request, responder)
 
     def _prompt_slot(self, responder, nlr):
         responder.reply(nlr)
@@ -1031,17 +1035,23 @@ class AutoEntityFilling:   # pylint: disable=R0902
         Invoke slot-filling as a direct call without requiring a decorator.
         """
         self._invoked = True
-        kwargs = {'domain': request.domain, 'intent': request.intent}
+
+        # ensures that the slot-filling function is targeted.
+        kwargs = {'targeted_only': True}
 
         # Set dialogue rule to auto_fill and
-        # store current rule for resetting once invoke is complete
+        # store current rule for resetting once invoke is complete.
 
-        name = self._entrance_handler.__name__
+        name = self._handler.__name__
 
         try:
+            # fetches the name of the function that invoked this class.
             _called_from = sys._getframe().f_back.f_code.co_name
             self._previous_rule = self._app.app_manager.dialogue_manager.handler_map[_called_from]
-        except (KeyError, AttributeError):
+        except (KeyError, AttributeError) as e:
+            self._logger.warning(
+                "Auto-Fill invoke failed with warning: '{}'".format(str(e))
+            )
             return
 
         try:
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -12,7 +12,6 @@
 # limitations under the License.
 
 """This module contains the dialogue manager component of MindMeld"""
-import sys
 import asyncio
 import copy
 import json
@@ -710,21 +709,22 @@ class AutoEntityFilling:
     """A special dialogue flow sublcass to implement Automatic Entitiy (Slot) Filling
     (AEF) that allows developers to prompt users for completing the missing
     requirements for entity slots.
-
-    Attributes:
-        app (Application): The application that initializes this flow.
     """
 
     _logger = mod_logger.getChild("AutoEntityFilling")
     """Class logger."""
 
     def __init__(self, handler, form, app):
+        """
+        handler (func): The function to which control is returned after completion of flow.
+        form (dict): Developer-defined slot-filling form.
+        app (Application): The application that initializes this flow.
+        """
         self._app = app
         self._handler = handler
         self._form = form
         self._local_entity_form = None
         self._prompt_turn = None
-        self._previous_rule = None
         self._check_attr()
 
     def _check_attr(self):
@@ -1020,20 +1020,7 @@ class AutoEntityFilling:
         # ensures that the slot-filling function is targeted.
         kwargs = {'targeted_only': True}
 
-        # Set dialogue rule to auto_fill and
-        # store current rule for resetting once invoke is complete.
-
         name = self._handler.__name__
-
-        try:
-            # fetches the name of the function that invoked this class.
-            _called_from = sys._getframe().f_back.f_code.co_name
-            self._previous_rule = self._app.app_manager.dialogue_manager.handler_map[_called_from]
-        except (KeyError, AttributeError) as e:
-            self._logger.warning(
-                "Auto-Fill invoke failed with warning: '{}'".format(str(e))
-            )
-            return
 
         try:
             # sets a dialogue rule for the handler passed in this invoke call to iteratively call
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -706,7 +706,7 @@ class DialogueFlow(DialogueManager):
         return handler
 
 
-class AutoEntityFilling:
+class AutoEntityFilling:   # pylint: disable=R0902
     """A special dialogue flow sublcass to implement Automatic Entitiy (Slot) Filling
     (AEF) that allows developers to prompt users for completing the missing
     requirements for entity slots.
@@ -725,6 +725,7 @@ class AutoEntityFilling:
         self._local_form = None
         self._prompt_turn = None
         self._invoked = False
+        self._previous_rule = None
         self._check_attr()
 
     def _check_attr(self):
@@ -911,16 +912,14 @@ class AutoEntityFilling:
         if self._invoked:
             kwargs = {'domain': request.domain, 'intent': request.intent}
             name = self._entrance_handler.__name__
-            if "previous_rule" in responder.frame:
+            if self._previous_rule:
                 try:
                     self._app.app_manager.dialogue_manager.add_dialogue_rule(
-                        name, responder.frame["previous_rule"], **kwargs
+                        name, self._previous_rule, **kwargs
                     )
                 except AssertionError:
-                    self._app.app_manager.dialogue_manager.handler_map[name] = responder.frame[
-                        "previous_rule"
-                    ]
-                responder.frame["previous_rule"] = None
+                    self._app.app_manager.dialogue_manager.handler_map[name] = self._previous_rule
+                self._previous_rule = None
 
         self._exit_flow(responder)
 
@@ -1038,12 +1037,12 @@ class AutoEntityFilling:
         # store current rule for resetting once invoke is complete
 
         name = self._entrance_handler.__name__
-        _called_from = sys._getframe().f_back.f_code.co_name
 
-        responder.frame["_called_from"] = _called_from
-        responder.frame[
-            "previous_rule"
-        ] = self._app.app_manager.dialogue_manager.handler_map[_called_from]
+        try:
+            _called_from = sys._getframe().f_back.f_code.co_name
+            self._previous_rule = self._app.app_manager.dialogue_manager.handler_map[_called_from]
+        except (KeyError, AttributeError):
+            return
 
         try:
             self._app.app_manager.dialogue_manager.add_dialogue_rule(name, self.__call__, **kwargs)
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -717,12 +717,13 @@ class AutoEntityFilling:
     _logger = mod_logger.getChild("AutoEntityFilling")
     """Class logger."""
 
-    def __init__(self, entrance_handler, form, app):
+    def __init__(self, handler, form, app, clear_prev=False):
         self._app = app
-        self._entrance_handler = entrance_handler
+        self._entrance_handler = handler
         self._form = form
         self._local_form = None
         self._prompt_turn = None
+        self._clear_prev = clear_prev
         self._check_attr()
 
     def _check_attr(self):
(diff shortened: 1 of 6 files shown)


mindmeld/components/dialogue.py
```diff
@@ -717,12 +717,13 @@ class AutoEntityFilling:
...
 
-    def __init__(self, entrance_handler, form, app):
+    def __init__(self, handler, form, app, clear_prev=False):
         self._app = app
-        self._entrance_handler = entrance_handler
+        self._entrance_handler = handler
         self._form = form
         self._local_form = None
         self._prompt_turn = None
+        self._clear_prev = clear_prev
```
serapio, line 726: Let's call it `clear_previous` for clarity.

mindmeld/components/dialogue.py
```diff
@@ -1011,6 +1013,56 @@ async def call_async(self, request, responder):
...
+        _reqd = []
+        if request.entities:
+            for e in request.entities:
+                _captured.append(e['type'])
+
+        for e in self._entity_form:
+            _reqd.append(e.entity)
+
+        # check if already filled in previous turn, reset call back to original
+        if sorted(_captured) == sorted(_reqd):
```
serapio, line 1028: If the order of the lists doesn't matter, could just use sets.

mindmeld/components/dialogue.py
```diff
@@ -876,20 +877,21 @@ def _validate(self, request, slot):
 
     def _initial_fill(self, request):
         """Performs the first pass and fills the entity form with entity values available
-        in the initial query.
+        in the initial query. Bypassed if `clear_prev` is set to True.
 
         Args:
             request (Request): The request object.
         """
-        for entity in request.entities:
```
vijay120, line 884: To prevent this indentation sprawl, you could do this:
```
if self._clear_prev:
    return

....

```

mindmeld/components/dialogue.py
ritvikshrivastava: Removed this.

mindmeld/components/dialogue.py
ritvikshrivastava: Not needed anymore, removed this as well.

mindmeld/components/dialogue.py
ritvikshrivastava: Reverted some changes, I don't think this is needed anymore.

mindmeld/components/dialogue.py
```diff
@@ -906,6 +908,20 @@ def _end_slot_fill(self, request, responder, async_mode):
...
+            if "previous_rule" in responder.frame:
+                try:
+                    self._app.app_manager.dialogue_manager.add_dialogue_rule(
+                        name, responder.frame["previous_rule"], **kwargs
+                    )
+                except AssertionError:
+                    self._app.app_manager.dialogue_manager.handler_map[name] = responder.frame[
+                        "previous_rule"
+                    ]
+                responder.frame["previous_rule"] = None
```
vijay120, line 923: Can you pop it? `responder.pop("previous_rule", None)`

mindmeld/components/dialogue.py
```diff
@@ -1011,6 +1027,38 @@ async def call_async(self, request, responder):
...
+        Invoke slot-filling as a direct call without requiring a decorator.
+        """
+        self._invoked = True
+        kwargs = {'domain': request.domain, 'intent': request.intent}
+
+        # Set dialogue rule to auto_fill and
+        # store current rule for resetting once invoke is complete
+
+        name = self._entrance_handler.__name__
+        _called_from = sys._getframe().f_back.f_code.co_name
```
vijay120, line 1041: We need to some error checking here. `sys._getframe().f_back` could be `None` for example.

mindmeld/components/dialogue.py
```diff
@@ -1011,6 +1027,38 @@ async def call_async(self, request, responder):
...
+
+        responder.frame["_called_from"] = _called_from
+        responder.frame[
+            "previous_rule"
+        ] = self._app.app_manager.dialogue_manager.handler_map[_called_from]
```
vijay120, line 1046: `_called_from ` could not be in the map

mindmeld/components/dialogue.py
```diff
@@ -1011,6 +1027,38 @@ async def call_async(self, request, responder):
...
+
+        try:
+            self._app.app_manager.dialogue_manager.add_dialogue_rule(name, self.__call__, **kwargs)
+        except AssertionError:
+            self._app.app_manager.dialogue_manager.handler_map[name] = self.__call__
```
vijay120, line 1051: What does the exception case do?

ritvikshrivastava: It bypasses the exception in adding new dialogue rules and updates the rule regardless.

vijay120: Is it possible to never collide with an existing handler? What if the name was unique?

mindmeld/components/dialogue.py
ritvikshrivastava: I think it will always return between a <function> and a <module> if called from outside. I'll check.

mindmeld/components/dialogue.py
```diff
@@ -1011,6 +1027,38 @@ async def call_async(self, request, responder):
...
+        """
+        Invoke slot-filling as a direct call without requiring a decorator.
+        """
+        self._invoked = True
+        kwargs = {'domain': request.domain, 'intent': request.intent}
+
+        # Set dialogue rule to auto_fill and
+        # store current rule for resetting once invoke is complete
+
+        name = self._entrance_handler.__name__
```
minhtuev, line 1040: probably should add a prefix or suffix with "autofill"

mindmeld/components/dialogue.py
ritvikshrivastava: This is for the case, when they restart the same flow. If the user calls say 'transfer' twice, this flow needs to be called again, and possibly the sub-flow again. This causes that exception.

mindmeld/components/dialogue.py
ritvikshrivastava: This actually is the name of the follow-up handler that it redirects to once the flow is finished.

mindmeld/components/dialogue.py
minhtuev: You can also do: `del responder.frame[''previous_rule"]` since `pop` will throw an error if the key does not exist.

ritvikshrivastava: I thought the second argument (None) in pop is default for that case itself?

mindmeld/components/dialogue.py
```diff
@@ -906,6 +909,18 @@ def _end_slot_fill(self, request, responder, async_mode):
             params=request.params,
         )
 
+        if self._invoked:
+            kwargs = {'domain': request.domain, 'intent': request.intent}
+            name = self._entrance_handler.__name__
+            if self._previous_rule:
+                try:
```
vijay120, line 916: Could you add a comment here saying why we do this try/exception block. It would be useful for maintaining this code.

mindmeld/components/dialogue.py
```diff
@@ -1011,6 +1026,38 @@ async def call_async(self, request, responder):
...
+        # Set dialogue rule to auto_fill and
+        # store current rule for resetting once invoke is complete
+
+        name = self._entrance_handler.__name__
+
+        try:
+            _called_from = sys._getframe().f_back.f_code.co_name
+            self._previous_rule = self._app.app_manager.dialogue_manager.handler_map[_called_from]
+        except (KeyError, AttributeError):
+            return
```
vijay120, line 1045: Add an logger error here.

source/userguide/dm.rst
```diff
@@ -659,6 +658,102 @@ For the use case of transferring money in a banking assistant application, the f
...
+    from mindmeld.core import FormEntity
+
+    form_transfermoney_1 = {
+        'entities':[
+            FormEntity(
+                entity='account_type',
+                role='account_from',
+                responses=['Sure. Transfer from which account?']
+                ),
+            ]
```
minhtuev, line 680: There are no exit keys and exit msg in form 1?

ritvikshrivastava: That's an optional param.

mindmeld/components/dialogue.py
```diff
@@ -906,6 +909,18 @@ def _end_slot_fill(self, request, responder, async_mode):
             params=request.params,
         )
 
+        if self._invoked:
+            kwargs = {'domain': request.domain, 'intent': request.intent}
```
minhtuev, line 913: We should add the full list of args here:
https://github.com/cisco/mindmeld/blob/master/mindmeld/components/dialogue.py#L104

```
        key_kwargs = (
            ("domain",),
            ("intent",),
            ("has_entity", "has_entities"),
            ("targeted_only",),
            ("default",),
        )

File: mindmeld/components/dialogue.py [hunk 13]
```diff
@@ -717,12 +718,14 @@ class AutoEntityFilling:
     _logger = mod_logger.getChild("AutoEntityFilling")
     """Class logger."""
 
-    def __init__(self, entrance_handler, form, app):
+    def __init__(self, handler, form, app):
         self._app = app
-        self._entrance_handler = entrance_handler
+        self._entrance_handler = handler
```
minhtuev, line 723: Should we rename this to just handler?

mindmeld/components/dialogue.py
```diff
@@ -705,7 +706,7 @@ def _get_dialogue_handler(self, dialogue_state):
         return handler
 
 
-class AutoEntityFilling:
+class AutoEntityFilling:   # pylint: disable=R0902
```
minhtuev, line 709: We can avoid this lint issue by grouping together the attributes (form & local form, etc)

mindmeld/components/dialogue.py
minhtuev: We should also update the function doc

mindmeld/components/dialogue.py
```diff
@@ -1011,6 +1026,38 @@ async def call_async(self, request, responder):
         """
         self(request, responder)
 
+    def invoke(self, request, responder):
+        """
+        Invoke slot-filling as a direct call without requiring a decorator.
+        """
+        self._invoked = True
+        kwargs = {'domain': request.domain, 'intent': request.intent}
```
minhtuev, line 1034: Same as above

mindmeld/components/dialogue.py
```diff
@@ -1011,6 +1026,38 @@ async def call_async(self, request, responder):
...
+
+        # Set dialogue rule to auto_fill and
+        # store current rule for resetting once invoke is complete
+
+        name = self._entrance_handler.__name__
+
+        try:
+            _called_from = sys._getframe().f_back.f_code.co_name
```
minhtuev, line 1042: We should add doc here

mindmeld/components/dialogue.py
minhtuev: We should add documentation to this exception

mindmeld/components/dialogue.py
minhtuev: Unless we have a conscious decision only to use domain & intent for mapping

minhtuev: Hmm nvm, I think I misread this section, you are mapping the domain and intent to the automatic slot filling handler itself.

mindmeld/components/dialogue.py
minhtuev: This probably means that the handler is unique and should not be reused right?

mindmeld/components/dialogue.py
```diff
@@ -1011,6 +1026,38 @@ async def call_async(self, request, responder):
...
+            self._previous_rule = self._app.app_manager.dialogue_manager.handler_map[_called_from]
+        except (KeyError, AttributeError):
+            return
+
+        try:
+            self._app.app_manager.dialogue_manager.add_dialogue_rule(name, self.__call__, **kwargs)
```
minhtuev, line 1048: Can we remove this rule after slot filing is done?

source/userguide/dm.rst
minhtuev: Probably for completeness we should add them to avoid confusion on the side of the developer

source/userguide/dm.rst
```diff
@@ -659,6 +658,102 @@ For the use case of transferring money in a banking assistant application, the f
...
+
+.. note::
+   
+    * The order of entities provided in the ``entities`` list in the form is important as the slots will be prompted in that order. 
+    .. |br|
+
+    * All entities that are required by the slot-filling form for an intent should be covered through example queries in the training files for that intent.
+    .. |br|    
+   
+    * For better training the entity recognizer corresponding to the slot-filling intent, a separate training file ``train_label_set`` covering examples of the entities to be captured by the form can be defined. You can find more details about defining this file and modifying the entity recognizer `here <https://www.mindmeld.com/docs/userguide/entity_recognizer.html>`_. This also allows intent and domain classifiers to be trained independently of such queries and learn appropriate context.
```
minhtuev, line 754: You should use `::doc::` directive when you reference an internal link

mindmeld/components/dialogue.py
ritvikshrivastava: Yup, that's exactly what the 916-922 block in ``_end_slot_fill`` does. If we don't remove it, the actions can't be repeated later :)

mindmeld/components/dialogue.py
minhtuev: I think we only need to set `targeted_only` to True and set the target dialogue state.

ritvikshrivastava: thanks for the help! 😄

mindmeld/components/dialogue.py
ritvikshrivastava: I looked into that, but even with comments, it might get confusing to reuse/group attributes while maintaining it at a later stage.

mindmeld/components/dialogue.py
ritvikshrivastava: Simplified the logic a bit, now it doesn't remove the rule anymore. It now sets it only facing the secondary function (and now the rule is also targeted), so this will stay in the set of rules but not hinder with any other rule in the app. I don't think there is any need to explicitly get rid of it.

mindmeld/components/dialogue.py
ritvikshrivastava: Ignore the above. The simplified logic for invoke gets rid of the extra invoke attribute.

source/userguide/dm.rst
```diff
@@ -572,8 +572,11 @@ As shown here, you can use the dialog flow functionality to effectively craft co
 Automatic Slot Filling for Entities
 -----------------------------------
 
-MindMeld provides a useful functionality for automatically prompting the user for missing entities or slots required to fulfill an intent. This is done via applying the ``@app.auto_fill`` decorator for the dialogue state handler requiring entities to be obtained prior to applying the functionality defined within.
+MindMeld provides a useful functionality for automatically prompting the user for missing entities or slots required to fulfill an intent. This is done via either applying the ``@app.auto_fill`` decorator to any dialogue state handler requiring entities to be obtained prior to applying the functionality defined within, or directly invoking a call to this feature at any point inside the handler.
```
serapio, line 575: this is pretty hard to read. maybe the two options should be bulleted list?

source/userguide/dm.rst
```diff
@@ -659,6 +658,106 @@ For the use case of transferring money in a banking assistant application, the f
                    "{account_from} to a {account_to}. Is that right?"]
         responder.reply(replies)
 
+
+Subset forms and invoking slot-filling within in a dialogue handler
```
serapio, line 662:
```suggestion
Subset forms and invoking slot-filling within a dialogue handler
```

source/userguide/dm.rst
```diff
@@ -659,6 +658,106 @@ For the use case of transferring money in a banking assistant application, the f
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+To support dynamically iterating through a form, we suggest the use of subset forms. Consider a case in which the initial form fetches a certain set of entities, and the flow ahead depends on the values of those captured entities. This requires capturing of further entities based on those values. For this, we recommend using subset forms.
```
serapio, line 664: why call these "subset forms"? it sounds to me like it would mean a form that only has a subset of the entities in another form. "subform" would work better.

ritvikshrivastava: That's a great suggestion. I updated this and addressed your other comments too! @serapio

mindmeld/components/dialogue.py
```diff
@@ -717,12 +718,13 @@ class AutoEntityFilling:
     _logger = mod_logger.getChild("AutoEntityFilling")
     """Class logger."""
 
-    def __init__(self, entrance_handler, form, app):
+    def __init__(self, handler, form, app):
```
minhtuev, line 721: We should update AutoEntityFiling doc (line 714)

mindmeld/components/dialogue.py
minhtuev: In the case where the user says "transfer" twice, do we need to explicitly remap the function? I think it is still the same function right?

mindmeld/components/dialogue.py
ritvikshrivastava: Even if all attributes are private?

tests/components/test_auto_fill.py
```diff
@@ -95,6 +98,47 @@ def test_auto_fill_validation_missing_entities(kwik_e_mart_app, qa_kwik_e_mart):
     )
 
 
+def handler_sub(request, responder):
```
minhtuev, line 101: To verify that ```handler_sub``` is invoked, you can actually pass in a mock function and assert that the mock function is called.

```
In [11]: from mock import MagicMock

In [12]: f = MagicMock()

In [15]: f.assert_called_once_with(1, 2)
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-15-4231415d7976> in <module>
----> 1 f.assert_called_once_with(1, 2)

~/.pyenv/versions/3.6.8/envs/mm/lib/python3.6/site-packages/mock/mock.py in assert_called_once_with(_mock_self, *args, **kwargs)
    945             msg = ("Expected '%s' to be called once. Called %s times." %
    946                    (self._mock_name or 'mock', self.call_count))
--> 947             raise AssertionError(msg)
    948         return self.assert_called_with(*args, **kwargs)
    949

AssertionError: Expected 'mock' to be called once. Called 0 times.

In [16]: f(1, 2)
Out[16]: <MagicMock name='mock()' id='4619359232'>

In [17]: In [15]: f.assert_called_once_with(1, 2)

File: source/userguide/dm.rst [hunk 24]
```diff
@@ -572,8 +572,14 @@ As shown here, you can use the dialog flow functionality to effectively craft co
 Automatic Slot Filling for Entities
 -----------------------------------
 
-MindMeld provides a useful functionality for automatically prompting the user for missing entities or slots required to fulfill an intent. This is done via applying the ``@app.auto_fill`` decorator for the dialogue state handler requiring entities to be obtained prior to applying the functionality defined within.
+MindMeld provides a useful functionality for automatically prompting the user for missing entities or slots required to fulfill an intent. This can done via either:
 
+- Applying the ``@app.auto_fill`` decorator to any dialogue state handler or
+- Directly invoking a slot-filling call inside a handler or
```
minhtuev, line 578:
```suggestion
- Directly invoking a slot-filling call inside a handler
```

source/userguide/dm.rst
```diff
@@ -572,8 +572,14 @@ As shown here, you can use the dialog flow functionality to effectively craft co
 Automatic Slot Filling for Entities
 -----------------------------------
 
-MindMeld provides a useful functionality for automatically prompting the user for missing entities or slots required to fulfill an intent. This is done via applying the ``@app.auto_fill`` decorator for the dialogue state handler requiring entities to be obtained prior to applying the functionality defined within.
+MindMeld provides a useful functionality for automatically prompting the user for missing entities or slots required to fulfill an intent. This can done via either:
 
+- Applying the ``@app.auto_fill`` decorator to any dialogue state handler or
```
minhtuev, line 577:
```suggestion
- Applying the ``@app.auto_fill`` decorator to any dialogue state handler
```

source/userguide/dm.rst
```diff
@@ -659,6 +661,106 @@ For the use case of transferring money in a banking assistant application, the f
                    "{account_from} to a {account_to}. Is that right?"]
         responder.reply(replies)
 
+
+Subforms and invoking slot-filling within a dialogue handler
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+To support dynamic iteration through a form, we suggest the use of additional subforms or follow-up forms. Consider a case in which the initial form fetches a certain set of entities, and the flow ahead depends on the values of those captured entities. This requires capturing of further entities based on those values. For this, we recommend using further subforms.
+
```
minhtuev, line 668: For symmetry, we should add:
```
**Examples**

File: source/userguide/dm.rst [hunk 27]
```diff
@@ -659,6 +661,106 @@ For the use case of transferring money in a banking assistant application, the f
...
+Take into consideration the previous example. Now, instead of a single pass form to capture all entities, say we want to check whether the account from which money is withdrawn is a savings account or not. If yes, then continue to fetch the other entities (using follow-up form ``form_transfermoney_2``), otherwise send appropriate response.
```
minhtuev, line 669: Let's consider the previous example.
```suggestion
Let's consider the previous example. Now, instead of a single pass form to capture all entities, suppose we want to check whether the account from which money is withdrawn is a savings account or not. If yes, then continue to fetch the other entities (using follow-up form ``form_transfermoney_2``), otherwise, send an appropriate response.
```

source/userguide/dm.rst
```diff
@@ -659,6 +661,106 @@ For the use case of transferring money in a banking assistant application, the f
                    "{account_from} to a {account_to}. Is that right?"]
         responder.reply(replies)
 
+
+Subforms and invoking slot-filling within a dialogue handler
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+To support dynamic iteration through a form, we suggest the use of additional subforms or follow-up forms. Consider a case in which the initial form fetches a certain set of entities, and the flow ahead depends on the values of those captured entities. This requires capturing of further entities based on those values. For this, we recommend using further subforms.
```
minhtuev, line 667:
```suggestion
For more complex interactions (such as dynamic iteration through a form), we can use additional subforms or follow-up forms. Consider a case in which the initial form fetches a certain set of entities, and the flow ahead depends on the values of those captured entities. This requires the capturing of further entities based on those values.
```

source/userguide/dm.rst
```diff
@@ -659,6 +661,106 @@ For the use case of transferring money in a banking assistant application, the f
...
+    from mindmeld.components.dialogue import AutoEntityFilling
+    from mindmeld.core import FormEntity
+
+    form_transfermoney_1 = {
+        'entities':[
+            FormEntity(
+                entity='account_type',
+                role='account_from',
+                responses=['Sure. Transfer from which account?']
+                ),
```
minhtuev, line 682: hints=['checking', 'checkings']

source/userguide/dm.rst
```diff
@@ -659,6 +661,106 @@ For the use case of transferring money in a banking assistant application, the f
...
+        }
+
+    form_transfermoney_2 = {
+        'entities':[
+            FormEntity(
+                entity='account_type',
+                role='account_to',
+                responses=['To which account?'],
+                retry_response=["That account is not correct. Transfer to which account?"]
+                hints=['checking', 'checkings']
```
minhtuev, line 696: remove checking?

source/userguide/dm.rst
```diff
@@ -659,6 +661,106 @@ For the use case of transferring money in a banking assistant application, the f
...
+                ),
+            ],
+        'max_retries': 1,
+        'exit_keys': ['cancel', 'quit', 'exit'],
+        'exit_msg': "Sorry I cannot help you with this transfer. Please try again."
+        }
+
+    @app.auto_fill(intent="transfer_money", form=form_transfermoney_1)
+    def transfer_money_handler(request, responder):
+        account = next((e['value'][0]['cname'] for e in request.entities if e['type'] == 'account_type'), None)
```
minhtuev, line 710: we should add a doc to explain this function

source/userguide/dm.rst
```diff
@@ -659,6 +661,106 @@ For the use case of transferring money in a banking assistant application, the f
...
+        if account == 'savings':
+            responder.frame['account_from'] = account
+            AutoEntityFilling(handler=transfer_money_followup_handler, form=form_transfermoney_2, app=app).invoke(request, responder)
+        else:
+            responder.reply('Sorry, you can only transfer money from a savings account.')
+
+    def transfer_money_followup_handler(request, responder):
```
minhtuev, line 717: same here

source/userguide/dm.rst
```diff
@@ -659,6 +661,106 @@ For the use case of transferring money in a banking assistant application, the f
...
+
+Alternatively, the standalone call to this feature can be called independently of the auto_fill decorated handler as well. Consider the following use-case to check account balance:
+
+.. code:: python
+
+    @app.handle(intent="check_balances")
+    def check_balance(request, responder):
+        user = User(request) # fetch user information
+        responder.frame['user'] = user
+        AutoEntityFilling(handler=check_balance_followup_handler, form=balance_form, app=app).invoke(request, responder)
```
minhtuev, line 739: do we define balance_form anywhere?

mindmeld/components/dialogue.py
minhtuev: I see, maybe update to remove attribute "app" and just create a docstring for ``__init__`` instead.

source/userguide/dm.rst
```diff
@@ -572,8 +572,14 @@ As shown here, you can use the dialog flow functionality to effectively craft co
 Automatic Slot Filling for Entities
 -----------------------------------
 
-MindMeld provides a useful functionality for automatically prompting the user for missing entities or slots required to fulfill an intent. This is done via applying the ``@app.auto_fill`` decorator for the dialogue state handler requiring entities to be obtained prior to applying the functionality defined within.
+MindMeld provides a useful functionality for automatically prompting the user for missing entities or slots required to fulfill an intent. This can done via either:
```
arushir, line 575: can `be` done

source/userguide/dm.rst
arushir, line 578: Remove the `or` at the end of this line

minhtuev:
Looks good, added some comments!

source/userguide/dm.rst
ritvikshrivastava: No, this one is without any hints. The check happens inside the dialogue handler for 'savings', if you go through the flow.

source/userguide/dm.rst
ritvikshrivastava: Done

tests/components/test_auto_fill.py
ritvikshrivastava: Awesome, added.

mindmeld/components/dialogue.py
ritvikshrivastava: I tried out a case where the same secondary handler could be used by two different AutoFill functions. This causes the mapping to shift, if I'm not wrong. So just ensuring that it is mapped for every invoke.

PR #196, opened 2020-07-25
don't try to load a system entity emap

serapio:
The entity resolver has been trying to count canonical entities for system entities, which generates a bunch of warnings. The end behavior is still correct, since the code doesn't actually depend on the no_canonical_entities property for system entities. It just is doing extra file io and generating logs.
```
Entity data file not found at '/.../entities/sys_duration/gazetteer.txt'. Proceeding with empty entity data.
Entity mapping file not found at '/.../entities/sys_duration/mapping.json'. Proceeding with empty entity data.
Entity map file not found at /.../entities/sys_duration/mapping.json
```

### DIFF
--- a/mindmeld/components/entity_resolver.py
+++ b/mindmeld/components/entity_resolver.py
@@ -80,7 +80,7 @@ class EntityResolver:
             canonical_entities = []
         else:
             canonical_entities = self._resource_loader.get_entity_map(self.type).get("entities", [])
-        self._no_canonical_entities = len(canonical_entities) == 0
+        self._no_canonical_entity_map = len(canonical_entities) == 0
 
     @property
     def _es_index_name(self):
@@ -200,7 +200,7 @@ class EntityResolver:
             clean (bool): If ``True``, deletes and recreates the index from scratch instead of
                           updating the existing index with synonyms in the mapping.json.
         """
-        if self._is_system_entity or self._no_canonical_entities:
+        if self._no_canonical_entity_map:
             return
 
         if not self._use_text_rel:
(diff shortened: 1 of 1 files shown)


vijay120:
Looks like a flaky test that is failing.

mindmeld/components/entity_resolver.py
```diff
@@ -76,10 +76,11 @@ def __init__(
         self._es_host = es_host
         self._es_config = {"client": es_client, "pid": os.getpid()}
 
-        num_canonical_entities = len(
-            self._resource_loader.get_entity_map(self.type).get("entities", [])
-        )
-        self._no_canonical_entities = num_canonical_entities == 0
+        if self._is_system_entity:
```
minhtuev, line 79: We should document here why system entities do not have canonical entities?

mindmeld/components/entity_resolver.py
```diff
@@ -76,10 +76,11 @@ def __init__(
...
+            canonical_entities = []
+        else:
+            canonical_entities = self._resource_loader.get_entity_map(self.type).get("entities", [])
+        self._no_canonical_entities = len(canonical_entities) == 0
```
minhtuev, line 83: does this property apply to system entities? @serapio  @arushir

vijay120: System entities dont have canonical maps since they are resolved externally (through Duckling), hence `_no_canonical_entities ` would be `0` for them. The logic seems correct.

minhtuev: should this property be None instead of 0 since canonical entities are not applicable to them?

vijay120: The variable is named `_no_canonical_entities`, which is basically a True or False question, so I think its valid for it to be False for system entities since they do not have canonical entities.

arushir: Yes, there is no entity map for system entities since the resolution is done via Duckling. Maybe we could update this variable to `no_canonical_entity_map` to make this more clear. The logic is good.

Issue #194, opened 2020-07-20
Pass all system entity value candidates through 


serapio:
Currently the system entity logic not only select a single entity type for a given span but also selects a single entity value. For example, the raw output from Duckling for "tomorrow at 4" includes both `value.value = "2020-07-21T04:00:00.000-07:00" and `value.values` a list of 2 candidate values capturing the 4 AM and 4 PM meanings.
```
{
        "body": "tomorrow at 4",
        "start": 1,
        "value": {
            "values": [
                {
                    "value": "2020-07-21T04:00:00.000-07:00",
                    "grain": "hour",
                    "type": "value"
                },
                {
                    "value": "2020-07-21T16:00:00.000-07:00",
                    "grain": "hour",
                    "type": "value"
                }
            ],
            "value": "2020-07-21T04:00:00.000-07:00",
            "grain": "hour",
            "type": "value"
        },
        "end": 14,
        "dim": "time",
        "latent": false
    }
```
In cases of many potential meanings, e.g. "Tuesday", the candidates are limited to 3.
MindMeld exposes `value` to the application as a list of length 1, so it would compatible and more informative to expose the list `value.values`.

PR #197, opened 2020-07-28
Surface up values field from duckling

vijay120:
This PR addresses this issue: https://github.com/cisco/mindmeld/issues/194

It ensures the values field in duckling is ingested into the Entity object in MindMeld


mindmeld/system_entity_recognizer.py
```diff
@@ -605,6 +599,16 @@ def get_candidates_for_text(
             return []
 
 
+def _construct_interval_helper(interval_item):
+    from_ = None
+    to_ = None
+    if "from" in interval_item:
+        from_ = interval_item["from"]["value"]
+    if "to" in interval_item:
```
minhtuev, line 607: elif?

mindmeld/system_entity_recognizer.py
```diff
@@ -628,24 +632,24 @@ def duckling_item_to_entity(item):
...
-            if "from" in item["value"]:
-                from_ = item["value"]["from"]["value"]
-            if "to" in item["value"]:
-                to_ = item["value"]["to"]["value"]
-
             # Some intervals will only contain one value. The other value will be None in that case
-            value["value"] = (from_, to_)
+            value["value"] = _construct_interval_helper(item["value"])
+            if "values" in item["value"]:
+                value["values"] = []
```
minhtuev, line 650: How about this?
```suggestion
                value["values"] = [_construct_interval_helper(interval_item) for interval_item in item["value"]["values"]]
```

tests/test_markup.py
```diff
@@ -177,6 +177,16 @@ def test_load_system(
         assert value in entity.entity.value["value"]
 
 
+def test_entity_values(query_factory):
+    processed_query = markup.load_query(
+        "lets meet {2 hours from now|sys_interval}", query_factory, query_options={"timestamp": 1592002800}
```
minhtuev, line 182: we should add more tests to cover the cases: "from ... to ...", "...to..."

serapio: This is a confusing test. "2 hours from now" is a sys_time, while "from now" is a sys_interval. I think better would be tests that explicitly test each one-sided interval, plus the bounded interval:
 - "let's meet after 3:00"
 - "let's meet until 3:00"
 - "let's meet from now until 3:00"

mindmeld/system_entity_recognizer.py
vijay120: done

mindmeld/system_entity_recognizer.py
```diff
@@ -605,6 +599,16 @@ def get_candidates_for_text(
             return []
 
 
+def _construct_interval_helper(interval_item):
+    from_ = None
```
minhtuev, line 603:
```suggestion
    from_ = interval_item.get("from", {}).get("value", None")
    to_ = interval_item.get("to", {}).get("value", None")
```

vijay120: done

serapio:
Currently the code produces output like this:
```
>>> pprint(nlp.process('is the store open Sunday'))
{'domain': 'store_info',
 'entities': [{'role': None,
               'span': {'end': 23, 'start': 18},
               'text': 'Sunday',
               'type': 'sys_time',
               'value': [{'additional_candidate_values': [{'grain': 'day',
                                                           'type': 'value',
                                                           'value': '2020-08-02T00:00:00.000-07:00'},
                                                          {'grain': 'day',
                                                           'type': 'value',
                                                           'value': '2020-08-09T00:00:00.000-07:00'},
                                                          {'grain': 'day',
                                                           'type': 'value',
                                                           'value': '2020-08-16T00:00:00.000-07:00'}],
                          'grain': 'day',
                          'value': '2020-08-02T00:00:00.000-07:00'}]}],
 'intent': 'get_store_hours',
 'text': 'is the store open Sunday'}
```
1. Why not directly use the "additional candidate values" as the top-level value list? It's already a list, and in other entity types that use the ES entity resolution, it is a comparable nbest list of candidates.
2. If we do keep this structure, I think we should name the candidate list comparable to our other candidate lists, e.g. using "nbest" or "alternate" instead of "additional". "Additional" implies that it doesn't include the top candidate.

vijay120:
>> Why not directly use the "additional candidate values" as the top-level value list? It's already a list, and in other entity types that use the ES entity resolution, it is a comparable nbest list of candidates.

Great point. The current schema follows what duckling provides, changing it have a consistent list structure would break backwards compatibility. For this reason, it makes sense to leave the current schema in place.

>> If we do keep this structure, I think we should name the candidate list comparable to our other candidate lists, e.g. using "nbest" or "alternate" instead of "additional". "Additional" implies that it doesn't include the top candidate.

Makes sense, will rename.

vijay120:
> We should also test the output of `process_query()`. In that flow we have a couple places where it looks like we're only passing up a single value, but maybe it is passing both value.value and value.values e.g.
> entity_resolver L337
> 
> ```
>         if self._is_system_entity:
>             # system entities are already resolved
>             return [top_entity.value]
> ```

`top_entity.value` refers to `Entity` class's `value` field, which contains the `alternate_values ` key in it, hence it surfaces this information.


> 
> nlp.py L1486
> 
> ```
>         entity.entity.value = self.entity_resolver.predict(entity_list)
> ```

The top level `entity` here is a `QueryEntity`, which contains a field `entity` which is an `Entity` class. This `Entity` class contains `value`, which is a dict containing `alternate_values`.

tests/test_markup.py
```diff
@@ -177,6 +177,28 @@ def test_load_system(
...
 
 
+@pytest.mark.load
+@pytest.mark.system
+@pytest.mark.parametrize(
+    "query, expected_values",
+    [("let's meet {after 3:00|sys_interval}", [('2020-06-12T15:00:00.000-07:00', None),
+                                               ('2020-06-13T03:00:00.000-07:00', None),
+                                               ('2020-06-13T15:00:00.000-07:00', None)]),
+     ("let's meet {until 3:00|sys_interval}",
```
serapio, line 187: "let's meet until 3" to avoid the weird bounded interval result

vijay120: done

PR #199, opened 2020-08-04
Feature/sys annotate

KKsharma99:
PR for the Automatic Annotator. This merge includes:

- Abstract Annotator class: Can be extended to create custom Annotators. (auto_annotator.py)
- SpacyAnnotator class: Can parse individual sentences or modify training/test files. (auto_annotator.py)
- Command line support: "annotate" and "unannotate" functions. (cli.py)
- Functions for config retrieval: "get_auto_annotator_config". (components/_config.py)
- Functions to get and filter all app file paths. (resource_loader.py)
- Tests for the Auto Annotator. (tests/test_auto_annotator.py)
- Misc. Black formatting changes.

Documentation:
https://github.com/KKsharma99/mindmeld/blob/feature/sys_annotate/source/userguide/auto_annotator.rst


mindmeld/resource_loader.py
```diff
@@ -417,6 +417,35 @@ def _traverse_labeled_queries_files(
...
+        Args:
+            file_pattern (str): A regex pattern to filter on.
+            file_paths (list): A list of file paths.
+
+        Returns:
+            list: A list of file paths.
+        """
+        all_file_paths = file_paths or self.get_all_file_paths()
+        matched_paths = [
+            path for path in all_file_paths if re.match(file_pattern, path)
```
minhtuev, line 441: we should try to compile the file pattern to make it faster for matching

```
In [5]: pattern = "(.)*\.txt"

In [6]: compiled_pattern = re.compile(pattern)

In [7]: compiled_pattern.match("training.txt")
Out[7]: <re.Match object; span=(0, 12), match='training.txt'>

In [8]: compiled_pattern.match("trainin")

File: mindmeld/components/_config.py [hunk 2]
```diff
@@ -899,3 +905,38 @@ def get_nlp_config(app_path=None, config=None):
...
+            _get_config_module(app_path),
+            "AUTO_ANNOTATOR_CONFIG",
+            DEFAULT_AUTO_ANNOTATOR_CONFIG,
+        )
+        return auto_annotator_config
+    except (OSError, IOError):
+        logger.info(
+            "No app configuration file found. Using the default automatic annotator config."
+        )
+        return _get_default_auto_annotator_config()
```
minhtuev, line 942: Why do we need to have a separate function to return the deepcopy? Can we just do the same thing in line 935?

mindmeld/cli.py
```diff
@@ -531,6 +532,33 @@ def _get_duckling_pid():
...
+
+
+@shared_cli.command("unannotate", context_settings=CONTEXT_SETTINGS)
+@click.option(
+    "--app-path",
+    required=True,
+    help="Needed to access data to unannotate and app config.",
+)
+def unannotate(app_path):
+    """Runs the automatic system entity annotator."""
```
minhtuev, line 556: update doc

mindmeld/cli.py
```diff
@@ -531,6 +532,33 @@ def _get_duckling_pid():
...
+    spacy_annotator = SpacyAnnotator(app_path=app_path, model=model)
+    spacy_annotator.annotate()
+    logger.info("Annotation Complete.")
+
+
+@shared_cli.command("unannotate", context_settings=CONTEXT_SETTINGS)
+@click.option(
+    "--app-path",
+    required=True,
+    help="Needed to access data to unannotate and app config.",
```
minhtuev, line 553: `to unannotate and app config` is not clear about what you do with app config

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,661 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from abc import ABC, abstractmethod
+import re
+import logging
+import os
+import importlib
+from tqdm import tqdm
+import spacy
```
minhtuev, line 19: we should add spacy to the requirements

mindmeld/cli.py
```diff
@@ -531,6 +532,33 @@ def _get_duckling_pid():
     return pid
 
 
+@shared_cli.command("annotate", context_settings=CONTEXT_SETTINGS)
+@click.option(
+    "--app-path",
+    required=True,
+    help="Needed to access data to annotate and app config.",
```
minhtuev, line 539: `to annotate and app config` is unclear

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,661 @@
...
+            )
+            tqdm_desc = "Processing " + path + ": "
+            for processed_query in tqdm(processed_queries, ascii=True, desc=tqdm_desc):
+                entity_types = file_entities_map[path]
+                if action == "annotate":
+                    self._annotate_query(
+                        processed_query=processed_query, entity_types=entity_types
+                    )
+                elif action == "unannotate":
+                    Annotator._unannotate_query(
```
vijay120, line 153: Why do we do `Annotator._unannotate_query` for the unannotate usecase but `self._annotate_query` for the annotate case?

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,661 @@
...
+        """
+        query_factory = QueryFactory.create_query_factory(self.app_path)
+        for path in file_entities_map:
+            processed_queries = Annotator._get_processed_queries(
+                file_path=path, query_factory=query_factory
+            )
+            tqdm_desc = "Processing " + path + ": "
+            for processed_query in tqdm(processed_queries, ascii=True, desc=tqdm_desc):
+                entity_types = file_entities_map[path]
+                if action == "annotate":
```
vijay120, line 148: We should make the two constants into enums

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,661 @@
...
+    """ Annotator class that uses spacy to generate annotations.
+    """
+
+    def __init__(self, app_path, config=None, model="en_core_web_lg"):
+        super().__init__(app_path=app_path, config=config)
+
+        self.nlp = SpacyAnnotator._load_model(model)
+        self.model = model
+        self.duckling = DucklingRecognizer.get_instance()
+        self.SYS_MAPPINGS = {
```
vijay120, line 317: Maybe `SYS_MAPPINGS ` could be renamed as `ANNOTATOR_TO_DUCKLING_ENTITY_MAPPINGS `

mindmeld/auto_annotator.py
KKsharma99: So far I have added "spacy>=2.3.0" to the requirements array in setup.py (Line 37). Which other files would I need to update as well?

mindmeld/auto_annotator.py
KKsharma99: It was to avoid pylint's "no-self-use" warning since the _unannotate_query() function does not use class attributes. Should I switch it back to self._unannotate_query() and add a comment that ignores the pylint error?

# pylint: disable=R0201

mindmeld/auto_annotator.py
minhtuev: Ah, my bad 👍

mindmeld/auto_annotator.py
KKsharma99: Would the below example work?

class AnnotatorAction(Enum):
    ANNOTATE = "annotate"
    UNANNOTATE = "unannotate"

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,708 @@
...
+    def __init__(self, app_path, config=None, resource_loader=None):
+        """ Initializes an annotator."""
+        self.app_path = app_path
+        self.config = get_auto_annotator_config(app_path=app_path, config=config)
+        self._resource_loader = (
+            resource_loader or ResourceLoader.create_resource_loader(app_path)
+        )
+        self.annotate_file_entities_map = self._get_file_entities_map(action="annotate")
+
+    def _get_file_entities_map(self, action="annotate"):
```
vijay120, line 54: I think this comment by @tmehlinger is quite prescient here: https://sqbu-github.cisco.com/WebexIntelligence/joint_bert_model/pull/1

Let's state types for every function using mypy in the function arguments so that we dont have to do these types of checks:
```
        if action == AnnotatorAction.ANNOTATE.value:
            rules = self.config[AnnotatorAction.ANNOTATE.value]
        elif action == AnnotatorAction.UNANNOTATE.value:
            rules = self.config[AnnotatorAction.UNANNOTATE.value]
```

Instead, we can do this:
```
        if action == AnnotatorAction.ANNOTATE:
            rules = self.config[AnnotatorAction.ANNOTATE]
        elif action == AnnotatorAction.UNANNOTATE:
            rules = self.config[AnnotatorAction.UNANNOTATE]
```

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,708 @@
...
+            filtered_paths = self._resource_loader.filter_file_paths(
+                file_pattern=pattern, file_paths=all_file_paths
+            )
+            for path in filtered_paths:
+                entities = self._get_entities(rule)
+                file_entities_map[path] = entities
+        return file_entities_map
+
+    @staticmethod
+    def _get_pattern(rule):
```
vijay120, line 85: Can we add a test for this function?

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,708 @@
...
+    def valid_entity_check(self, entity):
+        """ Determine if an entity type is valid.
+
+        Args:
+            entity (str): Name of entity to annotate.
+
+        Returns:
+            bool: Whether entity is valid.
+        """
+        return True
```
vijay120, line 152: Is this function needed?

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,708 @@
...
+        """
+        query_factory = QueryFactory.create_query_factory(self.app_path)
+        for path in file_entities_map:
+            processed_queries = Annotator._get_processed_queries(
+                file_path=path, query_factory=query_factory
+            )
+            tqdm_desc = "Processing " + path + ": "
+            for processed_query in tqdm(processed_queries, ascii=True, desc=tqdm_desc):
+                entity_types = file_entities_map[path]
+                if action == AnnotatorAction.ANNOTATE.value:
```
vijay120, line 184: Lets check against the Enum and not it's value here as well.

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,708 @@
...
+                    self._annotate_query(
+                        processed_query=processed_query, entity_types=entity_types
+                    )
+                elif action == AnnotatorAction.UNANNOTATE.value:
+                    self._unannotate_query(
+                        processed_query=processed_query, remove_entities=entity_types
+                    )
+
+            annotated_queries = list(dump_queries(processed_queries))
```
vijay120, line 193: This is a memory intensive op. Can we use `dump_query` instead as we sequential write the process query to disk? This should happen within the for loop itself.

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,708 @@
...
+        for o_entity in other_entities:
+            no_overlaps = [
+                Annotator._no_overlap(o_entity, b_entity) for b_entity in base_entities
+            ]
+            if all(no_overlaps):
+                additional_entities.append(o_entity)
+        return base_entities + additional_entities
+
+    @staticmethod
+    def _no_overlap(entity_one, entity_two):
```
vijay120, line 304: Lets add a test for this function

minhtuev (commented):
Do we need to include the glove data?

mindmeld/cli.py
```diff
@@ -531,6 +535,32 @@ def _get_duckling_pid():
     return pid
 
 
+@shared_cli.command("annotate", context_settings=CONTEXT_SETTINGS)
+@click.option(
+    "--app-path", required=True, help="Needed to locate the application config file.",
+)
+def annotate(app_path):
+    """Runs the annotation command of the Auto Annotator."""
+    importlib.import_module("mindmeld.auto_annotator")
```
minhtuev, line 544: why is this for?

mindmeld/cli.py
```diff
@@ -531,6 +535,32 @@ def _get_duckling_pid():
...
+    logger.info("Annotation Complete.")
+
+
+@shared_cli.command("unannotate", context_settings=CONTEXT_SETTINGS)
+@click.option(
+    "--app-path", required=True, help="Needed to locate the application config file.",
+)
+def unannotate(app_path):
+    """Runs the unannotation command of the Auto Annotator."""
+    importlib.import_module("mindmeld.auto_annotator")
```
minhtuev, line 557: same

mindmeld/resource_loader.py
```diff
@@ -417,6 +417,38 @@ def _traverse_labeled_queries_files(
                     if re.match(file_pattern, os.path.basename(file_path)):
                         yield a_domain, an_intent, file_path
 
+    def get_all_file_paths(self):
```
minhtuev, line 420: How about:
```suggestion
    def get_all_file_paths(self, file_pattern=".*.txt"):
```

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,708 @@
...
+
+        Args:
+            model (str): Spacy model ("en_core_web_sm", "en_core_web_md", or
+                "en_core_web_lg").
+
+        Returns:
+            nlp (spacy.lang.en.English): Spacy language model.
+        """
+
+        if model in ["en_core_web_sm", "en_core_web_md", "en_core_web_lg"]:
```
vijay120, line 381: Make these into a list of constants defined on the top of this file

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,708 @@
...
+        else:
+            error_msg = (
+                "Unknown Spacy model name: {!r}. Model must be 'en_core_web_sm',"
+                " 'en_core_web_md', or 'en_core_web_lg'".format(model)
+            )
+            raise ValueError(error_msg)
+
+    def valid_entity_check(self, entity):
+        entity = entity.lower().strip()
+        valid_entities = [
```
vijay120, line 398: Lets add to the constants.py file.

mindmeld/resource_loader.py
```diff
@@ -417,6 +417,38 @@ def _traverse_labeled_queries_files(
...
+        """ Get a list of text file paths across all intents.
+
+        Returns:
+            list: A list of all file paths.
+        """
+        file_iter = self._traverse_labeled_queries_files(file_pattern=".*.txt")
+        return [filename for _, _, filename in file_iter]
+
+    def filter_file_paths(self, file_pattern=".*", file_paths=None):
```
minhtuev, line 429: `get_all_file_paths` returns all text files, so why are we matching for all patterns here?

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,708 @@
...
+
+        candidates = self.duckling.get_candidates_for_text(entity["body"])
+        if len(candidates) == 0:
+            return
+
+        possible_values = []
+        for candidate in candidates:
+            if candidate["entity_type"] == "sys_number":
+                possible_values.append(candidate["value"]["value"])
+        entity["value"]["value"] = max(possible_values) / 100
```
vijay120, line 686: Does this have to be a float or int?

mindmeld/resource_loader.py
```diff
@@ -417,6 +417,38 @@ def _traverse_labeled_queries_files(
...
+
+        Args:
+            file_pattern (str): A regex pattern to filter on.
+            file_paths (list): A list of file paths.
+
+        Returns:
+            list: A list of file paths.
+        """
+        all_file_paths = file_paths or self.get_all_file_paths()
+        compiled_pattern = re.compile(file_pattern)
```
vijay120, line 440: Do the compilation on the caller of this function

mindmeld/auto_annotator.py
vijay120: Let's not annotate this whole PR with types. But let's use it for functions like these where enums are expected.

vijay120 (commented):
A few comments

mindmeld/auto_annotator.py
KKsharma99: Sure, will do

mindmeld/auto_annotator.py
KKsharma99: The abstract method valid_entity_check() returns True by default but developers are encouraged to overwrite this function in their implementation class to only support a custom list of entities.

Should I raise a NotImplementedException instead?

mindmeld/cli.py
KKsharma99: The auto_annotator needs to be loaded dynamically here because at bottom of the auto_annotator.py file the SpacyAnnotator is registered. This register function comes from models/helpers.py.

auto_annotator.py (ln 708): register_annotator("SpacyAnnotator", SpacyAnnotator)

Importing the module at the top of the file creates a pylint error since the module isn't directly used but just loaded so that ANNOTATOR_MAP = {} in models/helpers.py can be updated

mindmeld/auto_annotator.py
KKsharma99: It would be a float. For example, if the query was "thirty five percent" and the highest found value was 35. It will take that value and divide by 100 to give 0.35 as the identified percent-type entity value

mindmeld/resource_loader.py
KKsharma99: Updated to the suggested change

mindmeld/auto_annotator.py
KKsharma99: Added

mindmeld/auto_annotator.py
KKsharma99: Added tests

mindmeld/resource_loader.py
KKsharma99: Good point, I removed the default value for file_pattern

mindmeld/auto_annotator.py
KKsharma99: done

mindmeld/auto_annotator.py
KKsharma99: To clarify, we want to add the _no_overlap() function to constants.py? Double checking since constants.py doesn't have functions yet

KKsharma99:
> Do we need to include the glove data?

Removed glove data. Maybe we can add "mindmeld/data/*" to the .gitignore since the data folder is a generated folder?

mindmeld/resource_loader.py
KKsharma99: To clarify, are you saying to replace the function parameter 'file_pattern' with 'compiled_pattern'? This way the caller is responsible for running re.compile() and passing the resulting value into this func.

mindmeld/auto_annotator.py
KKsharma99: Could you please elaborate on how we can use mypy to be able to do the second type of checks?

Currently, I am getting an error when comparing a string with an enum value as in the case of the second set of checks above

mindmeld/cli.py
minhtuev: You can just do a pylint ignore for the module import at the top; we would like to avoid dynamic import as much as possible

tests/test_auto_annotator.py
```diff
@@ -0,0 +1,220 @@
...
+test_auto_annotator
+----------------------------------
+
+Tests for `auto_annotator` module
+"""
+import pytest
+
+from mindmeld.auto_annotator import SpacyAnnotator
+
+SA = SpacyAnnotator("")
```
minhtuev, line 14: you should configure this as a fixture object

is that okay not to pass in app path? you can pass in app path for either kwik e mart or home assistant

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,691 @@
...
+        Returns:
+            nlp (spacy.lang.en.English): Spacy language model.
+        """
+
+        if model in [EN_CORE_WEB_SM, EN_CORE_WEB_MD, EN_CORE_WEB_LG]:
+            logger.info("Loading Spacy model %s.", model)
+            try:
+                return spacy.load(model)
+            except OSError:
+                os.system("python -m spacy download " + model)
```
minhtuev, line 391: we should log a warning here

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,691 @@
...
+            entities (list): List of entity dictionaries.
+        """
+        doc = self.nlp(sentence)
+        spacy_entities = [
+            {
+                "body": ent.text,
+                "start": ent.start_char,
+                "end": ent.end_char,
+                "value": {"value": ent.text},
+                "dim": ent.label_.lower(),
```
minhtuev, line 427: should we append the entity type with prefix "spacy-" or "sp-" to distinguish this with duckling entity type? alternatively, we can also map spacy entity type to duckling entity type

minhtuev:
> > Do we need to include the glove data?
> 
> Removed glove data. Maybe we can add "mindmeld/data/*" to the .gitignore since the data folder is a generated folder?

Yes, please go ahead to add to the `.gitignore` folder

mindmeld/cli.py
KKsharma99: Got it, done

mindmeld/cli.py
KKsharma99: Updated to import at the top of the file

tests/test_auto_annotator.py
KKsharma99: Updated, used "kwik_e_mart"

KKsharma99:
> > > Do we need to include the glove data?
> > 
> > 
> > Removed glove data. Maybe we can add "mindmeld/data/*" to the .gitignore since the data folder is a generated folder?
> 
> Yes, please go ahead to add to the `.gitignore` folder

Added to `.gitignore`

mindmeld/auto_annotator.py
KKsharma99: Added warning

`logger.warning("%s not found. Downloading the model.", model)`

mindmeld/auto_annotator.py
KKsharma99: I think we should avoid adding a "spacy-" prefix to the entity types since a developer would want Mindmeld to detect {3pm|sys_time} for example as a sys_entity. If a prefix was used, then in this case MindMeld wouldn't recognize {3pm|spacy-sys_time} as this would be now be a custom entity.

When making these annotations, the SpacyAnnotator leverages Duckling and if the type is shared between the two, the SpacyAnnotator would only apply a label if the string (or substring) can be resolved by Duckling.

mindmeld/auto_annotator.py
KKsharma99: Updated to write one query at a time

mindmeld/auto_annotator.py
KKsharma99: Updated to use mypy. The function now takes AnnotatorAction as the type for the action parameter.

Needed to use the value in order to index into the config dict with a str

`rules = self.config[AnnotatorAction.ANNOTATE.value]`

mindmeld/auto_annotator.py
vijay120: Lets use `subprocess ` instead of `os.sysem` since thats what we use for other system calls in MindMeld. It also is non-blocking

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,694 @@
...
+            }
+            for ent in doc.ents
+        ]
+
+        entities = []
+        for entity in spacy_entities:
```
vijay120, line 436: Instead of having this huge `elif` block, why cant we have a map of entity type and function. There might be some duplicates, but I think its better for readability.

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,694 @@
...
+        candidates = self.duckling.get_candidates_for_text(entity["body"])
+
+        if len(candidates) == 0:
+            return
+
+        time_entities = ["sys_duration", "sys_interval", "sys_time"]
+        if entity_types:
+            time_entities = [e for e in time_entities if e in entity_types]
+
+        if SpacyAnnotator._resolve_time_exact_match(entity, candidates, time_entities):
```
vijay120, line 484: Add an explanation on why the conditionals are done in this order.

vijay120 (approved):
Please address the small nits. Looks good overall. Nice work!

mindmeld/auto_annotator.py
KKsharma99: Updated

mindmeld/auto_annotator.py
KKsharma99: Added

mindmeld/cli.py
```diff
@@ -531,6 +534,30 @@ def _get_duckling_pid():
     return pid
 
 
+@shared_cli.command("annotate", context_settings=CONTEXT_SETTINGS)
+@click.option(
+    "--app-path", required=True, help="Needed to locate the application config file.",
```
minhtuev, line 539:
```suggestion
    "--app-path", required=True, help="The application's path",
```

mindmeld/constants.py
```diff
@@ -15,3 +15,43 @@
...
+    "sys_product",
+    "sys_event",
+    "sys_law",
+    "sys_langauge",
+    "sys_work-of-art",
+    "sys_other-quantity",
+]
+
+
+def _no_overlap(entity_one, entity_two):
```
minhtuev, line 45: should this be a function of the entity itself `entity_one.has_overlap(entity_two)`?

KKsharma99: If there are future applications that would encounter overlapping entities, it makes sense to add this as a function of the class. (If so, which entity Class would this function be added to?) Otherwise, it may be okay to keep it as a separate method

mindmeld/cli.py
```diff
@@ -540,6 +543,30 @@ def _get_duckling_pid():
...
+    """Runs the annotation command of the Auto Annotator."""
+    config = get_auto_annotator_config(app_path=app_path)
+    annotator = create_annotator(app_path=app_path, config=config)
+    annotator.annotate()
+    logger.info("Annotation Complete.")
+
+
+@shared_cli.command("unannotate", context_settings=CONTEXT_SETTINGS)
+@click.option(
+    "--app-path", required=True, help="Needed to locate the application config file.",
```
ritvikshrivastava, line 560: Shouldn't the help message here be the same as 'annotate'?

mindmeld/cli.py
```diff
@@ -540,6 +543,30 @@ def _get_duckling_pid():
...
+)
+def unannotate(app_path):
+    """Runs the unannotation command of the Auto Annotator."""
+    config = get_auto_annotator_config(app_path=app_path)
+    annotator = create_annotator(app_path=app_path, config=config)
+    annotator.unannotate()
+    logger.info("Unannotation Complete.")
```
ritvikshrivastava, line 567: I don't think 'unannotation' is a word, maybe we can say 'Annotation Removal Compete'?

ritvikshrivastava (commented):
Some comments, but overall LGTM!

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,708 @@
...
+        other_entities = current_entities if overwrite else annotated_entities
+
+        additional_entities = []
+        for o_entity in other_entities:
+            no_overlaps = [
+                _no_overlap(o_entity, b_entity) for b_entity in base_entities
+            ]
+            if all(no_overlaps):
+                additional_entities.append(o_entity)
+        return base_entities + additional_entities
```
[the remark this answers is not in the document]
minhtuev: I think the extend function does not return anything (None)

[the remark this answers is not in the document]
ritvikshrivastava:
```suggestion
        base_entities.extend(additional_entities)
        return base_entities
```

You're right, can do this instead. 
'+' creates a new third value in memory for the concatenated list while extend modifies the primary list.

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,474 @@
...
+
+		“body”: (substring of sentence), 
+
+		“start”: (start index), 
+
+		“end”: (end index + 1), 
+
+		“dim”: (entity type), 
+
+		“value”: (resolved value, if it exists),
```
serapio, line 132:
```suggestion
		"value": (resolved value, if it exists), 
```

serapio (changes requested):
Will come back for more later today.

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,474 @@
...
+| "sys_event"            | No                      | Named hurricanes, battles, wars, sports events, etc.                        |
++------------------------+-------------------------+-----------------------------------------------------------------------------+
+| "sys_law"              | No                      | Named documents made into laws.                                             |
++------------------------+-------------------------+-----------------------------------------------------------------------------+
+| "sys_langauge"         | No                      | Any named language.                                                         |
++------------------------+-------------------------+-----------------------------------------------------------------------------+
+| "sys_work-of-art"      | No                      | Titles of books, songs, etc.                                                |
++------------------------+-------------------------+-----------------------------------------------------------------------------+
+| "sys_other-quantity"   | No                      | “40 degrees”                                                                |
```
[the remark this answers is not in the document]
KKsharma99: I have replaced "40 degrees" with "10 joules", "30 liters", "15 tons". These new examples are not resolved by duckling.

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,474 @@
...
++------------------------+-------------------------+-----------------------------------------------------------------------------+
+
+
+To detect entities in a single sentence first create an instance of the :class:`SpacyAnnotator` class.
+
+.. code-block:: python
+
+	from mindmeld.auto_annotator import SpacyAnnotator 
+	sa = SpacyAnnotator(app_path=”hr_assistant”, model=”en_core_web_lg”)
```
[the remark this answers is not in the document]
KKsharma99: Yes, Annotator specific configurations are added through the config. Updated the docs to reflect this. Good catch!

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,474 @@
...
+				"entities": "sys_person", 
+			}
+		],
+		"unannotate_supported_entities_only": True, 
+		"unannotate": None
+	}
+
+Before running the annotation, we can see the first four queries in the train.txt file for the :attr:`get_hierarchy_up` intent: 
+
+.. code-block:: python
```
serapio, line 344: We don't need any syntax highlighting for query data. 
```suggestion
.. code-block:: none
```

serapio (changes requested):

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,474 @@
...
+
+				"entities": "*", 
+
+			} 
+
+		], 
+
+		"unannotate_supported_entities_only": True, 
+
+		"unannotate": None,
```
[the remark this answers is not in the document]
KKsharma99: Updated to consistently be single space

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,474 @@
...
+
+		"entities": "sys_amount-of-money|sys_time", 
+
+	}
+
+The rule above would annotate all text files named "train" or "test" in the "faq" and "salary" domains. Only sys_amount-of-money and sys_time entities would be annotated.
+
+.. note::
+
+	The order of the annotation rules matters. A rule will overwrite a previous rule if there are conflicts.
```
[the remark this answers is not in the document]
KKsharma99: Yes, since the overwrite in the default config refers to whether existing annotations should be overwritten

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,474 @@
...
+
+``'spacy_model'`` (:class:`str`): :attr:`en_core_web_lg` is used by default for the best performance. Alternative options are :attr:`en_core_web_sm` and :attr:`en_core_web_md`. This parameter is optional and is specific to the use of the :class:`SpacyAnnotator`.
+If the selected english model is not in the current environment it will automatically be downloaded. Refer to Spacy's documentation to learn more about their `english models <https://spacy.io/models/en>`_.
+
+
+Using the Auto Annotator
+------------------------
+
+The Auto Annotator can be used by importing a class that implements the :class:`Annotator` abstract class in the :mod:`auto_annotator` module or through the command-line.
+We will demonstrate both approaches for Unannotation and Annotation using the :class:`SpacyAnnotator` class.
```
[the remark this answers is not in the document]
KKsharma99: That's correct, a more natural workflow is to annotate and then unannotate. While that makes more sense in practice, since blueprints are already annotated it's easier to demonstrate in this order. I have added this sentence to help clarify what the natural workflow is: 

"For the this tutorial we will first describe unannotation and then annotation, however, in a normal workflow you are likely to annotate first and then unannotate any annotations you are not pleased with."

source/userguide/auto_annotator.rst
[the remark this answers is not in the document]
serapio: Is it possible to reverse this behavior? In a system of ordered rules like this, you could have phantom blocking effects, e.g.
1. Rule 1 adds an entity: `{A B} C D`
2. Rule 2 replaces that entity with another `A {B C} D`
3. Rule 3 replaces that entity with another `A B {C D}`

The result is that Rule 1 fails to apply just because it conflicts with a rule that also didn't apply.

tests/test_auto_annotator.py
```diff
@@ -0,0 +1,223 @@
...
+            },
+            ".*/date/get_date/(train.txt|test.txt)",
+        ),
+        (
+            {"domains": "general", "intents": "*", "files": "*", "entities": "*"},
+            ".*/general/.+/.+",
+        ),
+        (
+            {"domains": "*", "intents": "*", "files": "*", "entities": "*"},
+            ".*/.+/.+/.+",
```
serapio, line 218: Add a case that uses the kleene star on something other than '.', e.g. "\d*" or "[A-Z]*". I think the current parser messes up.

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,709 @@
...
+
+        Args:
+            segment (str): Section of a rule dictionary ("domains", "intents", "entities").
+
+        Returns:
+            segment (str): Cleaned section of the rule dictionary.
+        """
+        segment = re.sub("[()]", "", segment)
+        segment = segment.replace(".*", ".+")
+        segment = segment.replace("*", ".+")
```
serapio, line 122: Looks like this will mess up if '*' is used for other purposes.

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,709 @@
...
+        """ Process an individual segment from a rule dictionary.
+
+        Args:
+            segment (str): Section of a rule dictionary ("domains", "intents", "entities").
+
+        Returns:
+            segment (str): Cleaned section of the rule dictionary.
+        """
+        segment = re.sub("[()]", "", segment)
+        segment = segment.replace(".*", ".+")
```
serapio, line 121: This might surprise developers too, e.g. "train.*_crowd.txt" where it is supposed to match "train_crowd.txt" and "train-foo_crowd.txt"

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,709 @@
...
+
+        Args:
+            model (str): Spacy model ("en_core_web_sm", "en_core_web_md", or
+                "en_core_web_lg").
+
+        Returns:
+            nlp (spacy.lang.en.English): Spacy language model.
+        """
+
+        if model in [EN_CORE_WEB_SM, EN_CORE_WEB_MD, EN_CORE_WEB_LG]:
```
serapio, line 387: Is the interface different for other languages? How hard would it be to let this same class support other languages?

serapio (changes requested):

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,709 @@
...
+            segment (str): Section of a rule dictionary ("domains", "intents", "entities").
+
+        Returns:
+            segment (str): Cleaned section of the rule dictionary.
+        """
+        segment = re.sub("[()]", "", segment)
+        segment = segment.replace(".*", ".+")
+        segment = segment.replace("*", ".+")
+        segment = segment.split("|")
+        segment = "|".join([x.strip() for x in segment])
```
ritvikshrivastava, line 124: Possible to merge the two lines:
```suggestion
        segment = "|".join([x.strip() for x in segment.split("|")])
```

Are you trying to remove just the newline characters here?
You can simply do the following without the whole (split, strip, join) process:
```suggestion
        segment.replace('\n', '')
```

mindmeld/auto_annotator.py
KKsharma99: These are good points about the unexpected results from the current regex preprocessing. We can do away with the preprocessing the regex pattern entirely and only combine the annotation rule dictionary to a single string. (".*/domains/intents/files")

This would allow developers to use regex as expected. A developer would need to use ".*" instead of "*" alone when defining annotations rules.

I have tested this out locally and it works fine provided the annotation rules are in proper regex. If matches are not found for a given rule, a warning is logged. This way the developer is informed of problematic annotation rules.

Should we move towards this approach instead?

mindmeld/auto_annotator.py
KKsharma99: Merged the two lines. This function aims to remove spacing within a regex rule for example, "(  note | chair  )" should be "(note|chair)".

May be removing this function entirely since the current preprocessing of regex could lead to unexpected results when used

mindmeld/auto_annotator.py
KKsharma99: Ah I see, the alternate implementation proposal in the comment above could resolve this

mindmeld/auto_annotator.py
KKsharma99: The interface is the same for other languages. The typical workflow is:

nlp = spacy.load("es_core_news_lg")
doc = nlp("Pollito Tito está desayunando en la cocina.")
ents = [(e.text, e.start_char, e.end_char, e.label_) for e in doc.ents]

"ents" would have entities in the same format as those used by the English model. I believe the same NER tags are used as well.

source/userguide/auto_annotator.rst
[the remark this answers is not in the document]
KKsharma99: The motivation behind this design is to be able to annotate entities from general to specific rules.

For example, say you want to annotate entities in the "date" domain in the HR assistant with the general rule to annotate sys_time and sys_date. However, for the "get_date_range_aggregate" you would also like to include "sys_interval" then you would a second rule specifying this intent and the additional entity. In this manner, the previous rule isn't overwritten but expanded upon at a more specific level.

An alternative to avoid any conflict is to write rules for each intent, however, this may quickly become very verbose and time consuming.

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,708 @@
...
+            sentence (str): Sentence to detect entities.
+
+        Returns:
+            entities (list): List of entity dictionaries.
+        """
+        raise NotImplementedError("Subclasses must implement this method")
+
+
+class SpacyAnnotator(Annotator):
+    """ Annotator class that uses spacy to generate annotations.
```
serapio, line 355: note English here

tests/test_auto_annotator.py
KKsharma99: Done

mindmeld/auto_annotator.py
serapio: What scenarios is this supposed to handle? Does it give the expected result for something like "35.85%"?

serapio (changes requested):
just a couple minor things left

KKsharma99:
> just a couple minor things left



> just a couple minor things left

Good catch. Fixed the issue, "36.46%" now correctly resolves to 0.3646

In general, the SpacyAnnotator can resolve percent examples such as the following:

        ("four percent", 0.04),
        ("4 percent", 0.04),
        ("eight percent", 0.08),
        ("thirty eight percent", 0.38),
        ("12%", 0.12),
        ("seventy two percent", 0.72),
        ("thirty percent", 0.3)

serapio (approved):

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,691 @@
...
+            sentence (str): Sentence to detect entities.
+
+        Returns:
+            entities (list): List of entity dictionaries.
+        """
+        raise NotImplementedError("Subclasses must implement this method")
+
+
+class SpacyAnnotator(Annotator):
+    """ (English) Annotator class that uses spacy to generate annotations.
```
minhtuev, line 334: Let's add the type of the entities that the SpacyAnnotator supports.

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,456 @@
...
+		"start": (start index), 
+		"end": (end index + 1), 
+		"dim": (entity type), 
+		"value": (resolved value, if it exists), 
+	}
+
+Registering Annotator to Enable command-line Use
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In order to use our :class:`CustomAnnotator` when running annotation/unannotation commands from the command-line, we must add the following line to the bottom of the :attr:`auto_annotator.py` file to register the class and ensure that it loads in the :attr:`cli.py` file.
```
minhtuev, line 407: Ideally for a person to implement their own custom annotator, they should not add to MindMeld's source code. Thoughts?

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,691 @@
...
+        ):
+            return entity
+
+    def _resolve_money(self, entity, sentence):
+        currency_symbols = ["$", "€", "£", "¥", "₹", "₩", "฿"]
+        for symbol in currency_symbols:
+            if symbol in sentence:
+                start = entity["start"]
+                if (start == 1 and sentence[0] == symbol) or (
+                    start > 1 and sentence[start - 2 : start] == " " + symbol
```
vijay120, line 576: Have we done some validation checks here to make sure `sentence[start - 2 : start]` doesnt lead to an out of index error? Or is this case never going to happen?

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,691 @@
...
+        entity["dim"] = self.ANNOTATOR_TO_DUCKLING_ENTITY_MAPPINGS[entity["dim"]]
+
+        candidates = self.duckling.get_candidates_for_text(entity["body"])
+        if len(candidates) == 0:
+            return
+
+        possible_values = []
+        for candidate in candidates:
+            if candidate["entity_type"] == "sys_number":
+                if "." in candidate["body"]:
```
vijay120, line 665: I think this function needs to validate that the input `candidate["body"]` is a float. Checking `"." in candidate["body"]` would not cut it since it would pass strings like `alex.bob`. Maybe try this:
```
try:
    float(element)
except ValueError:
    logger.error()
```

mindmeld/auto_annotator.py
KKsharma99: An out of index error would not occur since when "start > 1" start would have minimum value of 2. At this minimum value, "[start - 2: start]" would be "[0:2]".

In the case that "start > 1" is not true, python does short-circuiting such that the second clause is not evaluated 

To improve readability, I can replace "start > 1" with "start >= 2"

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,691 @@
...
+        Returns:
+            nlp (spacy.lang.en.English): Spacy language model.
+        """
+
+        if model in [EN_CORE_WEB_SM, EN_CORE_WEB_MD, EN_CORE_WEB_LG]:
+            logger.info("Loading Spacy model %s.", model)
+            try:
+                return spacy.load(model)
+            except OSError:
+
```
vijay120, line 370: Nit: remove empty lines here

vijay120 (commented):
The feature is very well documented! A few more comments here, please address them and we would be close to shipping this.

source/userguide/auto_annotator.rst
KKsharma99: That's a good point. We could make the cli command exclusive to the mindmeld inbuilt SpacyAnnotator and potential future mindmeld Annotators.

Developers can run their annotators in the same file with a small addition to the boiler plate code presented in the docs.

if __name__ == "__main__":
    custom_annotator = CustomAnnotator(app_path="hr_assistant")
    custom_annotator.annotate()

Thoughts on this approach?

minhtuev: Yup I think that should work

mindmeld/auto_annotator.py
KKsharma99: Since this function filters the candidates to be of type "sys_number", we can expect to receive either an int or float as a string and not a string like "alex.bob". Validating a float with a try-except works well to distinguish between numerical and non-numerical values, however, it doesn't allow us to distinguish between int and float as both would pass the try block.

We can make a slight adjustment and validate an int instead since something like int("3.4") gives a value error

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,691 @@
...
+                in an App to a list of entities.
+        """
+        query_factory = QueryFactory.create_query_factory(self.app_path)
+        for path in file_entities_map:
+            processed_queries = Annotator._get_processed_queries(
+                file_path=path, query_factory=query_factory
+            )
+            tqdm_desc = "Processing " + path + ": "
+            temp_path = path.split(".txt")[0] + "_temp.txt"
+            with open(temp_path, "a+") as outfile:
```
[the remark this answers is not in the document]
KKsharma99: Done

mindmeld/auto_annotator.py
KKsharma99: Added

KKsharma99: Added

tests/test_auto_annotator.py
```diff
@@ -0,0 +1,227 @@
...
+Tests for `auto_annotator` module
+"""
+import pytest
+
+from mindmeld.auto_annotator import SpacyAnnotator
+
+
+@pytest.fixture(scope="module")
+def get_SA():
+    return SpacyAnnotator("./kwik_e_mart")
```
minhtuev, line 17:
```suggestion
    return SpacyAnnotator(kwik_e_mart_app_path)
```

tests/test_auto_annotator.py
```diff
@@ -0,0 +1,227 @@
...
+
+Tests for `auto_annotator` module
+"""
+import pytest
+
+from mindmeld.auto_annotator import SpacyAnnotator
+
+
+@pytest.fixture(scope="module")
+def get_SA():
```
minhtuev, line 16:
```suggestion
def spacy_annotator(kwik_e_mart_app_path):
```

vijay120 (approved):

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,437 @@
...
+  - supports the development of custom Annotators.
+
+.. note::
+
+   The examples in this section require the :doc:`HR Assistant <../blueprints/hr_assistant>` blueprint application. To get the app, open a terminal and run ``mindmeld blueprint hr_assistant``.
+
+.. warning::
+
+   Changes by an Auto Annotator cannot be undone and Mindmeld does not backup query data. We recommend using version control software such as Github.
+
```
minhtuev, line 16: We should add a Quick Start section telling how to use these commands directly:
```
mindmeld annotate --app-path "hr_assistant"
mindmeld unannotate --app-path "hr_assistant"

Alternatively
.. code-block:: python
	from mindmeld.auto_annotator import SpacyAnnotator 
	sa = SpacyAnnotator(app_path="hr_assistant")
	sa.annotate()

File: mindmeld/auto_annotator.py [hunk 56]
```diff
@@ -0,0 +1,691 @@
...
+
+        Args:
+            entity (str): Name of entity to annotate.
+
+        Returns:
+            bool: Whether entity is valid.
+        """
+        raise NotImplementedError("Subclasses must implement this method")
+
+    def annotate(self):
```
minhtuev, line 140: should we pass an optional configuration in case the developers want to test, etc

```suggestion
    def annotate(self, conf=None):
```

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,691 @@
...
+        """ Annotate data based on configurations in the config.py file.
+        """
+        file_entities_map = self.annotate_file_entities_map
+        self._modify_queries(file_entities_map, action=AnnotatorAction.ANNOTATE)
+
```
minhtuev, line 145: same?

mindmeld/auto_annotator.py
KKsharma99: Annotator classes accept a custom config upon instantiation. For testing, developers can pass the config into the class instead of at the function level.

```
custom_config = { ... }

custom_annotator = CustomAnnotator(
     app_path="hr_assistant", config = custom_config
)
custom_annotator.annotate()
```

source/userguide/auto_annotator.rst
KKsharma99: Added a Quick Start section

mindmeld/components/_config.py
```diff
@@ -928,6 +937,37 @@ def get_nlp_config(app_path=None, config=None):
...
+        config (dict, optional): A config object to use. This will
+            override the config specified by the app's config.py file.
+            If necessary, this object will be expanded to a fully
+            specified config object.
+
+    Returns:
+        dict: The automatic annotator config.
+    """
+
+    if config:
```
minhtuev, line 955: is `config` here a useful param?  can we just remove it?

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,489 @@
...
+Let's take a look at the allowed values for each setting in an Auto Annotator configuration.
+
+
+``'annotator_class'`` (:class:`str`): The class in auto_annotator.py to use for annotation when invoked from the command line. By default, :class:`SpacyAnnotator` is used. 
+
+``'overwrite'`` (:class:`bool`): Whether new annotations should overwrite existing annotations in the case of a span conflict. False by default. 
+
+``'annotate'`` (:class:`list`): A list of annotation rules where each rule is represented as a dictionary. Each rule must have four keys: :attr:`domains`, :attr:`intents`, :attr:`files`, and :attr:`entities`.
+Annotation rules are combined internally to create Regex patterns to match selected files. The character :attr:`.*` can be used if all possibilities in a section are to be selected, while possibilities within
+a section are expressed with the usual Regex special characters, such as :attr:` . ` for any single character and :attr:`|` to represent "or".
```
minhtuev, line 365: I think the `.` character is not rendered correctly
image

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,489 @@
...
+
+.. warning::
+
+   Changes by an Auto Annotator cannot be undone and Mindmeld does not backup query data. We recommend using version control software such as Github.
+
+Quick Start
+-----------
+This section briefly explains the use of the unannotate and annotate commands. For more details, read the next section.
+
+First, copy and paste the config below into the :attr:`config.py` of the mindmeld app that you would like to remove annotations.
```
minhtuev, line 21:
```suggestion
First, copy and paste the configuration below into the :attr:`config.py` of the MindMeld app that you would like to operate on.
```

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,489 @@
...
+
+   The examples in this section require the :doc:`HR Assistant <../blueprints/hr_assistant>` blueprint application. To get the app, open a terminal and run ``mindmeld blueprint hr_assistant``.
+
+.. warning::
+
+   Changes by an Auto Annotator cannot be undone and Mindmeld does not backup query data. We recommend using version control software such as Github.
+
+Quick Start
+-----------
+This section briefly explains the use of the unannotate and annotate commands. For more details, read the next section.
```
minhtuev, line 19:
```suggestion
This section briefly explains the use of the ``annotate`` and ``unannotate`` commands. For more details, read the next section.
```

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,489 @@
...
+	AUTO_ANNOTATOR_CONFIG = { 
+
+		"annotator_class": "SpacyAnnotator",
+		"overwrite": False, 
+		"annotate": [{"domains": ".*", "intents": ".*", "files": ".*", "entities": ".*"}],
+		"unannotate_supported_entities_only": True, 
+		"unannotate": [{"domains": ".*", "intents": ".*", "files": ".*", "entities": ".*"}],
+	}
+
+To unannotate you can run python code or use the command-line.
```
minhtuev, line 34:
```suggestion
To "unannotate" the files, you can run Python code or use the command-line.
```

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,489 @@
...
+Python:
+
+.. code-block:: python
+
+	from mindmeld.auto_annotator import SpacyAnnotator 
+	sa = SpacyAnnotator(app_path="hr_assistant")
+
+	sa.unannotate()
+
+Command-line:
```
minhtuev, line 45: Let's show the Python command first

source/userguide/auto_annotator.rst
minhtuev: Can you explain the `unannotate` command here more? Why do we recommend people to `unannotate` first before running `annotate`?

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,489 @@
...
+.. code-block:: console
+
+	mindmeld unannotate --app-path "hr_assistant"
+
+
+If you see the following message, you need to update the unannotate parameter in your custom :attr:`AUTO_ANNOTATOR_CONFIG` dictionary in :attr:`config.py`. You can refer to the config specifications in the section above.
+
+.. code-block:: console
+
+	'unannotate' is None in the config. Nothing to unannotate.
```
minhtuev, line 98: We probably should update this error to be clearer?

```suggestion
	'unannotate' field is not configured or misconfigured in the `config.py`. We can't find any file to unannotate.
```

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,489 @@
...
+				"intents": "get_date_range_aggregate", 
+				"files": "train.txt",
+				"entities": "sys_time", 
+			}
+		], 
+	}
+
+.. note::
+
+	The content of :attr:`annotate` in the config has no effect on unannotation. Similarly, :attr:`unannotate` in the config has no affect on annotation. These processes are independent and are only affected by the corresponding parameter in the config.
```
minhtuev, line 125: What is the reason for having separate params for `annotate` and `unannotate`? should they be operating on the same files?

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,489 @@
...
+	class CustomAnnotator(Annotator):
+		""" Custom Annotator class used to generate annotations.
+		"""
+
+		def __init__(self, app_path, config=None):
+			super().__init__(app_path=app_path, config=config)
+			
+			# Add additional attributes if needed
+
+		def parse(self, sentence, entity_types=None):
```
minhtuev, line 418: where do the entity types come from? probably we should make it clear that these entity types are the ones we want to annotate

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,489 @@
...
+
+			return entities
+
+		def valid_entity_check(self, entity):
+			""" 
+			Args: entity (str): Name of entity to annotate.
+			Returns: bool: Whether entity is valid.
+			"""
+			entity = entity.lower().strip()
+			supported_entities = [
```
minhtuev, line 437: we should make this into a class/instance variable

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,691 @@
...
+            if remove_entities == ["*"]:
+                is_supported_entity = self.valid_entity_check(query_entity.entity.type)
+                if remove_supported_only and not is_supported_entity:
+                    keep_entities.append(query_entity)
+            elif query_entity.entity.type not in remove_entities:
+                keep_entities.append(query_entity)
+        processed_query.entities = tuple(keep_entities)
+
+    @abstractmethod
+    def parse(self, sentence, **kwargs):
```
minhtuev, line 313: we should implement the third abstract method

```
@property
@abstractmethod
def supported_entity_types(self):
   ...
```

this method will return the supported entity types for this particular annotator

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,691 @@
...
+        if self._resolve_exact_match(entity):
+            return entity
+        candidates = self.duckling.get_candidates_for_text(entity["body"])
+        if self._resolve_largest_substring(
+            entity, candidates, entity_types=["sys_number"], is_time_related=False
+        ):
+            return entity
+
+    def _resolve_money(self, entity, sentence):
+        currency_symbols = ["$", "€", "£", "¥", "₹", "₩", "฿"]
```
minhtuev, line 568: make this into a class constant

minhtuev (commented):
Adding some comments, mostly about the documentation

mindmeld/auto_annotator.py
ritvikshrivastava: Does the resolution here support all currency symbols? If yes, then there is a list of all currency symbols defined in ``constants.py``, you can reuse it.

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,691 @@
...
+    ANNOTATE = "annotate"
+    UNANNOTATE = "unannotate"
+
+
+class Annotator(ABC):
+    """
+    Abstract Annotator class that can be used to build a custom Annotation class.
+    """
+
+    def __init__(self, app_path, config=None, resource_loader=None):
```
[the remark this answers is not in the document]
KKsharma99: Sure and good point, will remove the resource loader param

mindmeld/components/_config.py
KKsharma99: Removing config from this function entirely. This function will now only be called from auto_annotator.py if a config is not present.

`self.config = config or get_auto_annotator_config(app_path=app_path)
`

source/userguide/auto_annotator.rst
KKsharma99: The Python command for "annotate" and "unannotate" is shown before the CLI command

Unannotate
- Python
- CLI

Annotate
- Python
- CLI

Is there a different structure that you would prefer?

source/userguide/auto_annotator.rst
KKsharma99: In a typical workflow, you would ofcourse `annotate` and then `unannotate`. Since a blueprint with existing annotations is being used, it seems easier to demonstrate unannotation first and then demonstrate annotation on unannotated text.

If this seems confusing, I can alternatively demonstrate annotation first and have users set "overwrite" to True and then demonstrate unannotating the all or a subset of the newly added annotations? This actually may better represent the real-world use case of this feature.

mindmeld/auto_annotator.py
KKsharma99: Done

source/userguide/auto_annotator.rst
KKsharma99: It may not always be the case that you would want to unannotate and annotate the same entities. Often times you would like to annotate a set of entities and then unannotate a subset that you realised you would rather not have.

By having these as separate parameters, a developer doesn't have to copy and paste a different set of rules each time they switch methods and doesn't risk accidentally unannotating their existing annotations

mindmeld/auto_annotator.py
KKsharma99: Added

mindmeld/auto_annotator.py
KKsharma99: Sounds good. With this abstract method valid_entity_check() will not need to be implemented because valid_entity_check() will simply check if the given entity is in the list provided from the implementation of this abstract method.

source/userguide/auto_annotator.rst
KKsharma99: Updating the docs for this function:

```
        Args:
            sentence (str): Sentence to detect entities.
            entity_types (list): List of entity types to annotate. If None, all
                possible entity types will be annotated.
```

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,724 @@
...
+        """
+        raise NotImplementedError("Subclasses must implement this method")
+
+    def annotate(self, **kwargs):
+        """ Annotate data based on configurations in the config.py file.
+
+        Args:
+            kwargs (dict, optional): Configuration overrides can be passed in as arguments.
+        """
+        for key, value in kwargs.items():
```
minhtuev, line 152: we probably do not want to permanently change the config, maybe just overriding the config for this operation

mindmeld/auto_annotator.py
```diff
@@ -0,0 +1,724 @@
...
+
+    def unannotate(self, **kwargs):
+        """ Unannotate data based on configurations in the config.py file. 
+
+        Args:
+            kwargs (dict, optional): Configuration overrides can be passed in as arguments.
+            unannotate_all (bool): Unannotate all entities.
+        """
+        for key, value in kwargs.items():
+            if key == "unannotate_all" and value:
```
minhtuev, line 168: same, we probably should not modify the base configuration

mindmeld/auto_annotator.py
minhtuev: @KKsharma99 - have you addressed this issue?

minhtuev (commented):
A few more comments

mindmeld/auto_annotator.py
KKsharma99: Yes, the current code correctly resolves "35.85%" to 0.3585

```
        possible_values = []
        for candidate in candidates:
            if candidate["entity_type"] == "sys_number":
                value = candidate["value"]["value"]
                if isinstance(value, float):
                    entity["value"]["value"] = value / 100
                    return entity
                else:
                    possible_values.append(value)
        entity["value"]["value"] = max(possible_values) / 100
        return entity
```

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,489 @@
...
+Getting Custom Parameters from the Config
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:attr:`spacy_model` is an example of an optional parameter in the config that is relevant only for a specific :class:`Annotator` class.
+
+.. code-block:: python
+
+	AUTO_ANNOTATOR_CONFIG = { 
+		... 
+		"spacy_model": "en_core_web_md",
```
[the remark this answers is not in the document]
KKsharma99: I haven't yet tried using another language model but I have seen on Spacy's website that the output of NER for models that support it uses the same interface, so it should likely work so long as duckling also supports that language as well

A developer can't use a non-english model currently since an exception would be thrown if the model is not in this list: [EN_CORE_WEB_SM, EN_CORE_WEB_MD, EN_CORE_WEB_LG]

mindmeld/auto_annotator.py
KKsharma99: Added

source/userguide/auto_annotator.rst
KKsharma99: Does this happen automatically if you use @property?

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,489 @@
...
+		... 
+	}
+
+:class:`SpacyAnnotator` checks if :attr:`spacy_model` exists in the config, and if it doesn't, it will use the default value of "en_core_web_lg".
+
+.. code-block:: python
```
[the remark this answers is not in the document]
KKsharma99: Removed

mindmeld/auto_annotator.py
KKsharma99: Without modifying the base config, I would need to pass in the current config downstream to each downstream function as these functions get configuration details from the self attribute and not from a function parameter.

Should I add a "config" param for the downstream functions in order to use an alternate config?

minhtuev: Yup let's do that instead

source/userguide/auto_annotator.rst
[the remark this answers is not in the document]
minhtuev: I see - can you run a quick test to see if that can work with non-english language? in that case, we should allow support for non-english language, but just add a warning if it's not in the list.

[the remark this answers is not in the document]
KKsharma99: I was able to load up a Spacy Spanish annotator using es_core_news_lg. I had it parse the following text:

`text = "El domingo, el Gobierno de Navarra, que ha registrado 675,3 casos por cada 100.000 habitantes en las dos últimas semanas, ordenó que todos los bares y restaurantes reduzcan su capacidad al 30% y cierren a las 22.00 horas."`

It was only able to identify "Gobierno de Navarra" as location and missed the numbers and time-related phrases. When I passed the same text into duckling (language="es"), duckling is able to identify many more entities (but not able to collapse candidates to a final set of course).

Will go ahead and add the support and warning as suggested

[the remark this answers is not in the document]
KKsharma99: Done

source/userguide/auto_annotator.rst
KKsharma99: Could we do the following in the class init as an alternative to using @ property? 

`self.supported_entity_types = self.get_supported_entity_types()`

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,457 @@
...
+
+.. warning::
+	By default, all files in all intents across all domains will be annotated with all supported entities. Before annotating consider including custom annotation rules in :attr:`config.py`. 
+
+``'unannotate_supported_entities_only'`` (:class:`boolean`): By default, when the unannotate command is used only entities that the Annotator can annotate will be eligible for removal. 
+
+``'unannotate'`` (:class:`list`): List of annotation rules in the same format as those used for annotation. These rules specify which entities should have their annotations removed. By default, :attr:`files` is None.
+
+``'spacy_model'`` (:class:`str`): :attr:`en_core_web_lg` is used by default for the best performance. Alternative options are :attr:`en_core_web_sm` and :attr:`en_core_web_md`. This parameter is optional and is specific to the use of the :class:`SpacyAnnotator`.
+If the selected English model is not in the current environment it will automatically be downloaded. Refer to Spacy's documentation to learn more about their `English models <https://spacy.io/models/en>`_.
```
minhtuev, line 371: You can add a note about non-English models here.

source/userguide/auto_annotator.rst
```diff
@@ -0,0 +1,457 @@
...
+				entity_types (list): List of entity types to parse. If None, all
+					possible entity types will be parsed.
+			Returns: entities (list): List of entity dictionaries.
+			"""
+
+			# TODO: Add custom parse logic
+
+			return entities
+
+		def valid_entity_check(self, entity):
```
minhtuev, line 410: Update this to `supported_entity_types`

minhtuev (approved):
Looks good, some comments about the documentation!

Issue #18, opened 2019-05-22
request.entities should be an immutable map


minhtuev:
`request.entities` should be an immutable map. Currently, it is a dictionary which is mutable.

minhtuev:
Should do the same for `history`, `nbest_transcripts_text`, `nbest_transcripts_entities`, `nbest_aligned_entities`

minhtuev:
Fixed by @KKsharma99

PR #201, opened 2020-08-11
Bugfix / Request.entities should be an immutable map

KKsharma99:
This PR resolves bug https://github.com/cisco/mindmeld/issues/18 

Added a custom converter function for tuple attributes of the Request class to ensure immutability (request.py). Updated the DialogueResponder's to_json() method to convert immutable maps in tuples to dicts for serialization (dialogue.py).

(Other changes are formatting changes from Black.)


mindmeld/components/dialogue.py
```diff
@@ -1212,6 +1214,11 @@ def to_json(instance):
         for attribute, value in vars(instance).items():
             if isinstance(value, (Params, Request, FrozenParams)):
                 serialized_obj[attribute] = DialogueResponder.to_json(value)
+            elif isinstance(value, tuple) and all(
+                isinstance(item, immutables.Map) for item in value
+            ):
+                items = [dict(item) for item in value]
+                serialized_obj[attribute] = tuple(items)
```
minhtuev, line 1221: You can use the generator expression directly for tuple
```
In [1]: tuple(i for i in range(5))
Out[1]: (0, 1, 2, 3, 4)
```

mindmeld/components/request.py
```diff
@@ -295,6 +295,10 @@ class FrozenParams(Params):
     dynamic_resource = attr.ib(default=immutables.Map(), converter=immutables.Map)
 
 
+def list_elems_to_immutable_map(value):
```
minhtuev, line 298: what is the type/structure for value?

KKsharma99: value represents the object that is being passed in for the given attribute so its type is the same as the type of the object

KKsharma99:
> Looks good, can you add a test for this bug?

Added tests

mindmeld/components/request.py
vijay120: Isnt the value supposed to be just a dictionary? For instance `immutables.Map(i)` cannot be done if `i = "cat"`, no?

KKsharma99: Yes, in this case it would be a tuple containing one or more dictionaries as that is the type/structure we are expecting for the respective Request attributes. It might be helpful if I change the name of this custom attrs converter to "tuple_elems_to_immutable_map" as it expects tuples not lists.

vijay120: It would also be good to check if value is a list of maps before converting it as well.

PR #205, opened 2020-08-18
Adding support for custom resolution in custom eval function

ritvikshrivastava:
This PR allows developers to take control of custom slot-entity resolution in the custom eval function for slot-filling. 

- For example a case in the Spanish blueprint where basic units of distance are not recognized by duckling, a statement such as "my height is 1 meter and 75 centimeters" becomes "yo tengo {1|sys_number} {metro|unit} y {75|sys_number} {centimetros|unit} de estatura". This requires multiple entities to make up one slot. It can be handled by the developer now in the custom validation function and a single/joint value returned for the slot as a 'custom resolved' value.

- Cases where there is some conversion required (such as weight between kilos and pounds), instead of doing so inside the handler later, this enables the flexibility to do so in the custom validation itself.

- retains the earlier custom validation logic entirely. Adds an update, documentation update and a corresponding unit test.

### DIFF
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -872,7 +872,7 @@ class AutoEntityFilling:
             # value is returned, then the validation succeeds.
 
             _validity = slot.custom_eval(request)
-            if not _validity:
+            if _validity is False:
                 # For checking 'false' return cases
                 return False, _resolved_value
       
--- a/mindmeld/components/dialogue.py
+++ b/mindmeld/components/dialogue.py
@@ -866,13 +866,21 @@ class AutoEntityFilling:
                 return False, _resolved_value
 
         if slot.custom_eval:
-            # Custom validation using function provided by developer. Should return True/False
-            # for validation status. If true, then continue, else fail overall validation.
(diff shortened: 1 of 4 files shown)


mindmeld/components/dialogue.py
```diff
@@ -866,13 +866,21 @@ def _validate(self, request, slot):
...
-
-            if not slot.custom_eval(request):
+            # Custom validation using function provided by developer.
+            # Should return True/False/Custom Resolution value for validation status.
+            # If false, overall validation fails. If either true or a custom resolved
+            # value is returned, then the validation succeeds.
+
+            _validity = slot.custom_eval(request)
+            if not _validity:
+                # For checking 'false' return cases
```
serapio, line 876: We should check `is False`, so that the developer can use falsy values like `0` and `None` as valid slot values.

ritvikshrivastava: Done!

minhtuev:
Is it possible to add a test that covers the case that this feature covers (i.e. multiple entities to make one slot)?

Issue #203, opened 2020-08-14
sklearn.externals import joblib is deprecated


jimilpatel24:
You can now directly do a **import joblib**

File to be edited - https://github.com/cisco/mindmeld/blob/master/mindmeld/gazetteer.py

See this for more details - https://github.com/scikit-optimize/scikit-optimize/pull/776

PR #207, opened 2020-08-18
Direct import of joblib

minhtuev:
You can now directly do a import joblib

See this for more details - scikit-optimize/scikit-optimize#776

https://github.com/cisco/mindmeld/issues/203


vijay120:
Is this backwards compatible?

minhtuev:
@vijay120 : it passed all the tests, so I think it's backward compatible

Issue #208, opened 2020-08-18
Backward incompatibility for entity recognition in mindmeld


jimilpatel24:
When you load the bots built in the pre-4.3 release (4.2.10 for example), we see the below error. The error is related to a change in the EntityLabelEncoder class, where you added self.system_entity_recognizer = SystemEntityRecognizer.get_instance(). This causes old apps to fail since they don't have the field.

Error: **EntityLabelEncoder** object has no attribute **system_entity_recognizer**

minhtuev:
Reopening this issue since we continue to observe this error.

I am able to reproduce with the kwik-e-mart blueprint for building MM4.2.11 and running with MM4.3.1:
```
Fitting entity recognizer: domain='store_info', intent='greet'
There are no labels in this label set, so we don't fit the model.
Fitting entity recognizer: domain='store_info', intent='help'
There are no labels in this label set, so we don't fit the model.
Fitting entity recognizer: domain='store_info', intent='exit'
There are no labels in this label set, so we don't fit the model.
Fitting entity recognizer: domain='store_info', intent='find_nearest_store'
There are no labels in this label set, so we don't fit the model.
Saving intent classifier: domain='store_info'
Saving entity classifier: domain='store_info', intent='get_store_hours'
Saving role classifier: domain='store_info', intent='get_store_hours', entity='sys_time'
Saving role classifier: domain='store_info', intent='get_store_hours', entity='store_name'
Saving entity classifier: domain='store_info', intent='greet'
Saving entity classifier: domain='store_info', intent='help'
Saving entity classifier: domain='store_info', intent='exit'
Saving entity classifier: domain='store_info', intent='find_nearest_store'
(mm42) ➜  blueprints git:(master) ✗ cd ..
(mm42) ➜  mindmeld-blueprints git:(master) ✗ cd blueprints
(mm42) ➜  blueprints git:(master) ✗ conda deactivate
(base) ➜  blueprints git:(master) ✗ conda activate mm43
(mm43) ➜  blueprints git:(master) ✗ ls
banking_blueprint food_ordering     home_assistant    hr_assistant      kwik_e_mart       log               template          video_discovery
(mm43) ➜  blueprints git:(master) ✗ python -m kwik_e_mart converse
[...]

PR #209, opened 2020-08-20
Initialize system_entity_recognizer attr for EntityLabelEncoder

minhtuev:
For applications already built with an older version of MM, the EntityLabelEncoder is not initialized with the system entity recognizer. This PR should address that.
https://github.com/cisco/mindmeld/issues/208

### DIFF
--- a/mindmeld/models/model.py
+++ b/mindmeld/models/model.py
@@ -1295,6 +1295,9 @@ class EntityLabelEncoder(LabelEncoder):
             list: A list of decoded labels per query
         """
         # TODO: support decoding multiple queries at once
+        if not hasattr(self, "system_entity_recognizer"):
+            # app built with older version of MM (< 4.3) does not save label encoder with sys recognizer
+            self.system_entity_recognizer = SystemEntityRecognizer.get_instance()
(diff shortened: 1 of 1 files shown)


mindmeld/models/model.py
```diff
@@ -1295,6 +1295,9 @@ def decode(self, tags_by_example, **kwargs):
             list: A list of decoded labels per query
         """
         # TODO: support decoding multiple queries at once
+        if not hasattr(self, "system_entity_recognizer"):
+            # app built with older version of MM (< 4.3) does not save label encoder with sys recognizer
+            self.system_entity_recognizer = SystemEntityRecognizer.get_instance()
```
vijay120, line 1300: Why do it in `decode` method instead of in `__init__ ` for example?

minhtuev: The initialization is already inside the `__init__` method; I suspect `__init__` is not being called for deserialized objects.

https://github.com/cisco/mindmeld/pull/209/files#diff-67b2786b2832b1a5d1ce4b306d7f2b67L1261

PR #215, opened 2020-08-27
Adding sentiment feature to domain and intent classifiers

vembar:
This PR adds a new sentiment feature that can be used to train the domain and intent classifiers. We use [nltk vader](https://www.nltk.org/howto/sentiment.html) library to extract the feature. To use it, you can add it to the config as follows

```
{
...
"features":{
    ...
    "sentiment": {"analyzer": <"composite"/"discrete">}
    }
}
```

`analyzer` controls whether we add a single feature ie, `composite` or 3 features ie, `discrete` representing positive, neutral and negative sentiment for each query.

Example:
```
>>>ic.view_extracted_features('you are so bad')
{'bag_of_words|length:1|ngram:you': 1, 'bag_of_words_stemmed|length:1|ngram:you': 1, 'bag_of_words|length:1|ngram:are': 1, 'bag_of_words_stemmed|length:1|ngram:are': 1, 'bag_of_words|length:1|ngram:so': 1, 'bag_of_words_stemmed|length:1|ngram:so': 1, 'bag_of_words|length:1|ngram:bad': 1, 'bag_of_words_stemmed|length:1|ngram:bad': 1,
 'in_vocab:IV|freq_bin:0': 0.25, 'in_vocab:IV|freq_bin:1': 0.25, 'in_vocab:IV|freq_bin:2': 0.25, 'in_vocab:IV|freq_bin:5': 0.25, 'exact|query:<OOV>': 1000, 
'tokens': 4, 'chars': 14, 'tokens_log': 1.6094379124341003, 'chars_log': 2.70805020110221,
 **'sentiment|composite': -0.6696**}

>>> ic.view_extracted_features('you are so bad')
{'bag_of_words|length:1|ngram:you': 1, 'bag_of_words_stemmed|length:1|ngram:you': 1, 'bag_of_words|length:1|ngram:are': 1, 'bag_of_words_stemmed|length:1|ngram:are': 1, 'bag_of_words|length:1|ngram:so': 1, 'bag_of_words_stemmed|length:1|ngram:so': 1, 'bag_of_words|length:1|ngram:bad': 1, 'bag_of_words_stemmed|length:1|ngram:bad': 1, 
'in_vocab:IV|freq_bin:0': 0.25, 'in_vocab:IV|freq_bin:1': 0.25, 'in_vocab:IV|freq_bin:2': 0.25, 'in_vocab:IV|freq_bin:5': 0.25, 'exact|query:<OOV>': 1000, 
'tokens': 4, 'chars': 14, 'tokens_log': 1.6094379124341003, 'chars_log': 2.70805020110221,
 **'sentiment|positive': 0.0, 'sentiment|negative': 0.6, 'sentiment|neutral': 0.4**}

```


minhtuev:
Do we recommend the users to use either "combined" or "separate" by default? Should we also add another field in the `request` object for sentiment?

vembar:
> Do we recommend the users to use either "combined" or "separate" by default? More broadly should we also add another field in the `request` object for sentiment?

Yeah, by default we will set it to `combined`. Not sure I understand the latter question - can you elaborate on what other fields we may require?

minhtuev:
Currently, we have NLP fields like "domain", "intent", etc in the request object. Should we add another field called "sentiment" whose value can be a single score or separate scores (positive, negative, neutral)?

vembar:
> Currently, we have NLP fields like "domain", "intent", etc in the request object. Should we add another field called "sentiment" whose value can be a single score or separate scores (positive, negative, neutral)?

Oh, I see. I think it's better to not do that right now since we don't plan on actively managing/contributing to the Vader library. We would not have control on the outputs (For example, it's not very robust with negations), so we would not be able to handle any user requests or corrections to the model.

mindmeld/models/model.py
```diff
@@ -44,6 +44,7 @@
     get_label_encoder,
     ingest_dynamic_gazetteer,
     register_label,
+    SENTIMENT_ANALYZER,
```
vijay120, line 47: Nit: maybe regroup this var closer to the other uppercase vars

mindmeld/models/query_features.py
```diff
@@ -991,6 +992,32 @@ def _extractor(query, resources):
...
+        (function) A feature extraction function that takes in a query and \
+            returns sentiment values across positive, negative and neutral
+
+    """
+    del kwargs
+
+    def _extractor(query, resources):
+        text = query.text
+        sentiment_scores = resources[SENTIMENT_ANALYZER].polarity_scores(text)
+        if intensity_type == 'combined':
```
vijay120, line 1010: Maybe we can use `compound` instead of `combined ` for the feature name?

arushir:
How about renaming `intensity_type` to `signal_type` or just `type`? The word "intensity" to describe this doesn't seem very intuitive to me.

source/userguide/domain_classifier.rst
```diff
@@ -355,6 +355,22 @@ Let's take a look at the allowed values for each setting in a domain classifier
...
+  |                       | Examples:                                                                                                  |
+  |                       |                                                                                                            |
+  |                       | ``{'intensity_type': 'combined'}``                                                                         |
+  |                       | - extracts a single feature representing the sentiment normalized to be between -1 (extreme negative) to   |
+  |                       |   +1 (extreme positive).                                                                                   |
+  |                       |                                                                                                            |
+  |                       | ``{'intensity_type': 'separate'}``                                                                         |
+  |                       | - extracts three features, one for each sentiment (positive, neutral and negative)                         |
```
arushir, line 372: Include a note on the range for these three signals as well

vembar:
> * Should we allow the users to have a way to configure the sentiment analyzer to use other libraries rather than vader?

They can still do that with custom features. AFAIK there are no better out-of-the-box, open-source/free alternatives currently.

> * Otherwise, should we include in the documentation a way to add a sentiment as a feature for the model?

That is what we are trying to do with this PR. If you mean documentation on adding their own sentiment feature, I think that is redundant at this point since this sort of applies to every in-built feature (like sys-entities, stemming etc). 

If folks feel strongly about this, I can add a note in the documentation that people can add their own sentiment feature through custom features.

minhtuev:
@vembar : yes let's add a note in the documentation regarding the custom feature for sentiment analysis

mindmeld/models/model.py
vembar: Done

Issue #180, opened 2020-07-10
Configure MindMeld tokenizer to preserve special characters


minhtuev:
We want to configure the tokenizer so that it can preserve certain special characters (such as "="). Currently the MindMeld [Tokenizer](https://github.com/cisco/mindmeld/blob/master/mindmeld/tokenizer.py) will convert most characters to empty strings except for a few special characters that are used in annotation. We want to be able to configure the MindMeld Tokenizer to preserve a certain set of characters.

This is useful because some product IDs might contain special characters in them ("S9500UK9-169=") which will be removed by the Tokenizer. When we train entity recognizer on the final string, it will not be able to recognize the original product ID in the input sequence ("S9500UK9-169" instead of "S9500UK9-169=").

PR #216, opened 2020-08-27
Configures MindMeld tokenizer to preserve special characters

ritvikshrivastava:
This PR does the following:

1. Resolves issue: #180.

    Creates an optional configuration field ``TOKENIZER_CONFIG`` for config.py.

     Allows developers to define regex patterns to detect special characters. Is supported when ``keep_special_chars`` is set to True for the tokenizer. 

    Usage examples:
```
TOKENIZER_CONFIG = {
    "allowed_patterns": ["\w+"]
}
``` 
for alphanumeric strings

or 
```
TOKENIZER_CONFIG = {
    "allowed_patterns": ["[a-zA-Z]"]
}
```
 for alphabet only strings

If no config is defined by developer, then ``DEFAULT_TOKENIZER_CONFIG`` is called. The ``DEFAULT_TOKENIZER_CONFIG`` for MM's current special character regex is now defined in the config file itself, populated using the ``_get_default_regex`` function and is passed into the ``allowed_patterns`` key.

2. Replaces the small/fixed list ("$¥") of currency-special characters with the full set defined under unicode.
3. Fixes lint issues


minhtuev:
Looks good, should we just configure `DEFAULT_TOKENIZER_CONFIG["pattern"]` explicitly to return the currency symbols and everything by default (instead of computing it again inside the tokenizer)?

mindmeld/components/_config.py
```diff
@@ -455,6 +455,10 @@
     },
 }
 
+DEFAULT_TOKENIZER_CONFIG = {
+    "patterns": [],
```
vijay120, line 459: rename to `whitelist_patterns`

mindmeld/tokenizer.py
```diff
@@ -50,60 +53,61 @@ def _init_regex(self):
 
         exception_chars = "\@\[\]\|\{\}'"  # noqa: W605
 
-        # This is more of a tactical fix at this moment. Will probably need a more generic way
-        # to handle all currency symbols.
-        currency_symbols = "$¥"
+        # fetches all currency symbols in unicode
+        currency_symbols = u"".join(
+            chr(i) for i in range(0xFFFF) if unicodedata.category(chr(i)) == "Sc"
+        )
```
vijay120, line 59: Could you write a comment explaining what the logic is here?

mindmeld/tokenizer.py
```diff
@@ -171,9 +175,21 @@ def _init_regex(self):
...
-            "(%s)" % ")|(".join(keep_special_regex_list), re.UNICODE
-        )
+        # and add app-specific regex config to the list if specified
+
+        if self.config["patterns"]:
+            self.keep_special_compiled = re.compile(
+                "((%s))&(%s)"
+                % (
+                    ")|(".join(keep_special_regex_list),
+                    ")|(".join(self.config["patterns"]),
```
vijay120, line 185: So here the whitelist set of chars are combined with the `keep_special_regex_list `? Whats the regex pattern here? Is it, if the `keep_special_regex_list ` set of chars or `self.config["patterns"]` patterns match, whitelist, ie allow them?

ritvikshrivastava: Yes. In the event of default configuration we continue with the current mindmeld regex, otherwise a combination of the current regex and the new pattern. 

Alternatively we can also do the following: in the case of a developer-defined tokenizer config, only apply that and not the default MM special chars list.

mindmeld/components/_config.py
arushir: Can we name it `allow_patterns` instead? There've recently been suggestions to use allowlist/blocklist instead of whitelist/blacklist to be more racially sensitive

arushir: Should we include the existing built-in patterns here as part of the config? Or at least a note on where to find them, so it's clear to the developer what is currently being used

ritvikshrivastava: Done!

mindmeld/tokenizer.py
ritvikshrivastava: Added, let me know if this works

mindmeld/components/_config.py
```diff
@@ -912,3 +918,102 @@ def get_nlp_config(app_path=None, config=None):
...
+        + "[^\w\d\s"  # noqa: W605
+        + exception_chars
+        + "]+(?=s)"
+    )
+
+    keep_special_regex_list.append("?P<underscore>_")  # noqa: W605
+    keep_special_regex_list.append("?P<begspace>^\s+")  # noqa: W605
+    keep_special_regex_list.append("?P<trailspace>\s+$")  # noqa: W605
+    keep_special_regex_list.append("?P<spaceplus>\s+")  # noqa: W605
+    keep_special_regex_list.append("?P<bar> '|' ")  # noqa: W605
```
serapio, line 990: this group should be named `apos_edge` or `apos_space`

mindmeld/tokenizer.py
```diff
@@ -112,43 +66,29 @@ def _init_regex(self):
...
-        keep_special_regex_list.append("?P<begspace>^\s+")  # noqa: W605
-
         regex_list.append("?P<trailspace>\s+$")  # noqa: W605
-        keep_special_regex_list.append("?P<trailspace>\s+$")  # noqa: W605
-
         regex_list.append("?P<spaceplus>\s+")  # noqa: W605
-        keep_special_regex_list.append("?P<spaceplus>\s+")  # noqa: W605
-
-        keep_special_regex_list.append("?P<bar> '|' ")  # noqa: W605
         regex_list.append("?P<bar> '|' ")  # noqa: W605
```
serapio, line 89: same here

mindmeld/tokenizer.py
```diff
@@ -170,10 +110,15 @@ def _init_regex(self):
...
         }
 
-        # Create a regular expression  from the dictionary keys
+        # Create compiled regex expressions
         self.keep_special_compiled = re.compile(
-            "(%s)" % ")|(".join(keep_special_regex_list), re.UNICODE
+            "(%s)"
+            % (
+                ")|(".join(self.config["allowed_patterns"]),
+            ),
```
serapio, line 118: If allowed_patterns doesn't make a well-formed regex here, is the failure interpretable for the developer?

serapio (approved):
minor comments

mindmeld/tokenizer.py
```diff
@@ -232,7 +177,11 @@ def multiple_replace(self, text, compiled):
             str: The text with replacement specified by self.replace_lookup
         """
         # For each match, look-up corresponding value in dictionary
-        return compiled.sub(self._one_xlat, text)
+        try:
+            return compiled.sub(self._one_xlat, text)
+        except KeyError:
+            # In case of custom/app-specific tokenizer configuration
```
minhtuev, line 183: Should we add logging here as well

mindmeld/components/_config.py
```diff
@@ -455,6 +456,11 @@
     },
 }
 
+DEFAULT_TOKENIZER_CONFIG = {
+    # populated in the `get_tokenizer_config` func
+    "allowed_patterns": [],
```
minhtuev, line 461: can we call directly here:
```
"allowed_patterns": _get_default_regex(exclude_from_norm)

File: mindmeld/tokenizer.py [hunk 9]
```diff
@@ -45,65 +48,16 @@ def _init_regex(self):
...
-            + "]+"
-            + "(?=\|)"
-        )  # noqa: W605
-        keep_special_regex_list.append(
-            "?P<escape2>(?P<escape2_replace>[\]\}]+)"
-            + "[^\w\d\s"  # noqa: W605
-            + exception_chars
-            + "]+(?=s)"
-        )  # noqa: W605
+        to_exclude = currency_symbols + "".join(self.exclude_from_norm)
```
minhtuev, line 60: if `exclude_from_norm` is None, won't this cause an error here?

minhtuev (commented):
Adding some comments - we should add tests as well

mindmeld/tokenizer.py
ritvikshrivastava: It's set as: ``self.exclude_from_norm = exclude_from_norm or []``

mindmeld/components/_config.py
ritvikshrivastava: I just did that to preserve order, otherwise the function will need to move above this config. That works too.

mindmeld/components/_config.py
```diff
@@ -912,3 +918,102 @@ def get_nlp_config(app_path=None, config=None):
...
+
+def _get_default_regex(exclude_from_norm):
+    # List of regex's for matching and tokenizing when keep_special_chars=True
+    keep_special_regex_list = []
+
+    exception_chars = "\@\[\]\|\{\}'"  # noqa: W605
+
+    # fetches all currency symbols in unicode by iterating through the character set and
+    # selecting the currency symbols based on the unicode currency category 'Sc'
+    currency_symbols = u"".join(
```
vijay120, line 931: Is this a resource intensive process, ie how many iters happens here? If it is intensive, lets just make a list of current symbols and add it to constants.py

mindmeld/tokenizer.py
vijay120: Since this logic is happening twice, I think it makes sense just to add this is constants.py

mindmeld/components/_config.py
ritvikshrivastava: It's a single iteration through the loop, so I think this setup should be fine (I've moved it to constants.py though)? It's forward looking as it will allow for any changes to the unicode currency structure to be automatically included in MindMeld without having to update the list manually.

mindmeld/components/_config.py
ritvikshrivastava: Just realised that it uses a parameter that's coming from the tokenizer class, so can't change this.

mindmeld/tokenizer.py
ritvikshrivastava: Done.

mindmeld/components/_config.py
```diff
@@ -912,3 +918,102 @@ def get_nlp_config(app_path=None, config=None):
         pass
 
     return _get_default_nlp_config()
+
+
+def _get_default_regex(exclude_from_norm):
```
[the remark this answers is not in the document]
ritvikshrivastava: Done.

vijay120 (approved):

source/userguide/tokenizer.rst
```diff
@@ -0,0 +1,27 @@
+Working with the Tokenizer
+==========================
+
+MindMeld provides the ability to custom configure the tokenizer for your application.
```
minhtuev, line 4:
```suggestion
MindMeld provides the ability to customize and configure the default tokenizer for your application.
```

source/userguide/tokenizer.rst
```diff
@@ -0,0 +1,27 @@
...
+
+The settings for tokenizer can be defined in your application configuration file, ``config.py``. The configuration must be defined as a dictionary with the name :data:`TOKENIZER_CONFIG` to override the tokenizer's default settings. If no custom configuration is defined, the default is used.
+
+
+Anatomy of the tokenizer configuration
+""""""""""""""""""""""""""""""""""""""
```
minhtuev, line 10: Follow the header pattern in other documents:
```suggestion
Anatomy of the tokenizer configuration
---------------------------------------
```

source/userguide/tokenizer.rst
```diff
@@ -0,0 +1,27 @@
...
+        "allowed_patterns": ['\w+'],
+    }
+
+.. code:: python
+
+   TOKENIZER_CONFIG = {
+        "allowed_patterns": ['[a-zA-Z]'],
+    }
+
+will allow the system to capture alphanumeric and only alphabetical strings respectively.
```
minhtuev, line 27: maybe you should give two different different examples

ritvikshrivastava: Now?

minhtuev (approved):
LGTM, nice tests 👍

PR #211, opened 2020-08-22
Implemented progress bars for classifier training

vijay120:
This PR implements progress bars for MindMeld classifiers, solving issue https://github.com/cisco/mindmeld/issues/200. The current implementation makes a few assumptions:
1. Each classifier takes the same time to train
2. Doesn't take into account data loading, which can take awhile.

PR #218, opened 2020-08-31
Upgrading Click to 7.1

minhtuev:
Upgrading Click lib to 7.1 to fix Click issue


minhtuev:
Picking the Click dep change from Vijay's PR
https://github.com/cisco/mindmeld/pull/211/files

PR #226, opened 2020-09-10
Add py37 to workflows

jjjacksn:
MindMeld currently supports Python 3.7 but it is not tested in our Github workflow. This addresses that oversight.


minhtuev:
Did you manage to find the bugfix for the issue where the tests are hanging after completed?

serapio (approved):
it looks so easy now!

Issue #214, opened 2020-08-25
Classifying role to null when there is a single role under an intent


sameerg07:
I have an entity x which has two roles x1 and x2 . x1 is used in an intent i1 and x2 is used in an intent i2 . All the expressions in i1 and i2 have single roles mapped in their expressions for an entity. Now, when I do nlp.process(<expression which has an entity with a single role>), it classifies the role of that entity to null. 

I understand that there is a lazy evaluation happening to avoid role classification, but readability wise it confuses me atleast, that why is my role being shown as null? 

Sample response:
```
{
  "result": {
    "domain": "restaurant",
    "entities": [
      {
        "role": null,
        "span": {
          "end": 8,
          "start": 6
        },
        "text": "one",
        "type": "sys_number",
        "value": [
          {
            "value": 1
          }
        ]
      },
      {
        "role": null,
        "span": {
          "end": 29,
          "start": 10
        },
        "text": "paneer butter masala",
        "type": "dish",
        "value": []
      }
    ],
    "intent": "order",
    "text": "order one paneer butter masala"
  },
  "status": "success"
}
```

where my query is `order one paneer butter masala` which is basically `order {one|sys_number|no_items} {paneer butter masala|dish}`

and `sys_number` has two roles `no_ppl,no_items`

training data in intent **order** looks like:

```
order {one|sys_number|no_items} {paneer butter masala|dish}
order {3|sys_number|no_items} {taco salads|dish}
order {5|sys_number|no_items} {butter naan|dish}
order {two|sys_number|no_items} {virgin mojito|dish}
order {five|sys_number|no_items} {tomato soup|dish}
```

serapio:
Why do you need the role label if it can be determined from the intent label? This design adds work to maintaining the annotation, while dialogue manager logic could just use the existing intent and entity labels.

sameerg07:
You should have role labels as well on the dialogue manager. At any point
of time, as an end user, it's not clear why it isn't classifying even
though there is a role. And coming to dialogue manager, flexibility of role
classification is also important I feel

PR #230, opened 2020-09-18
Adding the ability to process and return a single role label

minhtuev:
Adding the ability to process and return a single role label if only one role is provided

https://github.com/cisco/mindmeld/issues/214

### DIFF
--- a/mindmeld/components/role_classifier.py
+++ b/mindmeld/components/role_classifier.py
@@ -208,6 +208,8 @@ class RoleClassifier(Classifier):
         if not self._model:
             logger.error("You must fit or load the model before running predict")
             return
+        if len(self.roles) == 1:
+            return list(self.roles)[0]
         if not isinstance(query, Query):
             query = self._resource_loader.query_factory.create_query(query)
         gazetteers = self._resource_loader.get_gazetteers()
@@ -232,6 +234,8 @@ class RoleClassifier(Classifier):
         if not self._model:
             logger.error("You must fit or load the model before running predict")
             return
+        if len(self.roles) == 1:
+            return [(self.list(self.roles)[0], 1.0)]
         if not isinstance(query, Query):
             query = self._resource_loader.query_factory.create_query(query)
         gazetteers = self._resource_loader.get_gazetteers()
@@ -313,7 +317,7 @@ class RoleClassifier(Classifier):
                     labels.append(entity.entity.role)
 
         unique_labels = set(labels)
-        if len(unique_labels) == 1:
+        if len(unique_labels) == 0:
             # No roles
             return (), ()
         if None in unique_labels:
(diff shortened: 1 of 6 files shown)


tests/home_assistant/domains/smart_home/set_thermostat/test.txt
```diff
@@ -1,96 +1,96 @@
...
-adjust the thermostat to {66 degrees|sys_temperature}
-adjust the thermostat to {73|sys_temperature}
-change the thermostat to {72|sys_temperature}
-temperature to {65 degrees|sys_temperature}
-thermostat to {72 degrees|sys_temperature}
-make it {75|sys_temperature} in {bedroom|location}
-change the thermostat setting to {78 degrees|sys_temperature} in the {bathroom|location}
-set the temp in the {living room|location} to {72 degrees|sys_temperature} at {8 o clock tonight|sys_time}
-heating to {77 degrees|sys_temperature} in the {living room|location}
+change the thermostat to {67 degrees|sys_temperature|room_temperature} in the {living room|location}
```
minhtuev, line 1: Adding a single role type here, `room_temperature`

vijay120 (approved):

mindmeld/components/role_classifier.py
```diff
@@ -313,7 +317,7 @@ def _get_queries_and_labels(self, queries=None, label_set=DEFAULT_TRAIN_SET_REGE
                     labels.append(entity.entity.role)
 
         unique_labels = set(labels)
-        if len(unique_labels) == 1:
+        if len(unique_labels) == 0:
```
serapio, line 320: what is the `0` case? I thought unlabeled roles count as `label: None`, and just aren't printed. so if there are any entities, there would be some role labels in this list, right?

serapio (approved):
just a missed annotation and a question

mindmeld/components/role_classifier.py
minhtuev: Let me check but I assume that currently `label: None` is actually a case of bad annotation (line 323). So `unique_labels` refers to the set of all valid Role labels:
```
if None in unique_labels:
            bad_examples = [e for i, e in enumerate(examples) if labels[i] is None]	            bad_examples = [e for i, e in enumerate(examples) if labels[i] is None]
```

serapio: Does it not distinguish between entities that have a role vs entities that don't? e.g. `{8:00|sys_time|start_time}` vs `{December|sys_time}`

minhtuev: image
I think you can test it again locally, but apparently the current role classifier does not distinguish between entities that have a role and entities that don't. Here in this query ("change my alarm from 6am to 8am since it is already 3am") all three entities are annotated with a role type even though only two of them were marked so in training and test.

Issue #221, opened 2020-09-01
Remove Support for Python 3.5


jjjacksn:
Python 3.5 reaches end of life on 13 Sept 2020. We should remove official support for it so that we can use more modern python language features. Mostly I'm interested in f-strings!

minhtuev:
After discussing with the team, I think we can add deprecated warnings into this release, and officially deprecate Py3.5 in the next release.

PR #235, opened 2020-09-22
Add deprecation warnings for Python 3.5

minhtuev:
Python 3.5 reached the end of life on 13 Sept 2020. MindMeld will deprecate official support for Python 3.5 in the next release, so we are adding the deprecation warnings to this release.

https://github.com/cisco/mindmeld/issues/221

### DIFF
--- a/mindmeld/cli.py
+++ b/mindmeld/cli.py
@@ -48,6 +48,15 @@ click.disable_unicode_literals_warning = True
 
 CONTEXT_SETTINGS = {"help_option_names": ["-h", "--help"], "auto_envvar_prefix": "MM"}
 
+# deprecation warning for python 3.5
+if sys.version[:3] == "3.5":
+    deprecation_msg = (
+        "DEPRECATION: Python 3.5 reached end of life on 13 Sept 2020. MindMeld will deprecate"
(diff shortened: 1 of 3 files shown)


mindmeld/cli.py
```diff
@@ -48,6 +48,15 @@
 
 CONTEXT_SETTINGS = {"help_option_names": ["-h", "--help"], "auto_envvar_prefix": "MM"}
 
+# deprecation warning for python 3.5
+if sys.version_info < (3, 6):
+    deprecation_msg = (
+        "DEPRECATION: Python 3.5 reached end of life on 13 Sept 2020. MindMeld will deprecate"
+        " official support for Python 3.5 in the next release. Please consider migrating"
+        " your application to Python 3.6 and above."
```
jjjacksn, line 56: Ideally we would support the latest python (3.8) and encourage them to use that. At least I'd like to recommend 3.7 BUT we don't have continuous integration set up for those so this is probable best as is (3.6).

jjjacksn (approved):

tmehlinger (approved):

vijay120 (approved):

PR #236, opened 2020-09-24
Expose ES score from QA calls

arushir:
Minor fix to expose the score that Elasticsearch provides for QA calls

### DIFF
--- a/mindmeld/components/question_answerer.py
+++ b/mindmeld/components/question_answerer.py
@@ -833,7 +833,14 @@ class Search:
             es_query = self._build_es_query(size=size)
 
             response = self.client.search(index=self.index, body=es_query)
-            results = [hit["_source"] for hit in response["hits"]["hits"]]
+
+            # construct results, removing embedding metadata and exposing score
+            results = []
+            for hit in response["hits"]["hits"]:
+                item = {key: val for (key, val) in hit["_source"].items()
+                        if not key.endswith(EMBEDDING_FIELD_STRING)}
+                item['_score'] = hit['_score']
+                results.append(item)
             return results
         except EsConnectionError as e:
             logger.error(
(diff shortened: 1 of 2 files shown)


minhtuev (commented):
Looks good to me - can you add a test to assert that the score is available in the results?

mindmeld/components/question_answerer.py
```diff
@@ -833,7 +833,14 @@ def execute(self, size=10):
...
 
             response = self.client.search(index=self.index, body=es_query)
-            results = [hit["_source"] for hit in response["hits"]["hits"]]
+
+            # construct results, removing embedding metadata and exposing score
+            results = []
+            for hit in response["hits"]["hits"]:
+                item = {key: val for (key, val) in hit["_source"].items()
+                        if not key.endswith(EMBEDDING_FIELD_STRING)}
+                item['_score'] = hit['_score']
```
minhtuev, line 842: Small nit: should we use `score` instead of `_score` since "_" tends to denote private fields and properties?

```suggestion
                item['_score'] = hit['_score']
```

tests/components/test_question_answerer.py
```diff
@@ -274,53 +276,45 @@ def test_advanced_search_validation(answerer):
 @pytest.mark.bert
 @pytest.mark.es7
 def test_embedder_search_bert(food_ordering_with_bert):
-    BERT_EMBEDDING_LEN = 768
     res = food_ordering_with_bert.get(
         index="menu_items_bert", query_type="embedder", name="pasta with tomato sauce"
     )
     assert len(res) > 0
```
minhtuev, line 282: why do we remove these assertions?

minhtuev (approved):
Some comments, otherwise LGTM!

mindmeld/components/question_answerer.py
arushir: I wanted it to include some special character to it's less likely to overwrite any fields of the item itself. E.g. if there was an index of soccer games there might be a `score` field already

tests/components/test_question_answerer.py
arushir: I removed the embedding from the item returned from the question answerer, since it seems cleaner and we don't return similar metadata for text features

vijay120 (approved):

PR #240, opened 2020-10-02
Bug/custom tokenizer

ritvikshrivastava:
Fixed issues pointed out by @vembar.

### DIFF
--- a/mindmeld/tokenizer.py
+++ b/mindmeld/tokenizer.py
@@ -106,11 +106,11 @@ class Tokenizer:
             "apostrophe": (" ", None),
         }
 
-        # Create compiled regex expressions
-
-        if "allowed_patterns" in self.config:
+        # Check if custom pattern is being used or MM defined
+        if "allowed_patterns" in self.config and (self.config["allowed_patterns"]):
             self._custom = True
 
+        # Create compiled regex expressions
         try:
             self.keep_special_compiled = re.compile(
                 "(%s)"
@@ -184,11 +184,17 @@ class Tokenizer:
         """
         # For each match, look-up corresponding value in dictionary
         try:
+
+            # Checks if replacement can be found in pre-defined match object (non-custom).
+            # If no key in match object, go to custom tokenizer handling in Exception.
             filtered = compiled.sub(self._one_xlat, text)
 
+            # If no key error and custom tokenizer was involved
+            # then the token has unwanted special characters. Remove them and return.
             if self._custom:
                 return self.compiled.sub(self._one_xlat, text)
 
+            # Return filtered list if non-custom tokenizer.
             return filtered
 
         except KeyError:
--- a/mindmeld/tokenizer.py
+++ b/mindmeld/tokenizer.py
@@ -110,14 +110,12 @@ class Tokenizer:
 
         if "allowed_patterns" in self.config:
             self._custom = True
-        else:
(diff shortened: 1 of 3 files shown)


mindmeld/tokenizer.py
```diff
@@ -106,11 +106,11 @@ def _init_regex(self):
             "apostrophe": (" ", None),
         }
 
-        # Create compiled regex expressions
-
-        if "allowed_patterns" in self.config:
+        # Check if custom pattern is being used or MM defined
+        if "allowed_patterns" in self.config and (self.config["allowed_patterns"]):
```
minhtuev, line 110:
```suggestion
        if self.config.get("allowed_patterns"):
```

tests/test_tokenizer.py
```diff
@@ -225,4 +225,4 @@ def test_custom_tokenizer(tokenizer, custom_tokenizer):
     custom_normalized = custom_tokenizer.normalize(raw, True)
 
     assert normalized == "is s o b gonna be on at 8 p m"
-    assert custom_normalized == "is s.o.b. ,, gonna be on at 8 p.m."
+    assert custom_normalized == "is s.o.b. gonna be on at 8 p.m."
```
minhtuev, line 228: should we add more tests around this issue?

minhtuev (approved):
Looks good to me, are you going to create a doc for the Default Tokenizer behavior in a separate PR?

tests/test_tokenizer.py
ritvikshrivastava: Done!

mindmeld/tokenizer.py
```diff
@@ -105,19 +106,26 @@ def _init_regex(self):
...
+                    ),
                 ),
-                re.UNICODE
+                re.UNICODE,
             )
         except sre_constants.error:
             logger.error(
                 "Custom regex compilation failed for the following patterns: %s",
-                ")|(".join(self.config["allowed_patterns"])
+                ")|(".join(self.config["allowed_patterns"]),
```
serapio, line 128: Is it possible for some pathway to add characters to `exclude_from_norm` that would then make the default_allowed_patterns a bad regex? It may be better to not assume this error came from the custom pattern. Then you could do the string join once at the start/before the try block, and reuse it here.

serapio (commented):
just a doubt

mindmeld/tokenizer.py
```diff
@@ -179,12 +187,24 @@ def multiple_replace(self, text, compiled):
...
+
+            # Return filtered list if non-custom tokenizer.
+            return filtered
+
         except KeyError:
             # In case of custom/app-specific tokenizer configuration
             logger.info("Using custom tokenizer configuration.")
             re_str = compiled.findall(text)
-            return "".join([e[0] if len(e) > 1 else e for e in re_str])
+            return "".join([list(filter(None, e))[0] for e in re_str])
```
vijay120, line 207: I think a note on the list->filter->none trick here would be good.

vijay120 (approved):

mindmeld/tokenizer.py
minhtuev: Out of curiosity, why don't we do this:
```suggestion
            return "".join(list(filter(None, [e for e in re_str])))
```

PR #248, opened 2020-10-08
Fixing sys candidate feature extractor bug

vembar:
While extracting the sys_candidate feature, we are currently using the raw entity spans to extract features for normalized tokens which causes an index mismatch if there are special characters in the text. For example, for the query "Schedule a follow-up at 7", for the token "at" we end up extracting `sys_candidate|type:sys_time|granularity:<X>|pos:-1` and `sys_candidate|type:sys_time|granularity:<X>|pos:0` whereas the right feature should be `sys_candidate|type:sys_time|granularity:<X>|pos:0` and `sys_candidate|type:sys_time|granularity:<X>|pos:1` 

The fix for this is to iterate on the normalized spans in the extractor so it matches the normalized token indices.


minhtuev (approved):
LGTM, go ahead after fixing tests.

serapio (approved):

vijay120 (approved):

Issue #244, opened 2020-10-06
Serializing autofill form


minhtuev:
Currently, the autofill form is stored inside a dictionary object, which is shared among different sessions. This does not allow multiple sessions to share the autofill flow. So the solution should allow the autofill form to serialize for every request.

abhayana24:
@minhtuev @ritvikshrivastava any update on this issue ?

ritvikshrivastava:
Resolved with #262.

PR #262, opened 2020-12-16
Serialize autofill

ritvikshrivastava:
Fixes async behavior for slot-filling. Propagates the partially filled entities through the request object.

### DIFF
--- a/mindmeld/app_manager.py
+++ b/mindmeld/app_manager.py
@@ -292,8 +292,7 @@ class ApplicationManager:
     def _post_dm(self, request, dm_response):
         # Append this item to the history, but don't recursively store history
         prev_request = DialogueResponder.to_json(dm_response)
-        if "history" in prev_request:
-            prev_request.pop("history")
+        prev_request.pop("history")
 
         # limit length of history
         new_history = (prev_request,) + request.history
--- a/mindmeld/app_manager.py
+++ b/mindmeld/app_manager.py
@@ -295,9 +295,9 @@ class ApplicationManager:
         if "history" in prev_request:
             prev_request.pop("history")
 
-            # limit length of history
-            new_history = (prev_request,) + request.history
-            dm_response.history = new_history[: self.max_history_len]
+        # limit length of history
+        new_history = (prev_request,) + request.history
+        dm_response.history = new_history[: self.max_history_len]
 
         # validate outgoing params
         dm_response.params.validate_param("allowed_intents")
--- a/mindmeld/app_manager.py
+++ b/mindmeld/app_manager.py
@@ -146,7 +146,7 @@ class ApplicationManager:
         # We reset the current turn's responder's params
         response = self.responder_class(
             frame=frame,
-            form=[],
+            form={},
             params=Params(),
             slots={},
             history=history,
@@ -200,7 +200,7 @@ class ApplicationManager:
         params = freeze_params(params)
         history = history or []
         frame = frame or {}
-        form = form or []
+        form = form or {}
         context = context or {}
 
         allowed_intents, nlp_params, dm_params = self._pre_nlp(params, verbose)
@@ -223,7 +223,7 @@ class ApplicationManager:
         return modified_dm_responder
 
     async def _parse_async(
-        self, text, params=None, context=None, frame=None, history=None, verbose=False
+        self, text, params=None, context=None, frame=None, form=None, history=None, verbose=False
     ):
         """
         Args:
@@ -256,7 +256,7 @@ class ApplicationManager:
         context = context or {}
         history = history or []
         frame = frame or {}
-        form = form or []
+        form = form or {}
 
         allowed_intents, nlp_params, dm_params = self._pre_nlp(params, verbose)
         # TODO: make an async nlp
@@ -292,11 +292,12 @@ class ApplicationManager:
     def _post_dm(self, request, dm_response):
         # Append this item to the history, but don't recursively store history
         prev_request = DialogueResponder.to_json(dm_response)
-        prev_request.pop("history")
+        if "history" in prev_request:
+            prev_request.pop("history")
 
-        # limit length of history
-        new_history = (prev_request,) + request.history
-        dm_response.history = new_history[: self.max_history_len]
+            # limit length of history
+            new_history = (prev_request,) + request.history
+            dm_response.history = new_history[: self.max_history_len]
 
         # validate outgoing params
         dm_response.params.validate_param("allowed_intents")
--- a/mindmeld/app_manager.py
+++ b/mindmeld/app_manager.py
@@ -132,12 +132,13 @@ class ApplicationManager:
         # TODO: make an async nlp
         self.nlp.load()
 
-    def _pre_dm(self, processed_query, context, params, frame, history):
+    def _pre_dm(self, processed_query, context, params, frame, form, history):
         # We pass in the previous turn's responder's params to the current request
         request = self.request_class(
             context=context,
             history=history,
             frame=frame,
+            form=form,
             params=params,
             **processed_query
         )
@@ -145,6 +146,7 @@ class ApplicationManager:
         # We reset the current turn's responder's params
         response = self.responder_class(
             frame=frame,
+            form=[],
             params=Params(),
             slots={},
             history=history,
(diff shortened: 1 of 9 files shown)


mindmeld/components/dialogue.py
```diff
@@ -920,6 +920,7 @@ def _end_slot_fill(self, request, responder, async_mode):
             history=request.history or [],
             frame=responder.frame or {},
             params=request.params,
+            form=None,
```
vijay120, line 923:
```suggestion
            form=[],
```

mindmeld/components/request.py
```diff
@@ -351,6 +351,9 @@ class Request:
     nbest_aligned_entities = attr.ib(
         default=attr.Factory(tuple), converter=tuple_elems_to_immutable_map
     )
+    form = attr.ib(
+        default=None,
```
vijay120, line 355: I thought this would be an empty list?

mindmeld/components/dialogue.py
```diff
@@ -972,14 +981,20 @@ def __call__(self, request, responder):
             request (Request): The request object.
             responder (DialogueResponder): The responder object.
         """
+
+        # If form iteration in request object, continue using that. 
+        # If None, set to original form.
+        if request.form:
+            self._local_entity_form = request.form or None
```
vijay120, line 988:
```suggestion
        self._local_entity_form = request.form or []
```

vijay120 (commented):
Some comments

ritvikshrivastava:
#244

mindmeld/app_manager.py
```diff
@@ -285,7 +292,8 @@ def _pre_nlp(self, params, verbose=False):
     def _post_dm(self, request, dm_response):
         # Append this item to the history, but don't recursively store history
         prev_request = DialogueResponder.to_json(dm_response)
-        prev_request.pop("history")
+        if "history" in prev_request:
```
vijay120, line 295: We can remove this line even if it fails, since it will fail in master as well

mindmeld/components/dialogue.py
```diff
@@ -935,13 +937,25 @@ async def _end_slot_fill_async(self, request, responder):
         return await self._handler(request, responder)
 
     def _prompt_slot(self, responder, nlr):
+        """Prompts user for missing slot.
+
+        Args:
+            responder (DialogueResponder): responder object.
+            nlr (str): natural language response to prompt for the missing slot.
+        """
+        self._local_form = copy.deepcopy(self._form)
```
vijay120, line 946: Is there anyway we can consolidate `_local_form ` and `_form ` into one data structure instead of having two?

mindmeld/components/dialogue.py
```diff
@@ -972,14 +988,22 @@ def __call__(self, request, responder):
             request (Request): The request object.
             responder (DialogueResponder): The responder object.
         """
+
+        # If form iteration in request object, continue using that.
+        # If None, set to original form.
+        if request.form:
```
vijay120, line 994: `if request.form and request.form["entities"]`

mindmeld/components/dialogue.py
```diff
@@ -972,14 +988,22 @@ def __call__(self, request, responder):
...
+        else:
+            self._local_entity_form = None
+
         if request.text.lower() in self._exit_keys:
             responder.reply(self._exit_response)
             self._exit_flow(responder)
             return
 
-        self._set_next_turn(request, responder)
+        responder = self._set_next_turn(request, responder)
```
vijay120, line 1004: Why do we assign back to responder here while in the past, we didnt?

mindmeld/components/dialogue.py
```diff
@@ -988,6 +1012,18 @@ def __call__(self, request, responder):
             # Fill the form with the entities in the first query
             self._initial_fill(request)
 
+        # convert json response to FormEntity objects (deserialize)
+        for i, slot in enumerate(self._local_entity_form):
+            if not isinstance(slot, FormEntity):
```
vijay120, line 1017: It looks like we assume slot is a dict here, why dont we just check for `if not isinstance(slot, dict):` instead?

mindmeld/components/dialogue.py
```diff
@@ -988,6 +1012,18 @@ def __call__(self, request, responder):
...
+                dict_slot = slot
```
vijay120, line 1018: Do we need to create a new assignment here?

mindmeld/components/dialogue.py
```diff
@@ -988,6 +1012,18 @@ def __call__(self, request, responder):
...
+                slot = FormEntity(entity=dict_slot["entity"])
```
vijay120, line 1019: Why dont we expand the constructor of `FormEntity ` to take in kwargs so that we dont have to do this dict assignment and setattr steps?

mindmeld/components/dialogue.py
```diff
@@ -1233,8 +1272,15 @@ def to_json(instance):
                 serialized_obj[attribute] = tuple(dict(item) for item in value)
             elif isinstance(value, immutables.Map):
                 serialized_obj[attribute] = dict(value)
+            elif value and isinstance(value, dict) and "entities" in value:
```
vijay120, line 1275: I think we should just provide a unique key for the serialized form entity, something like `{'type': 'form_filling', ...}`, so our checks are more explicit instead of relying on a derivative property

mindmeld/components/dialogue.py
```diff
@@ -1233,8 +1272,15 @@ def to_json(instance):
...
+                # Serialize slot-filling form
+                if value["entities"] and any(isinstance(i, FormEntity) for i in value["entities"]):
+                    value = dict(value)
```
vijay120, line 1278: Why do we explicitly do `dict` here we already checked that `isinstance(value, dict)` in the above conditional?

mindmeld/components/dialogue.py
```diff
@@ -1233,8 +1272,15 @@ def to_json(instance):
...
+                    value["entities"] = list(formentity.to_dict() for formentity in value["entities"])
```
vijay120, line 1279: `formentity ` -> `form_entity `

mindmeld/core.py
```diff
@@ -792,6 +793,18 @@ def __init__(
         if self.custom_eval and not callable(custom_eval):
             raise TypeError("Invalid custom validation function type.")
 
+    def to_dict(self):
+        """Converts the entity into a dictionary"""
+        base = {"entity": self.entity}
+        for field in ["role", "responses", "retry_response", "value", "default_eval", "hints", "custom_eval"]:
```
vijay120, line 799: You could use `self.__dict__` to iterate through the writable attributes

vijay120 (changes requested):
Some changes requested

mindmeld/components/dialogue.py
ritvikshrivastava: We don't need to do this, because the responder object is carried in the class, but just for code completeness I added this. Can revert this.

mindmeld/components/dialogue.py
ritvikshrivastava: It fails to write into the dictionary without this explicit conversion

mindmeld/components/dialogue.py
ritvikshrivastava: Ah yes. Changing it to just read `` elif attribute is 'form'``

mindmeld/components/dialogue.py
ritvikshrivastava: Reverted.

mindmeld/components/dialogue.py
ritvikshrivastava: Removed class dependence, made it into a local variable (response_form).

mindmeld/components/dialogue.py
ritvikshrivastava: Updated

mindmeld/core.py
ritvikshrivastava: Thanks!

mindmeld/app_manager.py
ritvikshrivastava: Removed.

vijay120 (approved):

PR #263, opened 2020-12-19
Added es resolution build flag

vijay120:
This PR adds a flag to only run the entity resolver once every build. This insures that there aren't too many index construction calls to ES if `nlp.build` needs to be run multiple times. The limitation of this design is that if the entity maps change after `nlp.build` is done, the updated maps will not be ingest unless the user reinitializes the `nlp` object and runs build again.

### DIFF
--- a/mindmeld/components/entity_resolver.py
+++ b/mindmeld/components/entity_resolver.py
@@ -75,6 +75,7 @@ class EntityResolver:
         self._er_config = get_classifier_config("entity_resolution", app_path=app_path)
         self._es_host = es_host
         self._es_config = {"client": es_client, "pid": os.getpid()}
+        self.ready = False
 
         if self._is_system_entity:
             canonical_entities = []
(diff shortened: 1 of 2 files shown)


ritvikshrivastava (approved):

mindmeld/components/entity_resolver.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         self._er_config = get_classifier_config("entity_resolution", app_path=app_path)
         self._es_host = es_host
         self._es_config = {"client": es_client, "pid": os.getpid()}
+        self._already_fit = False
```
serapio, line 78: is this equivalent to `ready` in the classifiers?

vijay120: Yes, its the same, except we actually use the flag here as a conditional, while we dont use `ready` for the models for a practical purpose. I will match the name.

ritvikshrivastava:
Adding a note for later review (discussed with Vijay).

The build flag resolves the file descriptor issue for the same NLP object, but having a toggle in MindMeld for disabling elasticsearch would allow multiple nlp objects to be run on the same system if elasticsearch is not necessary for the application.

serapio (approved):

ritvikshrivastava (approved):

PR #266, opened 2021-01-05
Feature/MultiLingual-Annotator

KKsharma99:
This PR Includes:
- Translator Abstract Class, GoogleTranslator class
- MultiLingualAnnotator Class
- Annotation Heuristics that don't require a Translation Service
- Update to create_annotator() in model/helpers.py
- Integration with Non-English Spacy NER.
- Tests and Documentation

======================================================================
The MultiLingual Annotator leverages a translation service (Google Translate API) and Duckling in order to label entities in various languages. ([Presentation](https://cisco-my.sharepoint.com/:p:/p/kunshar2/Edma83Y6YJRCokDw6ReN7PkBd8Op1AOD3NgeGNjdtfo3TA?e=gGBgmK))

**Option 1: Using the Google Translate API (Paid)**

**STEP 0**:
In your terminal, make this branched folder your root directory. Then, install this version of MindMeld and the extras-requirements by running the following:

```
pip install -e .
pip install -r extras-requirements.txt
```

If you are facing issues with the commands above, you could directly install google translate by running:

`pip install google-cloud-translate`

**STEP 1**: 
Export the path to your Google credentials file. For example:

`export GOOGLE_APPLICATION_CREDENTIALS="/Users/joesmith2/Documents/google_application_credentials.json"`

**STEP 2**:
Add the config into your config.py file. Be sure to define your own annotation rules in "annotate". To better understand annotation rules go [here](https://www.mindmeld.com/docs/userguide/auto_annotator.html#auto-annotator-configuration). 

NOTE: If a translator is not provided then Duckling annotation heuristics will be used. (MLA will selected the largest non-overlapping candidate spans to label.) However, for higher precision, use the "GoogleTranslator".

```
AUTO_ANNOTATOR_CONFIG = { 
    "annotator_class": "MultiLingualAnnotator", 
    "overwrite": True,
    "annotation_rules": [{"domains": ".*", "intents": ".*", "files": ".*", "entities": ".*",}], 
    "unannotate_supported_entities_only": False, 
    "unannotation_rules": None,
    "translator": "GoogleTranslator",
}
```

**STEP 3**:
To run the MultiLingual Annotator, simply enter the command below. Be sure to replace the app path.

`mindmeld annotate --app-path "es_wxa"`


serapio (commented):
This worked pretty well on the spanish data I tried it on, but it tag as many entities as I expected it to. Given the requirement `"entities": ".*"` I thought it would try annotating sys_org and sys_location besides the ones the app actually uses (sys_time, sys_duration etc), but I don't see any orgs or locations tagged. Is that not supposed to work?

setup.py
```diff
@@ -39,6 +39,7 @@
     "pyyaml>=5.1.1",
     "spacy>=2.3.0",
     "mypy>=0.782",
+    "google-cloud-translate>=3.0.1",
```
vijay120, line 42: Add this in the `extras_require` field in setup.py

mindmeld/auto_annotator.py
```diff
@@ -19,7 +19,7 @@
 from enum import Enum
 from tqdm import tqdm
 import spacy
-
+from google.cloud import translate_v2
```
vijay120, line 22: I think we can add a conditional here saying to use auto_annotator, you have to use `google.cloud` extra requirements field.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+class GoogleTranslator(Translator):
+    """Class for translation using the Google Translate API."""
+
+    def __init__(self):
+        """Initializes the translate_client."""
+        self.translate_client = self.get_translate_client()
+
+    def get_translate_client(self):
+        """Creates a translation client after finding the credential path."""
+        GoogleTranslator._check_credential_exists()
```
vijay120, line 890: Does this raise an exception if the creds are bad? If so, we should catch the exception and display proper error messages.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+
+    def get_translator(self):
+        """Creates a Translator object based on the translator class specified in the config.
+        If a translator is not specified, Duckling heuristics will be used.
+
+        Returns:
+            translator (Translator): Translator specified in the config, None if not specified.
+        """
+        if self.config.get("translator"):
+            translator_class_name = self.config.get("translator", "GoogleTranslator")
```
vijay120, line 954: Why do we do this: `self.config.get("translator", "GoogleTranslator")`? This would return `GoogleTranslator` if `translator ` key is none.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+            if translator_class_name == "GoogleTranslator":
```
vijay120, line 955: Let's use the factory pattern here (quick reference: https://medium.com/design-patterns-in-python/factory-pattern-in-python-2f7e1ca45d3e) to instantiating translators. It probably will on net add more code than what's here, but its more extensible and readable.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+            sentence (str): Sentence to detect entities.
+            entity_types (list): List of entity types to parse. If None, all
+                    possible entity types will be parsed.
+            language (str): Language code for the language of the given sentence
+        Returns: entities (list): List of entity dictionaries.
+        """
+        language = language or self.language
+        params = {"sentence": sentence, "language": language}
+        duckling_entities = (
+            self._parse_with_translator(**params)
```
vijay120, line 978: For readability, its better to set all the arg params explicitly.

setup.py
vijay120: set the key to be something like `language_annotator`

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+
+        Args:
+            sentence (str): Sentence to detect entities.
+            language (str): Language code for the language of the given sentence
+        Returns: entities (list): List of duckling candidates.
+        """
+        candidates = self.en_annotator.duckling.get_candidates_for_text(
+            sentence, language=language
+        )
+        en_sentence = self.translator.translate(sentence, target_language="en")
```
vijay120, line 999: Use `ENGLISH_LANGUAGE_CODE` in `config.py`

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+            if self.translator
+            else self._parse_without_translator(**params)
+        )
+        if entity_types:
+            duckling_entities = [
+                e for e in duckling_entities if e["dim"] in entity_types
+            ]
+        return MultiLingualAnnotator.convert_duckling_candidates(duckling_entities)
+
+    def _parse_with_translator(self, sentence, language=None):
```
vijay120, line 988: Shouldnt `language` be not None? Since if it is None, then `en` is chosen, which then becomes a no-op function.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+        candidates = self.en_annotator.duckling.get_candidates_for_text(
+            sentence, language=language
+        )
+        en_sentence = self.translator.translate(sentence, target_language="en")
+        en_entities = self.en_annotator.parse(en_sentence)
+
+        selected_candidates = []
+        for entity in en_entities:
+            i = 0
+            while i < len(candidates):
```
vijay120, line 1005: Why not `for candidate in candidates`? It seems like you break when a match is found?

vijay120: With the above, you could disregard `i`

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+        Args:
+            spans (list): List of tuples representing candidate spans (start_index, end_index + 1).
+            sentence (str): The original sentence parsed by duckling.
+        Returns:
+            filtered_spans (list): List of spans that do not end mid-token.
+        """
+        space_indices = [i for i in range(len(sentence)) if sentence[i] == " "]
+        return [
+            span
+            for span in spans
```
vijay120, line 1099: `for start_index, end_index in spans`

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+            if span[1] in space_indices or span[1] >= len(sentence) - 2
```
vijay120, line 1100: `if end_index in space_indices or end_index >= len(sentence) - 2`

vijay120: I don't understand `end_index >= len(sentence) - 2`. Also, if a space is in-between a span, shouldnt you reject the span? ie even span index from start_index to end_index needs to be checked against space_indices.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+    @property
+    def supported_entity_types(self):  # pylint: disable=W0236
+        """
+        Returns:
+            supported_entity_types (list): List of supported entity types.
+        """
+        return SPACY_ANNOTATOR_SUPPORTED_ENTITIES
+
+    @staticmethod
+    def _has_overlap(span_one, span_two):
```
vijay120, line 1073: I think the `_has_overlap `, `_is_larger ` should be part of the `Span` class. You can implement the primitive functions for `<` ops there.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+
+    def _parse_with_translator(self, sentence, language=None):
+        """Parse helper function to be used if a translator is present.
+
+        Args:
+            sentence (str): Sentence to detect entities.
+            language (str): Language code for the language of the given sentence
+        Returns: entities (list): List of duckling candidates.
+        """
+        candidates = self.en_annotator.duckling.get_candidates_for_text(
```
vijay120, line 996: I think the candidates returned from `get_candidates_for_text ` should be a list of `Spans` instead of `tuples`. That way, the functions like `_has_overlap ` and `_is_larger` can be neatly placed in the `Span` class.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+                    if not (
+                        candidate["dim"] == "amount-of-money"
+                        and candidate["value"]["unit"] == "unknown"
+                    ):
+                        selected_candidates.append(candidate)
+                        break
+        return selected_candidates
+
+    @staticmethod
+    def convert_duckling_candidates(duckling_candidates):
```
vijay120, line 1046: This should be in the duckling class

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+            sentence, language=language
+        )
+        final_spans = MultiLingualAnnotator._get_final_spans(candidates, sentence)
+        selected_candidates = []
+        for span in final_spans:
+            for candidate in candidates:
+                if span == (candidate["start"], candidate["end"]):
+                    if not (
+                        candidate["dim"] == "amount-of-money"
+                        and candidate["value"]["unit"] == "unknown"
```
vijay120, line 1039: I am not convinced that this heuristic always holds true. Maybe duckling returns these values for a reason

mindmeld/auto_annotator.py
```diff
@@ -835,5 +839,300 @@ def valid_entity_check(self, entity):
...
+        """Finds the set of the largest non-overlapping candidates.
+
+        Args:
+            spans (list): List of tuples representing candidate spans (start_index, end_index + 1).
+        """
+        selected_spans = []
+        for span in spans:
+            no_overlaps = True
+            for i, selected_span in enumerate(selected_spans):
+                if MultiLingualAnnotator._has_overlap(span, selected_span):
```
vijay120, line 1114:
```
no_overlaps = MultiLingualAnnotator._has_overlap(span, selected_span)
if not no_overlaps:
    continue
if MultiLingualAnnotator._has_overlap(span, selected_span):
    selected_spans[i] = span
```

vijay120: Better indentation ^

vijay120 (changes requested):
A few comments

KKsharma99:
> This worked pretty well on the spanish data I tried it on, but it tag as many entities as I expected it to. Given the requirement `"entities": ".*"` I thought it would try annotating sys_org and sys_location besides the ones the app actually uses (sys_time, sys_duration etc), but I don't see any orgs or locations tagged. Is that not supposed to work?

@serapio Great to hear that! As for labelling non-spacy entities such as (sys_org, sys_location), that component is still in development. I'll be adding that in next

mindmeld/auto_annotator.py
KKsharma99: It's okay to have a a space in-between a span (ex: "dos horas"). Removed the filter function since it is not always valid. There are many cases of punctuation appearing right after the end of a span, here the spans should not be filtered out. (ex: "{en la mañana}?" or "{a las 2 PM|sys_time},")

mindmeld/auto_annotator.py
KKsharma99: If translator is None, NoOpTranslator() is now selected as the default

mindmeld/auto_annotator.py
KKsharma99: Refactored this function, using cleaner logic:

```
        for span in spans:
            has_overlaps = [
                MultiLingualAnnotator._has_overlap(span, selected_span)
                for selected_span in selected_spans
            ]
            if not any(has_overlaps):
                selected_spans.append(span)
```

mindmeld/auto_annotator.py
KKsharma99: Good catch, switched to `for candidate in candidates`.

mindmeld/auto_annotator.py
KKsharma99: Moved to the system_entity_recognizer.py as duckling_item_to_entity_dict(). Placed in parallel to related functions: duckling_item_to_query_entity(), duckling_item_to_entity().

setup.py
KKsharma99: Added to the `extras_require` field in setup.py with a key of `language_annotator` as suggested.

Also updated extras-requirements.txt to:
```
-e .[bot]
-e .[tensorflow]
-e .[bert]
-e .[examples]
-e .[language_annotator]

```

mindmeld/auto_annotator.py
KKsharma99: Added `__gt__, __ge__, __lt__, __le__`

mindmeld/auto_annotator.py
KKsharma99: Added `has_overlap(object)` to the span class. is_larger() is no longer needed since functions like `__gt__` are present

mindmeld/auto_annotator.py
KKsharma99: This is helpful for cases such as "I have 1 book". Duckling may recognize 1 as a sys_amount-of-money but it should be sys_number. In most cases, when sys_amount-of-money is detected but the unit is "unknown" it is a misclassified

mindmeld/auto_annotator.py
KKsharma99: Added. On a related note, None is no longer acceptable for 'translator'. This prevents the user from accidentally using a NoOpTranslator when they mean to use a GoogleTranslator.

mindmeld/auto_annotator.py
KKsharma99: Moved this out as a separate function.

mindmeld/auto_annotator.py
```diff
@@ -401,27 +416,35 @@ def parse(self, sentence, **kwargs):
...
-
-        self.model = self.config.get("spacy_model", EN_CORE_WEB_LG)
-        self.nlp = SpacyAnnotator._load_model(self.model)
-        self.duckling = DucklingRecognizer.get_instance()
+        self.language = language or self.language
+        self.locale = locale or self.locale
+        self.spacy_model_size = model_size or self.config.get("spacy_model_size", "lg")
+        self.model_name = self._get_spacy_model_name()
+        self.nlp = SpacyAnnotator._load_model(self.model_name)
         self.ANNOTATOR_TO_DUCKLING_ENTITY_MAPPINGS = {
```
vijay120, line 448: Why do we have this attr when its already defined outside of scope?

mindmeld/auto_annotator.py
```diff
@@ -467,10 +502,26 @@ def supported_entity_types(self):  # pylint: disable=W0236
         Returns:
             supported_entity_types (list): List of supported entity types.
         """
-        return SPACY_ANNOTATOR_SUPPORTED_ENTITIES
+        spacy_supported_entities = [e.lower() for e in self.nlp.get_pipe("ner").labels]
+        supported_entities = set()
+        for entity in spacy_supported_entities:
+            entity = "person" if entity == "per" else entity
+            if entity == "misc":
```
vijay120, line 509: I think more documentation here would be helpful on why we skip misc

mindmeld/auto_annotator.py
KKsharma99: This attr is currently defined in the SpacyAnnotator's init function since it is only used by the SpacyAnnotator as of now. Should I move this out to constants.py instead?

mindmeld/auto_annotator.py
```diff
@@ -835,5 +895,282 @@ def valid_entity_check(self, entity):
         return Entity.is_system_entity(entity) or entity in self.supported_entity_types
 
 
+class NoTranslationDucklingAnnotator(Annotator):
+    """Custom Annotator class used to generate annotations."""
+
+    def __init__(self, app_path, config=None):
```
vijay120, line 901: Its not very clear to me why the spacy annotator needs the language and locale in the constructor but the other annotator classes do not.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +895,282 @@ def valid_entity_check(self, entity):
...
+            sentence (str): Sentence to detect entities.
+            entity_types (list): List of entity types to parse. If None, all
+                    possible entity types will be parsed.
+            language (str, optional): Language as specified using a 639-1/2 code.
+            locale (str, optional): The locale representing the ISO 639-1 language code and \
+                ISO3166 alpha 2 country code separated by an underscore character.
+        Returns: entities (list): List of entity dictionaries.
+        """
+        locale = locale or self.locale
+        language = language or self.language
```
vijay120, line 1038: We should assert here that the language is not english

mindmeld/auto_annotator.py
```diff
@@ -835,5 +895,282 @@ def valid_entity_check(self, entity):
...
+        )
+        en_sentence = self.translator.translate(  # pylint: disable=E1128
+            sentence, target_language=ENGLISH_LANGUAGE_CODE
+        )
+        en_entities = self.en_annotator.parse(en_sentence, entity_types=entity_types)
+        final_candidates = []
+        for entity in en_entities:
+            value_matched_candidates = []
+            for candidate in candidates:
+                if entity["dim"] == candidate["entity_type"]:
```
vijay120, line 1051: To reduce indents, do:
```
if entity["dim"] != candidate["entity_type"]:
   continue
```

mindmeld/auto_annotator.py
```diff
@@ -835,5 +895,282 @@ def valid_entity_check(self, entity):
...
+                    if entity["value"] == candidate["value"]:
+                        value_matched_candidates.append(candidate)
+                    if (
```
vijay120, line 1054: Instead of this complex condition, why do we do this:
```
if entity["value"] == candidate["value"]:
    value_matched_candidates.append(candidate)
    continue
```

mindmeld/auto_annotator.py
```diff
@@ -835,5 +895,282 @@ def valid_entity_check(self, entity):
...
+        )
+        return merged_entities
+
+    @property
+    def supported_entity_types(self):  # pylint: disable=W0236
+        """
+        Returns:
+            supported_entity_types (list): List of supported entity types.
+        """
+        return SPACY_ANNOTATOR_SUPPORTED_ENTITIES
```
vijay120, line 1151: This probably needs to also be conditioned on language, no? Especially since this annotator depends on duckling candidates.

mindmeld/components/translators.py
```diff
@@ -0,0 +1,155 @@
...
+        Returns:
+            credential_path (str): Path to google application credentials.
+        """
+        try:
+            return os.environ["GOOGLE_APPLICATION_CREDENTIALS"]
+        except KeyError:
+            raise KeyError(
+                "Google credential path not found. Export 'GOOGLE_CREDENTIAL_PATH' as"
+                " an environment variable."
+            )
```
vijay120, line 112:
```suggestion
        except KeyError as error:
            raise KeyError(
                "Google credential path not found. Export 'GOOGLE_CREDENTIAL_PATH' as"
                " an environment variable."
            ) from error
```

vijay120 (commented):
Looks good so far!

mindmeld/auto_annotator.py
KKsharma99: Consistent now

mindmeld/auto_annotator.py
KKsharma99: Added

mindmeld/auto_annotator.py
KKsharma99: Added

mindmeld/auto_annotator.py
KKsharma99: Yes, that's correct. Planning to return the union of the entities supported by Spacy and Duckling. There isn't any documentation that includes a list of system entities supported by language for MindMeld's version of Duckling. The closest data available is in "allDimensions" in Duckling/Dimensions files. For example, Duckling/Dimensions/BG.hs:

```
  | ( allDimensions
  | ) where
  |  
  | import Duckling.Dimensions.Types
  |  
  | allDimensions :: [Some Dimension]
  | allDimensions =
  | [ This Distance
  | , This Duration
  | , This Numeral
  | , This Ordinal
  | ]
```

Would need some way to collect this data and then map dimensions to a list of entities. Open to suggestions for tackling this

KKsharma99: Added a dictionary that maps language codes to the supported entities. Added code to determine the supported entities for each class, this list is dynamically created based on the initialized language.

mindmeld/auto_annotator.py
```diff
@@ -50,39 +59,57 @@ class Annotator(ABC):
...
-        config = config or self.config
         all_file_paths = self._resource_loader.get_all_file_paths()
         file_entities_map = {path: [] for path in all_file_paths}
 
         if action == AnnotatorAction.ANNOTATE:
-            rules = config[AnnotatorAction.ANNOTATE.value]
+            rules = self.annotation_rules
         elif action == AnnotatorAction.UNANNOTATE:
-            rules = config[AnnotatorAction.UNANNOTATE.value]
+            rules = self.unannotation_rules
```
vijay120, line 112: Is there an else case or will that never happen? If its the latter, one can do this:
```
rules = self.annotation_rules if action == AnnotatorAction.ANNOTATE else self.unannotation_rules
```

mindmeld/auto_annotator.py
```diff
@@ -152,66 +179,37 @@ def valid_entity_check(self, entity):
...
             )
             return
         file_entities_map = self._get_file_entities_map(
-            action=AnnotatorAction.UNANNOTATE, config=config
-        )
-        self._modify_queries(
-            file_entities_map, action=AnnotatorAction.UNANNOTATE, config=config
+            action=AnnotatorAction.UNANNOTATE
         )
+        self._modify_queries(file_entities_map, action=AnnotatorAction.UNANNOTATE)
```
vijay120, line 204: You can create a separate helper function as follows:
```python
def update_query_with_action(self, action=AnnotatorAction.UNANNOTATE);
     file_entities_map = self._get_file_entities_map(action)
    self._modify_queries(file_entities_map, action)
```
and then call that from the `annotate ` and `unannotate ` methods

mindmeld/auto_annotator.py
```diff
@@ -362,32 +353,29 @@ def _resolve_conflicts(self, current_entities, annotated_entities, config):
...
-            config (dict): Config to use instead of the class config.
         """
-        config = config or self.config
-        remove_supported_only = config["unannotate_supported_entities_only"]
         keep_entities = []
         for query_entity in processed_query.entities:
             if remove_entities == ["*"]:
                 is_supported_entity = self.valid_entity_check(query_entity.entity.type)
-                if remove_supported_only and not is_supported_entity:
+                if self.unannotate_supported_entities_only and not is_supported_entity:
```
vijay120, line 370: `self.unannotate_supported_entities_only and not is_supported_entity and keep_entities.append(query_entity)`

mindmeld/auto_annotator.py
```diff
@@ -401,59 +389,85 @@ def parse(self, sentence, **kwargs):
...
+        super().__init__(
+            app_path,
+            annotation_rules=annotation_rules,
+            language=language,
+            locale=locale,
+            overwrite=overwrite,
+            unannotate_supported_entities_only=unannotate_supported_entities_only,
+            unannotation_rules=unannotation_rules,
+        )
+        self.spacy_model_size = spacy_model_size
```
vijay120, line 434: This parameter should be validated:
```python
if spacy_model_size not in {'lg', 'md', 'sm'}:
    raise ValueError(...)
```

mindmeld/auto_annotator.py
```diff
@@ -463,14 +477,53 @@ def _load_model(model):
...
+        filtered_entities = []
+        for entity in entities:
+            if entity in SPACY_ENTITIES_THAT_REQUIRE_DUCKLING:
+                if (
+                    self.language in DUCKLING_TO_SYS_ENTITY_MAPPINGS
+                    and entity in DUCKLING_TO_SYS_ENTITY_MAPPINGS[self.language]
+                ):
+                    filtered_entities.append(entity)
+            else:
+                filtered_entities.append(entity)
```
vijay120, line 522: I dont understand the else case, should we filter out entities that duckling doesnt support?

mindmeld/auto_annotator.py
```diff
@@ -764,33 +825,68 @@ def _resolve_person(self, entity):
...
         """
-        entity["dim"] = self.ANNOTATOR_TO_DUCKLING_ENTITY_MAPPINGS[entity["dim"]]
+        entity["dim"] = ANNOTATOR_TO_SYS_ENTITY_MAPPINGS[entity["dim"]]
 
-        if len(entity["body"]) >= 2 and entity["body"][-2:] == "'s":
-            entity["value"] = {"value": entity["body"][:-2]}
-            entity["body"] = entity["body"][:-2]
-            entity["end"] -= 2
+        if self.language == ENGLISH_LANGUAGE_CODE:
+            if len(entity["body"]) >= 2 and entity["body"][-2:] == "'s":
```
vijay120, line 831: These two conditionals can be combined into one.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +931,373 @@ def valid_entity_check(self, entity):
...
+        """
+        super().__init__(
+            app_path,
+            annotation_rules=annotation_rules,
+            language=language,
+            locale=locale,
+            overwrite=overwrite,
+            unannotate_supported_entities_only=unannotate_supported_entities_only,
+            unannotation_rules=unannotation_rules,
+        )
```
vijay120, line 967: You could do `super().__init__(*args, **kwargs)`

mindmeld/auto_annotator.py
```diff
@@ -401,59 +389,85 @@ def parse(self, sentence, **kwargs):
...
-        super().__init__(app_path=app_path, config=config)
+        super().__init__(
+            app_path,
+            annotation_rules=annotation_rules,
+            language=language,
+            locale=locale,
+            overwrite=overwrite,
+            unannotate_supported_entities_only=unannotate_supported_entities_only,
+            unannotation_rules=unannotation_rules,
+        )
```
vijay120, line 433: You could do `super.__init__(*args, **kwargs)` after removing `spacy_model_size ` from it.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +931,373 @@ def valid_entity_check(self, entity):
...
+        for span in spans:
+            has_overlaps = [
+                span.has_overlap(selected_span) for selected_span in selected_spans
+            ]
+            if not any(has_overlaps):
+                selected_spans.append(span)
+        return selected_spans
+
+    @staticmethod
+    def _filter_duckling_candidates(candidates):
```
vijay120, line 1039: I think this func can be named better: `_filter_out_bad_duckling_candidates`

mindmeld/auto_annotator.py
```diff
@@ -835,5 +931,373 @@ def valid_entity_check(self, entity):
...
+        final_candidates = []
+        for span in final_spans:
+            for candidate in filtered_candidates:
+                if span == Span(candidate["start"], candidate["end"] - 1):
+                    final_candidates.append(candidate)
+                    break
+        if entity_types:
+            final_candidates = [
+                e for e in final_candidates if e["entity_type"] in entity_types
+            ]
```
vijay120, line 1006:
```python
            final_candidates = [
                e for e in final_candidates if e["entity_type"] in entity_types
            ] if entity_types
```

mindmeld/auto_annotator.py
```diff
@@ -835,5 +931,373 @@ def valid_entity_check(self, entity):
...
+            en_annotator (SpacyAnnotator): A Spacy Annotator with language set to English ("en").
+            translator (str): A translator to use such as 'GoogleTranslator' or 'NoOpTranslator'.
+            language (str, optional): Language as specified using a 639-1/2 code.
+            locale (str, optional): The locale representing the ISO 639-1 language code and \
+                ISO3166 alpha 2 country code separated by an underscore character.
+            overwrite (bool): Whether to overwrite existing annotations with conflicting spans.
+            unannotate_supported_entities_only (bool): Only allow removal of supported entities.
+            unannotation_rules (list): List of Annotation rules.
+        """
+        super().__init__(
```
vijay120, line 1098: again, do the `*args and **kwargs` trick here to reduce verbosity. You can also pass in "extra" args that dont need to be used in `Annotator ` class so that you dont have to pop out args before calling super.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +931,373 @@ def valid_entity_check(self, entity):
...
+    @staticmethod
+    def _resolve_conflicts_entity_dicts(base_entities, other_entities):
+        non_overlapping_other_entities = []
+        for o_entity in other_entities:
+            no_overlaps = [
+                MultiLingualAnnotator._no_overlap_entity_dicts(o_entity, b_entity)
+                for b_entity in base_entities
+            ]
+            if all(no_overlaps):
+                non_overlapping_other_entities.append(o_entity)
```
vijay120, line 1295: `all(no_overlaps) and non_overlapping_other_entities.append(o_entity)`

mindmeld/auto_annotator.py
KKsharma99: Here we are ensuring that entities which require duckling are indeed supported by duckling for the given language. If the entity does not need duckling to be resolved (custom entities), then we can keep that entity.

mindmeld/auto_annotator.py
KKsharma99: language and spacy_model_size are currently validated in _get_spacy_model_name(self), should we move that code into the __init__()?

mindmeld/auto_annotator.py
```diff
@@ -463,14 +477,53 @@ def _load_model(model):
...
         Returns:
             supported_entity_types (list): List of supported entity types.
         """
-        return SPACY_ANNOTATOR_SUPPORTED_ENTITIES
+        spacy_supported_entities = [e.lower() for e in self.nlp.get_pipe("ner").labels]
+        supported_entities = set()
+        for entity in spacy_supported_entities:
+            if entity == "misc":
+                continue
+            if entity in ["time", "date", "datetime"]:
```
vembar, line 493: Is this not in the ANNOTATOR_TO_SYS_ENTITY_MAPPINGS dict because the same entity can map to different system entities?

mindmeld/auto_annotator.py
vembar: What is the difference between `ANNOTATOR_TO_SYS_ENTITY_MAPPINGS` and `SPACY_ENTITIES_THAT_REQUIRE_DUCKLING`?

source/userguide/auto_annotator.rst
```diff
@@ -235,19 +260,72 @@ Some of these entities are resolvable by Duckling.
...
++---------------------+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
+| sys_interval        | y  | y  | y  | y  | y  | y  | y  | n  | y  | y  | y  | y  | n  | y  | y  |
++---------------------+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
+| sys_language        | y  | n  | n  | n  | n  | n  | n  | n  | n  | y  | n  | n  | y  | y  | y  |
++---------------------+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
+| sys_law             | y  | n  | n  | n  | n  | n  | n  | n  | n  | n  | n  | n  | y  | y  | y  |
++---------------------+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
+| sys_loc             | y  | y  | y  | y  | y  | y  | y  | y  | y  | y  | n  | y  | y  | y  | y  |
++---------------------+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
+| sys_nrp             | y  | n  | n  | n  | n  | n  | n  | n  | n  | y  | n  | n  | y  | y  | y  |
```
vembar, line 289: Noticed this being referred to as sys_norp in a couple of places.

mindmeld/auto_annotator.py
```diff
@@ -401,59 +389,85 @@ def parse(self, sentence, **kwargs):
...
-    """ (English) Annotator class that uses spacy to generate annotations.
-    Supported entities include: "sys_time", "sys_interval", "sys_duration", "sys_number",
-    "sys_amount-of-money", "sys_distance", "sys_weight", "sys_ordinal", "sys_quantity",
-    "sys_percent", "sys_org", "sys_loc", "sys_person", "sys_gpe", "sys_norp", "sys_fac",
-    "sys_product", "sys_event", "sys_law", "sys_langauge", "sys_work-of-art", "sys_other-quantity",
+    """Annotator class that uses spacy to generate annotations.
+    Depending on the language, supported entities can include: "sys_time", "sys_interval",
+    "sys_duration", "sys_number", "sys_amount-of-money", "sys_distance", "sys_weight",
+    "sys_ordinal", "sys_quantity", "sys_percent", "sys_org", "sys_loc", "sys_person",
+    "sys_gpe", "sys_norp", "sys_fac", "sys_product", "sys_event", "sys_law", "sys_langauge",
```
vembar, line 396: Nit: change to sys_nrp

source/userguide/auto_annotator.rst
vembar: Left a comment in one of the code files. The other reference is in the previous table for "supported entities"

mindmeld/system_entity_recognizer.py
```diff
@@ -698,6 +698,23 @@ def duckling_item_to_query_entity(query, item, offset=0):
...
+def duckling_item_to_entity_dict(item):
+    """Converts a duckling candidate into a an entity dictionary to be used by Annotators.
+    Args:
+        item (dict): The duckling item
+
+    Returns:
+        entity (dict): Converted entity
+    """
+    return {
+        "body": item["body"],
```
vembar, line 710: Should this be renamed to "text" to keep it consistent with our existing schema?

source/userguide/auto_annotator.rst
```diff
@@ -304,23 +382,115 @@ For example, we can restrict the output of the previous example by doing the fol
...
+
+The :class:`MultiLingualAnnotator` will use the language and locale specified in the :attr:`LANGUAGE_CONFIG` (:attr:`config.py`) if it used through the command-line.
+
+.. code-block:: python
+	
+	LANGUAGE_CONFIG = {'language': 'es'}
+
+Many Spacy non-English NER models have limited entity support. To overcome this, the :class:`MultiLingualAnnotator` translates the sentence to English and detects entities
+using the English NER model. The English detected entities are compared against duckling candidates for the non-English sentence. Duckling candidates with a match between the type and value of the entity or the translated body text
+are selected. If a translation service is not available, the :class:`MultiLingualAnnotator` selects the duckling candidates with the largest non-overlapping spans. The sections below describe the steps to setup the annotator depending on whether a translation service is being used.
```
vembar, line 398: Seems to convey we only return the entities that overlap with the english NER and non-english duckling models. Consider adding another sentence in the end to clarify that the final list includes these overlapping entities + non-English spacy entites?

```
Duckling candidates with a match between the type and value of the entity or the translated body text
are selected and returned as valid entities in addition to the ones detected by the non-English NER models. 
```

vembar: Many Spacy non-English NER models have limited entity support. To overcome this, the :class:`MultiLingualAnnotator` translates the sentence to English and detects entities
using the English NER model.
```suggestion
Many Spacy non-English NER models have limited entity support. To overcome this, in addition to the entities detected by non-English NER models, the :class:`MultiLingualAnnotator` translates the sentence ...
```

source/userguide/auto_annotator.rst
```diff
@@ -507,6 +723,6 @@ Getting Custom Parameters from the Config
 		... 
 	}
 
-:class:`SpacyAnnotator` checks if :attr:`spacy_model` exists in the config, and if it doesn't, it will use the default value of "en_core_web_lg".
+If a :class:`SpacyAnnotator` is created using the command-line with will use the value for :attr:`spacy_model_size` that exists in the config during instantiation.
```
vembar, line 726: Nit: 
```suggestion
If a :class:`SpacyAnnotator` is created using the command-line, it will use the value for :attr:`spacy_model_size` that exists in the config during instantiation.
```

mindmeld/auto_annotator.py
KKsharma99: Done

source/userguide/auto_annotator.rst
KKsharma99: Good catch, this should be "sys_norp" as to align with Spacy's labelling scheme.

mindmeld/auto_annotator.py
KKsharma99: Moved "file_entities_map = self._get_file_entities_map(action)" into the "self._modify_queries()" helper function since this is shared code between the only two callers of this function.

source/userguide/auto_annotator.rst
KKsharma99: Done

mindmeld/auto_annotator.py
KKsharma99: Moved to init()

mindmeld/auto_annotator.py
KKsharma99: Done

mindmeld/auto_annotator.py
KKsharma99: Good point, only using `SPACY_SYS_ENTITIES_NOT_IN_DUCKLING` now. 

`ANNOTATOR_TO_SYS_ENTITY_MAPPINGS` maps the labels that Spacy generates across all languages to MindMeld sys_entities. These sys_entities can include both entities that are resolvable by duckling and those that aren't

mindmeld/auto_annotator.py
```diff
@@ -463,14 +456,53 @@ def _load_model(model):
...
+        supported_entities = set()
+        for entity in spacy_supported_entities:
+            if entity == "misc":
+                continue
+            if entity in ["time", "date", "datetime"]:
+                supported_entities.update(["sys_time", "sys_duration", "sys_interval"])
+            elif entity in ANNOTATOR_TO_SYS_ENTITY_MAPPINGS:
+                supported_entities.add(ANNOTATOR_TO_SYS_ENTITY_MAPPINGS[entity])
+            else:
+                supported_entities.add("sys_" + entity)
```
vijay120, line 477: Nit: `f"sys_{entity}"`

mindmeld/auto_annotator.py
```diff
@@ -764,33 +804,59 @@ def _resolve_person(self, entity):
...
             entity (dict): A resolved entity dict.
         """
-        entity["dim"] = self.ANNOTATOR_TO_DUCKLING_ENTITY_MAPPINGS[entity["dim"]]
+        entity["dim"] = ANNOTATOR_TO_SYS_ENTITY_MAPPINGS[entity["dim"]]
 
-        if len(entity["body"]) >= 2 and entity["body"][-2:] == "'s":
+        if (
+            self.language == ENGLISH_LANGUAGE_CODE
+            and len(entity["body"]) >= 2
+            and entity["body"][-2:] == "'s"
```
vijay120, line 812: Nit: This conditional can be packaged in a boolean function with a descriptive name like `is_plural_entity` or something more specific.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +901,320 @@ def valid_entity_check(self, entity):
...
+            for candidate in filtered_candidates:
+                if span == Span(candidate["start"], candidate["end"] - 1):
+                    final_candidates.append(candidate)
+                    break
+        if entity_types:
+            final_candidates = [
+                e for e in final_candidates if e["entity_type"] in entity_types
+            ]
+        return [
+            duckling_item_to_entity_dict(candidate) for candidate in final_candidates
```
vijay120, line 961: return a list of entity objects as discussed

mindmeld/auto_annotator.py
```diff
@@ -835,5 +901,320 @@ def valid_entity_check(self, entity):
...
+            overwrite (bool): Whether to overwrite existing annotations with conflicting spans.
+            unannotate_supported_entities_only (bool): Only allow removal of supported entities.
+            unannotation_rules (list): List of Annotation rules.
+        """
+        super().__init__(*args, **kwargs)
+        assert (
+            self.language != ENGLISH_LANGUAGE_CODE
+        ), "The 'language' for a TranslationDucklingAnnotator cannot be set to English."
+        translator = kwargs.get("translator")
+        self.translator = TranslatorFactory().get_translator(translator)
```
vijay120, line 1045: `get_translator ` expects a positional arg that cannot be None but `translator ` could be None. Let's handle this gracefully.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +901,320 @@ def valid_entity_check(self, entity):
...
+        )
+        en_sentence = self.translator.translate(  # pylint: disable=E1128
+            sentence, target_language=ENGLISH_LANGUAGE_CODE
+        )
+        en_entities = self.en_annotator.parse(en_sentence, entity_types=entity_types)
+        final_candidates = []
+        for entity in en_entities:
+            value_matched_candidates = []
+            for candidate in candidates:
+                if entity["dim"] != candidate["entity_type"]:
```
vijay120, line 1075: I would recommend adding comments for these rules to future readability. Its highly specific.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +901,320 @@ def valid_entity_check(self, entity):
...
+            unannotation_rules (list): List of Annotation rules.
+        """
+        super().__init__(*args, **kwargs)
+        self.translator = kwargs.get("translator", "NoOpTranslator")
+        self.en_annotator = SpacyAnnotator(
+            app_path=self.app_path,
+            language=ENGLISH_LANGUAGE_CODE,
+            locale=ENGLISH_US_LOCALE,
+        )
+        if self.language != ENGLISH_LANGUAGE_CODE:
```
vijay120, line 1138: Does it makes sense for anyone to use `language=en` here? The MLA seems to be use for only non-english usecases, maybe we should enforce that?

mindmeld/auto_annotator.py
```diff
@@ -835,5 +901,320 @@ def valid_entity_check(self, entity):
...
+            self.en_annotator.supported_entity_types
+        ).intersection(DUCKLING_TO_SYS_ENTITY_MAPPINGS[self.language])
+        return list(supported_entity_types)
+
+
+class MultiLingualAnnotator(Annotator):
+    """Custom Annotator class used to generate annotations."""
+
+    def __init__(self, *args, **kwargs):
+        """Initializes a TranslationDucklingAnnotator.
```
vijay120, line 1117: Initializes a MultiLingualAnnotator

vijay120: It would be helpful to write more detailed docs here on exactly what MLA is and how it is different from other annotators like TranslationAnnotator and NonTranslation annotator.

mindmeld/auto_annotator.py
```diff
@@ -835,5 +901,320 @@ def valid_entity_check(self, entity):
...
+            for candidate in candidates
+            if not (
+                candidate["dim"] == "amount-of-money"
+                and candidate["value"].get("unit") == "unknown"
+            )
+        ]
+
+
+class TranslationDucklingAnnotator(Annotator):
+    """Custom Annotator class used to generate annotations."""
```
vijay120, line 1023: More detailed docs here on what TranslationDucklingAnnotator does would be helpful

mindmeld/auto_annotator.py
```diff
@@ -835,5 +901,320 @@ def valid_entity_check(self, entity):
         return Entity.is_system_entity(entity) or entity in self.supported_entity_types
 
 
-register_annotator("SpacyAnnotator", SpacyAnnotator)
-register_annotator("BootstrapAnnotator", BootstrapAnnotator)
+class NoTranslationDucklingAnnotator(Annotator):
+    """Custom Annotator class used to generate annotations."""
```
vijay120, line 905: More detailed docs here on what NoTranslationDucklingAnnotator does would be helpful. Comparing and contrasting this with the other annotators would also be great

mindmeld/auto_annotator.py
```diff
@@ -835,5 +901,320 @@ def valid_entity_check(self, entity):
...
+            translator (str): A translator to use such as 'GoogleTranslator' or 'NoOpTranslator'.
+            language (str, optional): Language as specified using a 639-1/2 code.
+            locale (str, optional): The locale representing the ISO 639-1 language code and \
+                ISO3166 alpha 2 country code separated by an underscore character.
+            overwrite (bool): Whether to overwrite existing annotations with conflicting spans.
+            unannotate_supported_entities_only (bool): Only allow removal of supported entities.
+            unannotation_rules (list): List of Annotation rules.
+        """
+        super().__init__(*args, **kwargs)
+        self.translator = kwargs.get("translator", "NoOpTranslator")
```
vijay120, line 1132: self.translator = kwargs.get("translator", NoOpTranslator.__name__)

mindmeld/auto_annotator.py
```diff
@@ -835,5 +901,320 @@ def valid_entity_check(self, entity):
...
+        if self.language != ENGLISH_LANGUAGE_CODE:
+            self.duckling_annotator = self._get_duckling_annotator()
+            self.non_en_annotator = SpacyAnnotator(
+                app_path=self.app_path,
+                language=self.language,
+                locale=self.locale,
+            )
+
+    def _get_duckling_annotator(self):
+        if self.translator != "NoOpTranslator":
```
vijay120, line 1147: `if self.translator != NoOpTranslator.__name__:`

mindmeld/components/translators.py
```diff
@@ -0,0 +1,164 @@
...
+        pass
+
+    def get_translate_client(self):
+        return
+
+    def detect_language(self, text):
+        return
+
+    def translate(self, text, target_language):
+        return
```
vijay120, line 78: shouldnt NoOpTranslator just return the text? If it returns None, I can see the caller getting confused.

mindmeld/components/translators.py
```diff
@@ -0,0 +1,164 @@
...
+
+class Translator(metaclass=ABCMeta):
+    """Abstract Translator Base Class for Translators to be used by Mindmeld."""
+
+    def __init__(self):
+        """Creates a translation client after finding the credential path."""
+        self.translate_client = None
+
+    @abstractmethod
+    def get_translate_client(self):
```
vijay120, line 34: Do subclasses have to implement this method? Looks like its specific to service-based translators like google. Maybe the subclass should only implement it.

mindmeld/components/translators.py
```diff
@@ -0,0 +1,164 @@
...
+    @staticmethod
+    def get_translator(translator):
+        """A static method to get a translator
+
+        Args:
+            translator (str): Name of the desired translator class
+        Returns:
+            (Translator): Translator Class
+        """
+        if translator == "NoOpTranslator":
```
vijay120, line 157: use the class name to check

mindmeld/core.py
```diff
@@ -124,6 +124,10 @@ def shift(self, offset):
         """
         return Span(self.start + offset, self.end + offset)
 
+    def has_overlap(self, other):
+        """Determines whether two spans overlap."""
+        return self.end >= other.start and other.end >= self.start
```
vijay120, line 129: I guess this is on type of overlap, the others being one span is inside the other span length. For this case, does it break the operations of the caller or is this type of overlap fine?

mindmeld/system_entity_recognizer.py
```diff
@@ -269,7 +269,7 @@ def get_response(self, data):
         """
         try:
             response = requests.request(
-                "POST", self.url, data=data, timeout=float(SYS_ENTITY_REQUEST_TIMEOUT)
+                "POST", self.url, data=data, timeout=float(3)  # SYS_ENTITY_REQUEST_TIMEOUT
```
vijay120, line 272: fix this

source/userguide/auto_annotator.rst
```diff
@@ -304,23 +382,115 @@ For example, we can restrict the output of the previous example by doing the fol
...
+using the English NER model. The English detected entities are compared against duckling candidates for the non-English sentence. Duckling candidates with a match between the type and value of the entity or the translated body text
+are selected. If a translation service is not available, the :class:`MultiLingualAnnotator` selects the duckling candidates with the largest non-overlapping spans. The sections below describe the steps to setup the annotator depending on whether a translation service is being used.
+
+Annotating with a Translation Service (Google)
+''''''''''''''''''''''''''''''''''''''''''''''
+The :class:`MultiLingualAnnotator` can leverage the Google Translation API to better detect entities in non-English sentences. To use this feature, export your Google application credentials.
+
+.. code-block:: console
+
+	export GOOGLE_APPLICATION_CREDENTIALS="/Users/joesmith2/Documents/google_application_credentials.json"
```
vijay120, line 406: I think just have a generic `/<YOUR_PATH>/google_application_credentials.json` would be less confusing. Else, people will think they have to add the creds in the documents path.

tests/test_auto_annotator.py
```diff
@@ -9,14 +9,47 @@
...
 @pytest.fixture(scope="module")
-def spacy_annotator(kwik_e_mart_app_path):
-    return SpacyAnnotator(kwik_e_mart_app_path)
+def en_mla(kwik_e_mart_app_path):
+    return MultiLingualAnnotator(
+        app_path=kwik_e_mart_app_path, language="en", translator="NoOpTranslator"
+    )
 
 
+@pytest.fixture(scope="module")
```
vijay120, line 22: Is there a way to test `translator="GoogleTranslator"`without calling google translate? maybe through a mock?

vijay120 (commented):
Mostly nits. Very close to being approved :)

mindmeld/auto_annotator.py
KKsharma99: Done

mindmeld/auto_annotator.py
KKsharma99: Handling by first checking if `translator` is `None` then checking if it is `NoOpTranslator`. Raising AssertionErrors accordingly.

mindmeld/auto_annotator.py
KKsharma99: Added

mindmeld/components/translators.py
KKsharma99: Good point, returning text

mindmeld/system_entity_recognizer.py
KKsharma99: Updated

mindmeld/core.py
KKsharma99: This function covers the case where one span is inside the other. Example:

span_one = (1,10), span_two(5,8)
10 >= 5: span_two starts before span_one ends
8 >= 1: span_two starts after span_one starts

mindmeld/system_entity_recognizer.py
KKsharma99: Good point, removed the usage of "entity_dict". All Annotators now return QueryEntity objects from the parse() command.

mindmeld/components/translators.py
KKsharma99: Makes sense, removed this method from the base class.

tests/test_auto_annotator.py
KKsharma99: Added a TODO statement in test_auto_annotator.py as discussed

mindmeld/auto_annotator.py
KKsharma99: All annotators now return QueryEntity objects. Tests have been updated accordingly, working on updating the docs to reflect this change as well.

mindmeld/auto_annotator.py
KKsharma99: The MLA is currently designed to work for all languages, both English and Non-English. This makes it so that developers don’t confused on when to use a SpacyAnnotator versus MLA. Under the hood, MLA is returning the output of a SpacyAnnotator if the language is English.

mindmeld/auto_annotator.py
```diff
@@ -50,39 +58,60 @@ class Annotator(ABC):
...
 
-        if action == AnnotatorAction.ANNOTATE:
-            rules = config[AnnotatorAction.ANNOTATE.value]
-        elif action == AnnotatorAction.UNANNOTATE:
-            rules = config[AnnotatorAction.UNANNOTATE.value]
+        rules = (
+            self.annotation_rules
+            if action == AnnotatorAction.ANNOTATE
+            else self.unannotation_rules
+        )
```
vijay120, line 114: I think this is dangerous since if the caller has not explicitly set `ANNOTATE `, the unannotate code will be executed. I think we should just throw an error here if its not ANNOTATE or UNANNOTATE

mindmeld/auto_annotator.py
```diff
@@ -335,125 +320,126 @@ def _item_to_query_entity(item, processed_query):
...
+        """Get the name of a Spacy Model.
 
-    @staticmethod
-    def _load_model(model):
-        """ Load Spacy English model. Download if needed.
+        Returns:
+            spacy_model_name (str): Name of the Spacy NER model
+        """
+        model_type = "web" if self.language in SPACY_ANNOTATOR_WEB_LANGUAGES else "news"
+        return "_".join([self.language, "core", model_type, self.spacy_model_size])
```
vijay120, line 426: Nit: It might be clearer to do a fstring here: `f"{self.language}-{core}-..."`

mindmeld/auto_annotator.py
```diff
@@ -463,14 +449,53 @@ def _load_model(model):
...
+        """This function generates a list of supported entities for the given language.
+        These entities labels are mapped to MindMeld sys_entities.
+        The "misc" spacy entity is skipped since the category too broad to be
+        helpful in an application.
+
         Returns:
             supported_entity_types (list): List of supported entity types.
         """
-        return SPACY_ANNOTATOR_SUPPORTED_ENTITIES
+        spacy_supported_entities = [e.lower() for e in self.nlp.get_pipe("ner").labels]
```
vijay120, line 460: What exactly is `get_pipe `, it seems out of place to be a public method of the nlp object.

mindmeld/auto_annotator.py
```diff
@@ -764,33 +806,70 @@ def _resolve_person(self, entity):
         Returns:
             entity (dict): A resolved entity dict.
         """
-        entity["dim"] = self.ANNOTATOR_TO_DUCKLING_ENTITY_MAPPINGS[entity["dim"]]
+        entity["dim"] = ANNOTATOR_TO_SYS_ENTITY_MAPPINGS[entity["dim"]]
 
-        if len(entity["body"]) >= 2 and entity["body"][-2:] == "'s":
+        if self._is_plural_entity(entity):
```
vijay120, line 811: awesome!

mindmeld/auto_annotator.py
```diff
@@ -835,5 +923,337 @@ def valid_entity_check(self, entity):
...
+    """ The TranslationDucklingAnnotator detects entities in non-English sentences using
+    a translation service and Duckling by following these steps:
+        1. The non-English sentence is translated to English.
+        2. Spacy detects entities in the translated English sentence.
+        3. Duckling detects non-English entities in the non-English sentence.
+        4. A heuristic in parse() is used to match and filer the non-English entities
+        against the English entities.
+        5. The final set of filtered non-English entities are returned.
+    Unlike the NoTranslationDucklingAnnotator, this annotator uses a translation service.
+    Unlike the MultiLingualAnnotator, this annotator does not use non-English Spacy NER models.
```
vijay120, line 1056: Well done!

mindmeld/core.py
```diff
@@ -907,7 +931,11 @@ def _is_same_span(target, other):
 
 
 def _is_overlapping(target, other):
+    overlap = _get_overlap(target, other)
+    return overlap and not _is_subset(target, other) and not _is_superset(target, other)
```
vijay120, line 935: very descriptive!

vijay120 (approved):
Small nits that you should address, but its a green light for me.

mindmeld/auto_annotator.py
KKsharma99: Done

mindmeld/auto_annotator.py
KKsharma99: Done

mindmeld/auto_annotator.py
KKsharma99: It's a public method to get the component of your nlp. https://spacy.io/api/language#get_pipe

vembar (approved):
Nice work!

PR #276, opened 2021-02-10
Feature/Augmentor-Paraphraser

ritvikshrivastava:
Adds query-augmentation functionality to MindMeld:

1. Creates an abstract Augmentor class.
2. Adds EnglishParaphraser.
3. Adds MultiLingual Paraphraser (es, fr, pt, it, ro).
4. Documentation.

Example Usage:
1. Install Dependencies
```
pip install -e". [augment]"
```

2. Update app configuration (add to ``config.py``):

(a) For English:
```
AUGMENTATION_CONFIG = {
    "augmentor_class": "EnglishParaphraser",
    "batch_size": 8,
    "paths": [{"domains": ".*", "intents": ".*", "files": ".*",}],
    "path_suffix": ".augmented.txt",
}
```
(b) For MultiLingual Paraphraser:
```
LANGUAGE_CONFIG = {
    'language': 'es',
    'locale': 'es_MX'
}. # not needed for paraphraser if provided in CLI, this is fallback lang config for the purpose of paraphraser.

AUGMENTATION_CONFIG = {
    "augmentor_class": "MultiLingualParaphraser",
    "batch_size": 8,
    "paths": [{"domains": "greetings", "intents": "exit", "files": "train.txt",}],
    "path_suffix": ".augmented.txt",
}
```
3. Augment
```
mindmeld augment --app-path "path" --language "lang_code"
```

Generates ``*-augment.txt`` from ``*.txt``


mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def augment_queries(self, queries, **kwargs):
+        augmented_queries = []
+        for query in queries:
+            augmented_queries.extend(self._get_response(query, **kwargs))
+        return augmented_queries
+
+
+class MultiLingualParaphraser(Augmentor):
+    """Paraphraser class for generating paraphrases based on language code of the app
```
vembar, line 202: Modify comment to indicate that this only supports romance languages (French, Italian, Spanish, Portuguese, and Romanian) right now.

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+            num_beams=num_beams,
+            num_return_sequences=num_return_sequences,
+            temperature=1.5,
+        )
+        paraphrases = self.tokenizer.batch_decode(translated, skip_special_tokens=True)
+        return paraphrases
+
+    def augment_queries(self, queries, **kwargs):
+        augmented_queries = []
+        for query in queries:
```
vembar, line 196: We can actually make use of batching here to speed up the process.

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+            temperature=1.0,
+        )
+        translated_queries = list(set(translated_queries))
+
+        back_translated_queries = self._translate(
+            template=lambda text: f">>{self.language}<< {text}",
+            queries=translated_queries,
+            model=self.target_model,
+            tokenizer=self.target_tokenizer,
+            num_beams=5,
```
vembar, line 274: can reduce num_beams and num_sequences to 3 since we wan to generate at most 10 queries

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+
+        Args:
+            app_path (str): The location of the MindMeld app
+            config (dict, optional): A config object to use. This will
+                override the config specified by the app's config.py file.
+        """
+        super().__init__(app_path=app_path, config=config)
+
+        self.torch_device = "cuda" if torch.cuda.is_available() else "cpu"
+        self.language, _ = get_language_config(app_path=app_path)
```
vembar, line 217: Maybe it's better to add a check to ensure we support that language code and return an informative message if not.
You can find the supported language codes under source languages here: https://huggingface.co/Helsinki-NLP/opus-mt-ROMANCE-en

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+        """Initializes an English paraphraser.
+
+        Args:
+            app_path (str): The location of the MindMeld app
+            config (dict, optional): A config object to use. This will
+                override the config specified by the app's config.py file.
+        """
+        super().__init__(app_path=app_path, config=config)
+
+        model_name = "tuner007/pegasus_paraphrase"
```
vembar, line 163: @vijay120 Should we also have a local copy of these model in an S3 bucket or something as a fallback in case huggingface decides to phase it out for some reason?

mindmeld/components/_config.py
```diff
@@ -464,11 +464,28 @@
     },
 }
 
+DEFAULT_AUGMENTATION_CONFIG = {
```
vembar, line 467: Also, like we discussed, would be nice to expose some of the generation parameters to the developer through the config (especially for the multilingual models)

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+    )
+
+try:
+    from transformers import PegasusForConditionalGeneration, PegasusTokenizer
+    from transformers import MarianMTModel, MarianTokenizer
+except ImportError:
+    logger.info(
+        "Library not found: 'transformers'. Run 'pip install mindmeld[augment]'"
+        " to install."
+    )
```
vijay120, line 43: It would be cleaner to add this code block for trying to import torch into a `register_dependecies` function and follow the pattern in `mindmeld/path.py::607`

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def __init__(self, app_path, config=None):
+        """Initializes an augmentor.
+
+        Args:
+            app_path (str): The location of the MindMeld app
+            config (dict, optional): A config object to use. This will
+                override the config specified by the app's config.py file.
+        """
+        self.app_path = app_path
```
vijay120, line 61: I think its better to explicitly pass in the args used by this class, like `language` and `locale` to this class instead of deriving it from the `app_path`. I understand this is a practice we have used in the past, but we should move away from it.

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+        self.config = config or get_augmentation_config(app_path=app_path)
```
vijay120, line 62: Similar to my above statement, the caller of this class is responsible for explicitly passing the args and the responsibility of this class is to paraphrase given those primitives.

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+        self._resource_loader = ResourceLoader.create_resource_loader(app_path)
```
vijay120, line 63: resource loader should also be passed in

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+
+        Args:
+            queries (list): List of queries.
+
+        Return:
+            augmented_queries (list): List of augmented queries.
+        """
+        raise NotImplementedError("Subclasses must implement this method")
+
+    def _get_files(self, config):
```
vijay120, line 91: Pass in explicit paths

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+        Args:
+            app_path (str): The location of the MindMeld app
+            config (dict, optional): A config object to use. This will
+                override the config specified by the app's config.py file.
+        """
+        self.app_path = app_path
+        self.config = config or get_augmentation_config(app_path=app_path)
+        self._resource_loader = ResourceLoader.create_resource_loader(app_path)
+
+    def augment(self, config, **kwargs):
```
vijay120, line 65: enumerate the keyword args here

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+            filtered_paths (list): List of file paths to be augmeted.
+        """
+        all_file_paths = self._resource_loader.get_all_file_paths()
+
+        if not config["paths"]:
+            logger.warning(
+                """'paths' field is not configured or misconfigured in the `config.py`.
+                 Can't find files to augment."""
+            )
+            return
```
vijay120, line 107: should this be an empty list?

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+
+        rules = config["paths"]
+
+        filtered_paths = []
+
+        for rule in rules:
+            pattern = _get_pattern(rule)
+            compiled_pattern = re.compile(pattern)
```
vijay120, line 115: I would first validate the patterns before renumerating the paths since we can pick up invalid patterns earlier

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+        return queries
+
+    def _write_files(self, path, augmented_queries):
+        """Writes augmented queries to a new file in the path.
+
+        Args:
+            path (str): File path to the original file.
+            augmented_queries (list): List of augmented queries returned by class.
+        """
+        write_path = path.strip(".txt") + ".augmented.txt"
```
vijay120, line 143: Add a suffix arg to the function instead of hardcoding ".augmented.txt"

mindmeld/augmentation.py
vijay120: I agree. @jre21 could you help us to upload the model to one of our local aws buckets?

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+        Args:
+            query (str): An application query.
+            num_return_sequences (int, optional): Maximum number of paraphrases to be generated.
+            num_beams (int, optional):
+
+        Return:
+            paraphrases (list(str)): List of paraphrased queries.
+        """
+        batch = self.tokenizer.prepare_seq2seq_batch(
+            [query], truncation=True, padding="longest", max_length=60
```
vijay120, line 182: Is it worth exposing some of the params here like `max_length ` to the end dev?

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+        )
+        translated = self.model.generate(
+            **batch,
+            max_length=60,
```
vijay120, line 186: add these params as constants if they are not configured

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+        self.en_model = MarianMTModel.from_pretrained(en_model_name)
+        self.en_model.to(self.torch_device)
+
+        target_model_name = "Helsinki-NLP/opus-mt-en-ROMANCE"
+        self.target_tokenizer = MarianTokenizer.from_pretrained(target_model_name)
+        self.target_model = MarianMTModel.from_pretrained(target_model_name).to(
+            self.torch_device
+        )
+
+    def _translate(self, **kwargs):
```
vijay120, line 230: enumerate kwargs

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+        queries = kwargs.pop("queries")
+        model = kwargs.pop("model")
+        tokenizer = kwargs.pop("tokenizer")
+
+        queries = [template(query) for query in queries]
+        encoded = tokenizer.prepare_seq2seq_batch(queries, return_tensors="pt")
+        for key in encoded:
+            encoded[key] = encoded[key].to(self.torch_device)
+
+        translated = model.generate(**encoded, **kwargs)
```
vijay120, line 250: enumerate

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+        translated_queries = tokenizer.batch_decode(
+            translated, skip_special_tokens=True
+        )
+        return translated_queries
+
+    def augment_queries(self, queries):
+        translated_queries = self._translate(
+            template=lambda text: f"{text}",
```
vijay120, line 258: isnt this just `str(input)`? Why do we need to do this?

mindmeld/augmentation.py
```diff
@@ -0,0 +1,285 @@
...
+            queries=queries,
+            model=self.en_model,
+            tokenizer=self.en_tokenizer,
+            num_beams=5,
```
vijay120, line 262: Again, I would set all these params as global vars and consolidate them somewhere for easy changing

mindmeld/auto_annotator.py
```diff
@@ -85,7 +87,7 @@ def _get_file_entities_map(self, action: AnnotatorAction, config):
             rules = config[AnnotatorAction.UNANNOTATE.value]
 
         for rule in rules:
-            pattern = Annotator._get_pattern(rule)
+            pattern = _get_pattern(rule)
```
vijay120, line 90: Wont this fail?

mindmeld/components/_config.py
```diff
@@ -464,11 +464,28 @@
...
+    "augmentor_class": "EnglishParaphraser",
```
vijay120, line 468: Do we have an opinion of if there would be multiple english paraphrasers in the future? If not, can we just use the locale and language to decide which paraphraser to use instead of specifying it here?

mindmeld/auto_annotator.py
vijay120: Nvm looks like its being imported. If so, remove the leading camelcase since its no longer a private function

tests/test_augmentation.py
```diff
@@ -0,0 +1,35 @@
...
+@pytest.mark.augment
+@pytest.mark.parametrize(
+    "query, num_paraphrases",
+    [
+        ("some text", 10),
+        ("another text", 5),
+        ("yet another text", 2),
+    ],
+)
+def test_num_paraphrases(english_paraphraser, query, num_paraphrases):
```
vijay120, line 31: Can we also test for the results to make sure they are deterministic and make sense?

tests/test_augmentation.py
```diff
@@ -0,0 +1,35 @@
...
+    paraphrases = english_paraphraser.augment_queries(
+        [query], num_return_sequences=num_paraphrases
+    )
+    assert num_paraphrases == len(paraphrases)
```
vijay120, line 35: Also please test for the multilingual one as well.

vijay120 (changes requested):
Some comments

mindmeld/augmentation.py
ritvikshrivastava: This function takes in the config and resolves the rules set for paths

mindmeld/augmentation.py
ritvikshrivastava: The kwargs are variable, and change for different models. Can't explicitly set them here.

mindmeld/augmentation.py
ritvikshrivastava: The invalid patterns should be removed in the filtering step that follows.

mindmeld/augmentation.py
ritvikshrivastava: The kwargs are variable here again (and a longer list). I was thinking of doing that in the docstring instead of explicitly enumerating them in the kwargs.

mindmeld/augmentation.py
vijay120: Ok got it. Then it makes sense.

mindmeld/augmentation.py
vijay120: I think its better for readability if paths was just pass in. The config is not being used for anything else here.

mindmeld/augmentation.py
vijay120: So it looks like this function needs some compulsory arguments, like template, queries, models etc. There are two ways to do this:
a) Make the compulsory arguments position arguments with no defaults
b) Make these compulsory keyword args as follows:
```
def _translate(self, *, template, queries, model, tokenizer, **kwargs):
```

mindmeld/augmentation.py
vijay120: @ritvikshrivastava I can upload the model to the s3 bucket, let me know.

setup.py
```diff
@@ -92,12 +92,16 @@
...
+            'transformers~=3.5.1; python_version>="3.6"',
+            'sentence-transformers~=0.3; python_version>="3.6"',
             "elasticsearch>=7.0",
         ],
         "examples": [
             'connexion>=2.7.0; python_version>="3.6"',
         ],
+        "augment":[
+            'torch~=1.7.0; python_version>="3.6"',
+            'transformers~=3.5.1; python_version>="3.6"'
```
vembar, line 104: I think we may have to explicitly install SentencePiece library which the multilingual paraphraser needs.

mindmeld/augmentation.py
ritvikshrivastava: Added as default params in the constructors of respective paraphrasers, as discussed.

mindmeld/augmentation.py
ritvikshrivastava: Took approach b)

mindmeld/augmentation.py
ritvikshrivastava: Added a check in the abstract constructor.

mindmeld/components/_config.py
ritvikshrivastava: The config now has a params dict that can be populated with configurable model parameters. We should provide a detailed documentation of acceptable parameters in the documentation.

mindmeld/augmentation.py
```diff
@@ -0,0 +1,315 @@
...
+        if not _is_module_available("transformers"):
+            raise ModuleNotFoundError(
+                "Library not found: 'transformers'. Run 'pip install mindmeld[augment]' to install."
+            )
+
+    def _check_lang_support(self):
+        if self.lang not in SUPPORTED_LANG_CODES:
+            raise UnsupportedLanguageError(
+                f"'{self.lang}' is not supported yet. "
+                "English, Spanish, French, Portuguese and Italian are currently supported."
```
vembar, line 73: Nit: Add language codes as well. 

"English(en), Spanish(es), French(fr), Portuguese(pt) and Italian(it) are currently supported."

vembar: Also add Romanian (ro)

mindmeld/augmentation.py
```diff
@@ -0,0 +1,315 @@
...
+            encoded[key] = encoded[key].to(self.torch_device)
+
+        translated = model.generate(**encoded, **kwargs)
+        translated_queries = tokenizer.batch_decode(
+            translated, skip_special_tokens=True
+        )
+        return translated_queries
+
+    def augment_queries(self, queries):
+        translated_queries = self._translate(
```
vembar, line 293: Shouldn't we set a batch size here and then call translate()?

mindmeld/augmentation.py
```diff
@@ -0,0 +1,315 @@
...
+        self.en_model.to(self.torch_device)
+
+        target_model_name = "Helsinki-NLP/opus-mt-en-ROMANCE"
+        self.target_tokenizer = MarianTokenizer.from_pretrained(target_model_name)
+        self.target_model = MarianMTModel.from_pretrained(target_model_name).to(
+            self.torch_device
+        )
+
+        # Use default model params if not customized for app
+        self.params = params or {
```
vembar, line 265: Though we agreed on having two sets of parameters here. One for the pivot language translation and the other for the "back-translation".

mindmeld/components/_config.py
vembar: Same comment as before. Are we now having just a single set of parameters for both forward and back translations?

mindmeld/augmentation.py
vembar:
```
for i in range(0, len(queries), batch_size):
    query_batch = queries[i:i+batch_size]
    augmented_queries.extend(self._get_response(query_batch, **kwargs))
```
Maybe set `batch_size` to 8?

mindmeld/augmentation.py
```diff
@@ -0,0 +1,315 @@
...
+            model_name, force_download=False
+        ).to(torch_device)
+
+        # Use default model params if not customized for app
+        self.params = params or {
+            "max_length": 60,
+            "num_beams": 10,
+            "num_return_sequences": 10,
+            "temperature": 1.5,
+        }
```
vembar, line 196: We should update the key values with the user params rather than expect them to set all of them?

ritvikshrivastava: Yeah made this change already. Will push it up.

vembar (changes requested):
A few comments

mindmeld/augmentation.py
```diff
@@ -0,0 +1,324 @@
...
+
+        Args:
+            path (str): File path to the original file.
+            augmented_queries (list): List of augmented queries returned by class.
+        """
+        write_path = path.strip(".txt") + suffix
+
+        with open(write_path, "w") as outfile:
+            for query in augmented_queries:
+                outfile.write(query.strip("\n") + "\n")
```
vijay120, line 154: I think `query.rstrip` might be more general here

mindmeld/augmentation.py
vijay120: Lets do this in a subsequent PR

mindmeld/augmentation.py
```diff
@@ -0,0 +1,324 @@
...
+        # Update default params with user model config
+        self.params = {
+            "max_length": 60,
+            "num_beams": 10,
+            "num_return_sequences": 10,
+            "temperature": 1.5,
+        }
+
+        if params:
+            self.params.update(params)
```
vijay120, line 200: As discussed, lets try and get rid of params and instead enumerate the primitive args for the model in the constructor with good defaults.

mindmeld/augmentation.py
```diff
@@ -0,0 +1,324 @@
...
+        # Update default params with user model config
+        self.params = {
+            "num_beams": 3,
+            "num_return_sequences": 3,
+            "top_k": 0,
+            "temperature": 1.0,
+        }
+
+        if params:
+            self.params.update(params)
```
vijay120, line 277: same here.

mindmeld/augmentation.py
ritvikshrivastava: Created batching within translate.

mindmeld/components/_config.py
ritvikshrivastava: Added.

mindmeld/augmentation.py
```diff
@@ -0,0 +1,340 @@
...
+                **self.params,
+            )
+            all_generated_queries.extend(
+                self.tokenizer.batch_decode(generated, skip_special_tokens=True)
+            )
+        return all_generated_queries
+
+    def augment_queries(self, queries, **kwargs):
+        augmented_queries = []
+        augmented_queries = self._get_response(queries, **kwargs)
```
vembar, line 235: return unique paraphrases here similar to the Multilingual one.

mindmeld/components/_config.py
```diff
@@ -464,11 +464,34 @@
...
+            "intents": ".*",
+            "files": ".*",
+        }
+    ],
+    "params": {
+        "fwd_params": {},
+        "reverse_params": {},
+        "batch_size": 8,
+    },
+    "path_suffix": ".augmented.txt"
```
vembar, line 481: Nit: Would suggest changing it to `_augment.txt` or `-augment.txt`

tests/test_augmentation.py
```diff
@@ -0,0 +1,45 @@
...
+@pytest.mark.extras
+@pytest.mark.parametrize(
+    "query",
+    [
+        ("some text"),
+        ("another text"),
+        ("yet another text"),
+    ],
+)
+def test_num_paraphrases(english_paraphraser, query):
```
vembar, line 42: add a test for the multilingual paraphraser? You can use a predetermined example and check for equivalence.

mindmeld/augmentation.py
```diff
@@ -0,0 +1,340 @@
...
+
+        target_model_name = "Helsinki-NLP/opus-mt-en-ROMANCE"
+        self.target_tokenizer = MarianTokenizer.from_pretrained(target_model_name)
+        self.target_model = MarianMTModel.from_pretrained(target_model_name).to(
+            self.torch_device
+        )
+
+        # Update default params with user model config
+        self.fwd_params = {
+            "num_beams": 3,
```
vembar, line 281: Can we also set max_length to 60 here?

mindmeld/augmentation.py
```diff
@@ -0,0 +1,340 @@
...
+        template = lambda text: f">>{self.lang}<< {text}"
+        translated_queries = [template(query) for query in set(translated_queries)]
+
+        reverse_translated_queries = self._translate(
+            queries=translated_queries,
+            model=self.target_model,
+            tokenizer=self.target_tokenizer,
+            **self.reverse_params,
+        )
+        augmented_queries = list(set(p.lower() for p in reverse_translated_queries))
```
vembar, line 334: From my experiments with WxA, for some of the shorter queries, it seems to generate sentences with just punctuations sometimes. One way of eliminating them could be to retain queries that have some alpha-numeric characters.

source/userguide/augmentation.rst
```diff
@@ -0,0 +1,103 @@
+Data Augmentation in MindMeld
+=============================
+
+MindMeld provides the ability to augment data to collect more diverse queries and make the applications more robust.
```
vembar, line 4: When starting out to build a new conversational application in a custom domain, it is often the case that we have very limited training data to train models. In such scenarios, one way of bootstrapping your models is by using data augmentation techniques to increase the size of our training set. MindMeld now provides the ability to augment data via automatic paraphrasing in English and other languages to help make the applications more robust.

source/userguide/augmentation.rst
```diff
@@ -0,0 +1,103 @@
...
+--------------------------
+
+The :attr:`DEFAULT_AUGMENTATION_CONFIG` shown below is the default config for an Augmentor.
+A custom config can be included in :attr:`config.py` by duplicating the default config and renaming it to :attr:`AUGMENTATION_CONFIG`.
+
+.. code-block:: python
+
+    DEFAULT_AUGMENTATION_CONFIG = {
+        "augmentor_class": "EnglishParaphraser",
+        "paths": [
```
vembar, line 41: I think we're missing documentation for batch_size?

source/userguide/augmentation.rst
```diff
@@ -0,0 +1,103 @@
...
+``'augmentor_class'`` (:class:`str`, default: :class:`EnglishParaphraser`): The augmentor class that can specified for deciding on the augmentation tool to be used. Currently, we recommend using :class:`EnglishParaphraser` for English language applications and :class:`MultiLingualParaphraser` for all other supported languages.
+
+``'paths'`` (:class:`list`): List of regex based path rules to select files to be augmented. These files become the baseline for the augmentors and are used to generate new data. By default, all training and testing files are augmented.
+
+``'path_suffix'`` (:class:`str`): The default suffix that is appended to the name of the original file being augmented to generate new files with the augmented data.
+
+English Paraphraser
+-------------------
+
+The English paraphraser uses a state-of-the-art text summarization model `PEGASUS <https://ai.googleblog.com/2020/06/pegasus-state-of-art-model-for.html>`_ for generating paraphrases. This model has been fine-tuned for the task of paraphrasing.
```
vembar, line 62: Should we also cite the fine tuned model here? @vijay120

source/userguide/augmentation.rst
```diff
@@ -0,0 +1,103 @@
...
+
+    mindmeld augment --app-path <app_path> --lang "en"
+
+In the config for this paraphraser class, the ``'augmentor_class'`` should be set to :class:`EnglishParaphraser`.
+
+
+Multi-Lingual Paraphraser
+-------------------------
+
+The multi-lingual paraphraser in MindMeld uses back-translation as the underlying concept to generate paraphrases. Given an application in one of the supported languages, the forward model translates the current set of queries to English, generating a number of English translations. Next, the reverse model translates each of the English translations into one or more queries in the original language. This results in a paraphrased set of queries in the original language.
```
vembar, line 77: The multi-lingual paraphraser in MindMeld uses machine translation as the underlying concept to generate paraphrases.

vembar: Same question as in the previous question. I think we should be citing the helsinki-NLP models here?

tests/test_augmentation.py
```diff
@@ -0,0 +1,87 @@
...
+@pytest.mark.extras
+@pytest.mark.parametrize(
+    "query",
+    [
+        ("aumenta el volumen"),
+    ],
+)
+def test_spanish_paraphrases(multilingual_paraphraser, query):
+    paraphrases = multilingual_paraphraser.augment_queries([query])
+    assert "Aumentar el volumen." in paraphrases
```
vembar, line 87: wont this fail since we lower case everything before returning the paraphrases?

mindmeld/augmentation.py
vembar: Consider something like this - 
```
augmented_queries = list(set([p.lower() for p in reverse_translated_queries if any(c.isalnum() for c in p)]))
```

mindmeld/augmentation.py
```diff
@@ -0,0 +1,345 @@
...
+            "max_length": 60,
+            "num_beams": 3,
+            "num_return_sequences": 3,
+            "temperature": 1.0,
+            "top_k": 0,
+        }
+
+        self.reverse_params = {
+            "max_length": 60,
+            "num_beams": 5,
```
vembar, line 292: Flip num_beam/num_return_sequences values - 5 for forward and 3 for reverse.

mindmeld/augmentation.py
```diff
@@ -0,0 +1,345 @@
...
+    def __init__(self, lang, paths, path_suffix, resource_loader):
+        """Initializes an augmentor.
+
+        Args:
+            lang (str): The lang code for paraphrasing
+            paths (list): Path rules for fetching relevant files to Paraphrase.
+            path_suffix (str): Suffix to be added to new augmented files.
+            resource_loader (object): Resource Loader object for the application.
+        """
+        self.lang = lang
```
vijay120, line 51: maybe `language_code` would be more expressive here

mindmeld/augmentation.py
```diff
@@ -0,0 +1,345 @@
...
+        self.paths = paths
```
vijay120, line 52: `files_to_paraphrase`

mindmeld/augmentation.py
vijay120: maybe do a `lang.lower()` here as well

mindmeld/augmentation.py
```diff
@@ -0,0 +1,345 @@
...
+            path (str): File path.
+
+        Return:
+            queries (list): List of queries in the file path.
+        """
+        with open(path, "r") as f:
+            queries = f.readlines()
+        return queries
+
+    def _write_files(self, path, augmented_queries, suffix):
```
vijay120, line 145: I think `_write_to_file` seems more descriptive?

mindmeld/augmentation.py
```diff
@@ -0,0 +1,345 @@
...
+                " and Spanish (es) are currently supported."
+            )
+
+    def augment(self, **kwargs):
+        """Augments queries given initial queries in application."""
+        filtered_paths = self._get_files(paths=self.paths)
+
+        for path in tqdm(filtered_paths):
+            queries = self._read_path_queries(path)
+            augmented_queries = self.augment_queries(queries, **kwargs)
```
vijay120, line 85: Add a ToDo here to consider using a generator here and write to file incrementally in case we run of mem in the future. Currently, if the file size is too large, it might fail here.

mindmeld/augmentation.py
```diff
@@ -0,0 +1,345 @@
...
+
+        Args:
+            queries (list): List of queries.
+
+        Return:
+            augmented_queries (list): List of augmented queries.
+        """
+        raise NotImplementedError("Subclasses must implement this method")
+
+    def _get_files(self, paths=None):
```
vijay120, line 100: `path_rules`?

mindmeld/augmentation.py
```diff
@@ -0,0 +1,345 @@
...
+        all_file_paths = self._resource_loader.get_all_file_paths()
+
+        if not paths:
+            logger.warning(
+                """'paths' field is not configured or misconfigured in the `config.py`.
+                 Can't find files to augment."""
+            )
+            return []
+
+        rules = paths
```
vijay120, line 118: This assignment seems redundant.

mindmeld/augmentation.py
```diff
@@ -0,0 +1,345 @@
...
+            pattern = get_pattern(rule)
+            compiled_pattern = re.compile(pattern)
+            filtered_paths.extend(
+                self._resource_loader.filter_file_paths(
+                    compiled_pattern=compiled_pattern, file_paths=all_file_paths
+                )
+            )
+        return filtered_paths
+
+    def _read_path_queries(self, path):
```
vijay120, line 132: same, this is a utility function

mindmeld/augmentation.py
vijay120: I also think these functions seem like utility functions that dont need to be a member of this class. The prefix can also be generalized

mindmeld/augmentation.py
```diff
@@ -0,0 +1,345 @@
...
+        )
+        self.tokenizer = PegasusTokenizer.from_pretrained(model_name)
+        self.model = PegasusForConditionalGeneration.from_pretrained(
+            model_name, force_download=False
+        ).to(torch_device)
+
+        # Update default params with user model config
+        self.batch_size = batch_size
+
+        self.params = {
```
vijay120, line 198: maybe its better to call this `default_paraphraser_model_params`

mindmeld/augmentation.py
```diff
@@ -0,0 +1,332 @@
...
+
+        Return:
+            paraphrases (list(str)): List of paraphrased queries.
+        """
+        all_generated_queries = []
+        for pos in range(0, len(queries), self.batch_size):
+            batch = self.tokenizer.prepare_seq2seq_batch(
+                queries[pos : pos + self.batch_size],
+                truncation=True,
+                padding="longest",
```
vijay120, line 206: this should also be `self.default_paraphraser_model_params["padding"]`

mindmeld/augmentation.py
```diff
@@ -0,0 +1,332 @@
...
+                **batch,
+                **self.default_paraphraser_model_params,
+            )
+            all_generated_queries.extend(
+                self.tokenizer.batch_decode(generated, skip_special_tokens=True)
+            )
+        return all_generated_queries
+
+    def augment_queries(self, queries, **kwargs):
+        augmented_queries = []
```
vijay120, line 219: remove this line

mindmeld/augmentation.py
```diff
@@ -0,0 +1,332 @@
...
+        self.batch_size = batch_size
+
+        self.default_paraphraser_model_params = {
+            "max_length": 60,
+            "num_beams": 10,
+            "num_return_sequences": 10,
+            "temperature": 1.5,
+        }
+
+    def _get_response(self, queries):
```
vijay120, line 192: maybe `_generate_paraphrases`?

mindmeld/cli.py
```diff
@@ -785,6 +792,36 @@ def _get_auto_annotator_config(app_path, overwrite=False, unannotate_all=False):
...
+    "--app-path",
+    required=True,
+    help="The application's path.",
+)
+@click.option(
+    "--lang",
+    help="Augmentation language.",
+)
+@click.option(
+    "--batch_size",
```
vijay120, line 806: not sure we should expose this?

mindmeld/cli.py
```diff
@@ -785,6 +792,36 @@ def _get_auto_annotator_config(app_path, overwrite=False, unannotate_all=False):
...
 
 
+@shared_cli.command("augment", context_settings=CONTEXT_SETTINGS)
+@click.option(
+    "--app-path",
+    required=True,
+    help="The application's path.",
+)
+@click.option(
+    "--lang",
```
vijay120, line 802: I would advice to use the flag name `language` here to keep it consistent with existing api terminology: https://www.mindmeld.com/docs/userguide/internationalization.html?highlight=language%20code

mindmeld/augmentation.py
ritvikshrivastava: Tokenizer and model params bunched together?

mindmeld/cli.py
```diff
@@ -785,6 +792,36 @@ def _get_auto_annotator_config(app_path, overwrite=False, unannotate_all=False):
...
+    help="Augmentation language.",
```
vijay120, line 803: Does the language code have to follow a certain format, like ISO 639? If so, its good to mention it here.

mindmeld/constants.py
```diff
@@ -294,3 +294,17 @@
...
+    "files" into a regex pattern.
+
+    Args:
+        rule (dict): An annotation or augmentation rule.
+
+    Returns:
+        pattern (str): Regex pattern specifying allowed file paths.
+    """
+    pattern = [rule[x] for x in ["domains", "intents", "files"]]
+    return ".*/" + "/".join(pattern)
```
vijay120, line 310: use f-strings here

mindmeld/models/helpers.py
```diff
@@ -100,6 +101,50 @@ def create_annotator(config):
...
+                    "domains": ".*",
+                    "intents": ".*",
+                    "files": ".*",
+                }
+            ],
+        )
+        path_suffix = config.get("path_suffix", "-augment.txt")
+        return AUGMENTATION_MAP[config["augmentor_class"]](
+            batch_size=batch_size,
+            lang=lang,
```
vijay120, line 138: Lets standardize to `language`

ritvikshrivastava: Just made that change everywhere. all 'lang' has been changed to 'language'

source/userguide/augmentation.rst
```diff
@@ -0,0 +1,106 @@
+Data Augmentation in MindMeld
+=============================
+
+When starting out to build a new conversational application in a custom domain, it is often the case that we have very limited training data to train models. In such scenarios, one way of bootstrapping your models is by using data augmentation techniques to increase the size of our training set. MindMeld now provides the ability to augment data via automatic paraphrasing in English and other languages to help make the applications more robust.
```
vijay120, line 4:
```suggestion
When starting out to build a new conversational application in a custom domain, it is often the case that we have very limited training data to train models. In such scenarios, one way of bootstrapping these models is by using data augmentation techniques to increase the size of our training set. MindMeld now provides the ability to augment data via automatic paraphrasing in English and other languages to help make the applications more robust.
```

mindmeld/cli.py
ritvikshrivastava: updated!

source/userguide/augmentation.rst
```diff
@@ -0,0 +1,106 @@
...
+The following section explains this in more detail.
+
+Using the Augmentors
+--------------------
+
+The data augmentation tool in MindMeld is a command line functionality. We demonstrate below the use-cases and configurations that
+can be defined to get the best augmentation results based on the application.
+
+Currently, we support data augmentation through paraphrasing for the following languages: English (en), French (fr), Italian (it), Portugese (pt),
+Romanian (ro), Spanish (es). This is done with the use of two models: the English paraphraser and the multi-lingual paraphraser.
```
vijay120, line 27: state the language code format developers have to use

mindmeld/cli.py
ritvikshrivastava: Discussing with @vembar , we thought this might be useful for individuals to modify depending on the processing capabilities of their machines?

source/userguide/augmentation.rst
vijay120: Yes, also providing the link to the finetuned pegasys model in huggingface would be great

mindmeld/cli.py
ritvikshrivastava: Good point. ISO 639-1. Added that.

mindmeld/augmentation.py
ritvikshrivastava: made a ``default_tokenizer_params`` separately.

mindmeld/cli.py
vembar: @vijay120 increasing or decreasing batch size will affect latency. So thought it might be useful to expose to the user?

vijay120: Is latency important during augmentation time? Also does the latency vary a lot depending on the batch size? I would prefer more parsimonious configs in the CLI flags to prevent developers from grokking too much and trying the tool out, while using `config.py` to do more advanced changes. However, if we think batch_size is something developers have to worry about when trying out the tool, we can add that.

ritvikshrivastava: Removed from CLI, can be still modified in the config.

vembar (approved):

mindmeld/augmentation.py
```diff
@@ -0,0 +1,340 @@
...
+        raise NotImplementedError("Subclasses must implement this method")
+
+    def _validate_generated_query(self, query):
+        """Validates whether augmented query has atleast one alphanumeric character
+
+        Args:
+            query (str): Generated query to be validated.
+        """
+        pattern = re.compile("^.*[a-zA-Z0-9].*$")
+        if pattern.search(query):
```
vijay120, line 108: nit: `return pattern.search(query) and True`

mindmeld/models/helpers.py
```diff
@@ -100,6 +101,50 @@ def create_annotator(config):
         raise KeyError(msg.format(config["annotator_class"]))
 
 
+def create_augmentor(config, language, resource_loader):
```
vijay120, line 104: I think this function should be in the `mindmeld/augmentation.py` in a factory class that returns augmenter objects.

vijay120 (approved):

PR #277, opened 2021-02-15
Fixed regex patterns

vijay120:
Fixed regex patterns for the rasa converter. Currently, the converter would do this:

**rasa entry**: `get the eox info for [12.2(33)sxj](software_version)`:
**mindmeld entry current**: `get the eox info for {12.233}sxj|software_version}`
**mindmeld entry with bug fix**: `get the eox info for {12.2(33)sxj|software_version}`


KKsharma99 (approved):
Looks good

mindmeld/converter/rasa.py
```diff
@@ -77,18 +80,12 @@ def _create_intent_training_file(self, intent_directory):
 
     @staticmethod
     def _does_intent_ex_contain_entity(intent_example):
-        return len(re.findall(r"\[.*\]\(.*\)", intent_example)) > 0
+        return len(RASA_ENTITY_REGEX.findall(intent_example)) > 0
 
-    def _write_intent_with_entinty(self, intent_f, intent_example):
+    def _write_intent_with_entity(self, intent_f, intent_example):
```
jjjacksn, line 85: nice

mindmeld/converter/rasa.py
```diff
@@ -26,6 +26,9 @@
 
 logger = logging.getLogger(__name__)
 
+RASA_ENTITY_REGEX = re.compile(r"(\[(.*?)\]\((.*?)\))")
+MINDMELD_ENTITY_REGEX = re.compile(r"\{.*?\}")
```
jjjacksn, line 30: I think this would also match `some query with weird characters []()` and convert to `some query with weird characters {|}`

Is this desirable?

We could change it by using `.+` instead of `.*`

jjjacksn (commented):
Looks good, but one question about desired behavior

jjjacksn (changes requested):
Thinking about it some more -- can we add a test for this?

tests/converter/test_rasa.py
```diff
@@ -0,0 +1,77 @@
...
+        "What is the End of Life Product Restart Time for [XYZ2990-ESPN19-N](PID) ?",
+        "What is the End of Life Product Restart Time for {xyz2990-espn19-n|pid} ?",
+        {'{xyz2990-espn19-n|pid}'}
+    ),
+    (
+        "XyZ date for [21.1(5)L&](proDucT_version1) - XyZ date for [21.1(5)L&](product_version2)",
+        "XyZ date for {21.1(5)l&|product_version1} - XyZ date for {21.1(5)l&|product_version2}",
+        {'{21.1(5)l&|product_version1}', '{21.1(5)l&|product_version2}'}
+    ),
+]
```
jjjacksn, line 67: Do we think users will be okay with the case differences here?

- *21.1(5)L&* becomes *21.1(5)**l**&*
- *XYZ2990-ESPN19-N* becomes *xyz2990-espn19-n*

jjjacksn (approved):
LGTM. I'm just not 100% we want the converter to be lower casing the entity texts

tests/converter/test_rasa.py
vijay120: fixed.

PR #279, opened 2021-02-19
Regex feature rfc

vijay120:
This PR is a RFC for the regex pattern matching functionality.

### DIFF
--- a/rfcs/regex_feature.md
+++ b/rfcs/regex_feature.md
@@ -1,0 +1,89 @@
+# Regex feature
+
+A feature based on regex pattern matching that MindMeld's NER models can use to detect custom entities. 
+
+- Developers expect NER models to correctly infer that the query `AVET128-321` is a `pid` entity 
+without providing surrounding sentence context to the classifier. This is more relevant in services that
+just use MindMeld for NLP while doing their own dialog management. 
+- Developers also want granular distinctions between entities that occur in the same context, 
+for example, `look up {SKU120|sku}` and `look up {123-ABC-12|pid}`. Currently, we ask developers to differ
+these types of distinction to the business logic in the dialogue manager.
+- DialogFlow, NLTK and other similar platforms have the regex matching feature. Moreover, we already have an 
+orthographic matching feature in MindMeld. This feature would augment that.
+
+
+## Usage
+
+The developer makes a new file called `regex_matches.txt` for the entity type in the entities folder that needs 
+these regex matches like this following:
+
+```text
+entities
+├── account_type
+│   ├── gazetteer.txt
+│   ├── mapping.json
+│   └── regex_matches.txt
+├── autopay_status
+│   └── mapping.json
+└── credit_amount
+    └── mapping.json
+```
+
+Inside the `regex_matches.txt` file, the developer lists regex patterns compliant with python syntax that need 
+to be matched for this custom entity. 
+
+```text
+\w{4}[0-9]{3}\-[0-9]{3}
+[0-9]{3}\-[0-9]{3}
+```
+
+In the above syntax, the `account_type` entity would match words like `ABCD-123-281` (first pattern) 
+and `123-456` (second pattern).
+
+Finally, the developer would have the ability to turn on/off the regex feature through the model config in 
+config.py using the key,value `regex-seq: True`. The feature will be off by default.
+
+```python
+ENTITY_RECOGNIZER_CONFIG = {
+    "features": {
+        "bag-of-words-seq": {
+            "ngram_lengths_to_start_positions": {
+                1: [-2, -1, 0, 1, 2],
+                2: [-2, -1, 0, 1],
+            }
+        },
+        "sys-candidates-seq": {"start_positions": [-1, 0, 1]},
+        "regex-seq": True
+    }
+}
+```
+
+If one is using the `view_extracted_features` method of the entity classifier to inspect the features of a query,
+the following result would be shown:
+
+```python
+er.view_extracted_features("look up 123-456")
+```
+
+```text
+[{...}, {...}, {'regex_match|type:account_type:1', ...}]
+```
+
+## Problems with this approach
+
+- The processing time during inference will increase significantly since each input token has to be matched against 
+each regex pattern for every entity type.
+- The training time will also increase for the similar reason
+- The feature could cast a wide net on every token, for example `.*` would match against all tokens. However since the 
+classifier would able to learn from this high noise signal, it would reduce it's dependence on it to make a classification.
+- There could be overlaps with regex patterns from different entity types, but the classifier would be able to learn from
+those patterns.
+
+## Alternative ideas
+
+- Have the regex mapping in `config.py` instead. This idea is a bit confusing since it doesnt localize all the entity 
+logic in the same folder location.
+- Not have this feature since it could be abused and developers will be confused. Moreover, the performance costs could 
+add up to an already slow platform.
+
+
--- a/rfcs/regex_feature.md
+++ b/rfcs/regex_feature.md
@@ -78,6 +78,8 @@ each regex pattern for every entity type.
 classifier would able to learn from this high noise signal, it would reduce it's dependence on it to make a classification.
 - There could be overlaps with regex patterns from different entity types, but the classifier would be able to learn from
 those patterns.
+- This approach locks-in the one-language one-mindmeld application idea since these regex patterns will not translate
+across languages.
 
 ## Alternative ideas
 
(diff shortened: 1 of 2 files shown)


rfcs/regex_feature.md
```diff
@@ -0,0 +1,91 @@
+# Regex feature
+
+A feature based on regex pattern matching that MindMeld's NER models can use to detect custom entities. 
+
+- Developers expect NER models to correctly infer that the query `AVET128-321` is a `pid` entity 
+without providing surrounding sentence context to the classifier. This is more relevant in services that
+just use MindMeld for NLP while doing their own dialog management. 
+- Developers also want granular distinctions between entities that occur in the same context, 
+for example, `look up {SKU120|sku}` and `look up {123-ABC-12|pid}`. Currently, we ask developers to differ
```
jre21, line 9:
```suggestion
for example, `look up {SKU120|sku}` and `look up {123-ABC-12|pid}`. Currently, we ask developers to defer
```

rfcs/regex_feature.md
```diff
@@ -0,0 +1,91 @@
...
+[{...}, {...}, {'regex_match|type:account_type:1', ...}]
+```
+
+## Problems with this approach
+
+- The processing time during inference will increase significantly since each input token has to be matched against 
+each regex pattern for every entity type.
+- The training time will also increase for the similar reason
+- The feature could cast a wide net on every token, for example `.*` would match against all tokens. However since the 
+classifier would able to learn from this high noise signal, it would reduce it's dependence on it to make a classification.
```
jre21, line 78:
```suggestion
classifier would be able to learn from this high noise signal, it would reduce it's dependence on it to make a classification.
```

rfcs/regex_feature.md
```diff
@@ -0,0 +1,91 @@
...
+
+```text
+[{...}, {...}, {'regex_match|type:account_type:1', ...}]
+```
+
+## Problems with this approach
+
+- The processing time during inference will increase significantly since each input token has to be matched against 
+each regex pattern for every entity type.
+- The training time will also increase for the similar reason
```
vembar, line 76: What if the user just wants regex matching for an entity (which seems like the most common usecase to me)? If my understanding is right, in the current setup, the user would end up creating a model that only uses regex matching as a feature. That seems redundant? Maybe the user can add it either as a feature or a separate key in the entity config. If it's the latter, we just use regex matching as the ER algo.

jre21:
I don't feel like I have enough in-the-weeds experience to weigh the benefits vs drawbacks on this particular proposal, but I'm definitely in favor of the rfc process and glad you're taking the initiative on it.  Overall, available in config but off by default seems like a good balance for a feature that is getting requests from developers but is also resource-intensive.

jre21:
I don't feel like I have enough in-the-weeds experience to weigh the benefits vs drawbacks on this particular proposal, but I'm definitely in favor of the rfc process and glad you're taking the initiative on it.  Overall, available in config but off by default seems like a good balance for a feature that is getting requests from developers but is also resource-intensive.

rfcs/regex_feature.md
```diff
@@ -0,0 +1,89 @@
...
+```text
+\w{4}[0-9]{3}\-[0-9]{3}
+[0-9]{3}\-[0-9]{3}
+```
+
+In the above syntax, the `account_type` entity would match words like `ABCD-123-281` (first pattern) 
+and `123-456` (second pattern).
+
+Finally, the developer would have the ability to turn on/off the regex feature through the model config in 
+config.py using the key,value `regex-seq: True`. The feature will be off by default.
```
KKsharma99, line 44: We should consider providing a warning message if "regex-seq" is off but "regex_matches.txt" files are found. A developer may create regex rules but forget to set "regex-seq" to True since it is off by default.

rfcs/regex_feature.md
```diff
@@ -0,0 +1,89 @@
...
+- The feature could cast a wide net on every token, for example `.*` would match against all tokens. However since the 
+classifier would be able to learn from this high noise signal, it would reduce it's dependence on it to make a classification.
+- There could be overlaps with regex patterns from different entity types, but the classifier would be able to learn from
+those patterns.
+- This approach locks-in the one-language one-mindmeld application idea since these regex patterns will not translate
+across languages.
+
+## Alternative ideas
+
+- Have the regex mapping in `config.py` instead. This idea is a bit confusing since it doesnt localize all the entity
```
KKsharma99, line 86: Would this look like a dictionary with mappings passed in as the value for "regex-seq"?  `"regex-seq" = {"entity": ["\regex_pattern\",...]}`

If so, this list could get very large. I agree that this would be more confusing than using txt files. You may also need additional checks here to see if the entity listed in the mapping is valid.

rfcs/regex_feature.md
vijay120: `What if the user just wants regex matching for an entity (which seems like the most common usecase to me)`: I think this is a valid use-case, but the solution proposed `user can add it either as a feature or a separate key in the entity config` will lock the platform into this singular use-case.

I think users would like to complement existing signals, like bag-of-words, gazetteers etc with the regex matching feature to capture a regular pattern of queries that are not being recognized with the former approaches (without having to add every combination to gaz). 

Maybe the architecture could be smart in a way where if the only feature is the regex feature, we could short-circuit the model building and just use that to label queries but use the models if not. One issue with approach is there could be collisions between the regex tagger and the normal entity tagger.

rfcs/regex_feature.md
```diff
@@ -0,0 +1,89 @@
...
+- There could be overlaps with regex patterns from different entity types, but the classifier would be able to learn from
+those patterns.
+- This approach locks-in the one-language one-mindmeld application idea since these regex patterns will not translate
+across languages.
+
+## Alternative ideas
+
+- Have the regex mapping in `config.py` instead. This idea is a bit confusing since it doesnt localize all the entity 
+logic in the same folder location.
+- Not have this feature since it could be abused and developers will be confused. Moreover, the performance costs could
```
KKsharma99, line 88: If performance is a concern, we can consider using regex matching for entities prior to training instead of during. This could be integrated with the BootstrapAnnotator. 
Pros: Users can specify regex patterns and clearly see what is being labelled before it goes into the classifier, performance is not a major concern since this runs offline
Cons: It would be harder to update or remove regex-based labels

We can also consider making both of these approaches available.

rfcs/regex_feature.md
vijay120: I think this makes sense. One thing to note is that we currently dont warn devs about other entity features like `in-gaz` when the gaz file is present, for example. Ideally, we should just do this intelligently, ie if the `regex_matches.txt`, use it as a feature. But the issue with this is it becomes confusing to developers who don't want to use it even if they have the `regex_matches.txt` file.

rfcs/regex_feature.md
KKsharma99: Actually, it may be better to separate this class out. We could build a RegexAnnotator

rfcs/regex_feature.md
```diff
@@ -0,0 +1,89 @@
...
+
+A feature based on regex pattern matching that MindMeld's NER models can use to detect custom entities. 
+
+- Developers expect NER models to correctly infer that the query `AVET128-321` is a `pid` entity 
+without providing surrounding sentence context to the classifier. This is more relevant in services that
+just use MindMeld for NLP while doing their own dialog management. 
+- Developers also want granular distinctions between entities that occur in the same context, 
+for example, `look up {SKU120|sku}` and `look up {123-ABC-12|pid}`. Currently, we ask developers to defer
+these types of distinction to the business logic in the dialogue manager.
+- DialogFlow, NLTK and other similar platforms have the regex matching feature. Moreover, we already have an
```
tmehlinger, line 11: How well does regex matching perform in those frameworks? We have concerns below about performance but if implementing this puts us on part with similar tools, I don't know if it's that worrisome.

On the other hand, if the performance gap is that broad, we may want to carefully consider how to implement this.

vijay120: We havent done a formal performance analysis between DF and MindMeld, we might want to plan on that in the future.

rfcs/regex_feature.md
vijay120: I dont quite follow the design here. Are you suggest caching word to regex  match signal outputs before training and refer to this map during training?

KKsharma99: Exploring the idea that certain applications may have entities in their training data that they could label much faster with regex. For example, if they wanted to auto annotate emails, urls or phone numbers they could specify a regex rule to catch that. This is slightly different from using Regex matches as a feature as this would be used for faster pre-training labelling.

vijay120: Got it and this makes sense. I think this is possible in the current framework where the developer could create a custom entity config for the entity in question in config.py with the only feature being the regex feature.

PR #280, opened 2021-02-24
added allowed entities rfc

vijay120:
RFC for allowed entities

### DIFF
--- a/rfcs/allowed_entities.md
+++ b/rfcs/allowed_entities.md
@@ -1,0 +1,132 @@
+# Allowed entities feature
+
+This feature allows MindMeld's entity classifier to more precisely tag entities in 
+queries without context.
+
+## Problem
+
+Supposed for an intent `transfer_money`, there are training queries like this following:
+```text
+move money from {savings|account_type} to {checkings|account_type}
+transfer to {savings|account_type} from {checkings|account_type}
+collect money into {savings|account_type}
+```
+
+The entity classifier currently uses the following information to make it's classification:
+- Contextual words
+- Gazetteers
+- Whether it has seen the phrase before (in-vocab)
+
+However, in multi-turn dialogues where MindMeld's NLP service is only used (and not its
+dialogue management), such a conversation is possible:
+```text
+User> Transfer money
+Bot> From which account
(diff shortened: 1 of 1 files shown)


rfcs/allowed_entities.md
```diff
@@ -0,0 +1,132 @@
...
+Here, the query is `John's investment account`, which is a phrase MindMeld has never seen before, so doing this won't help
+since it won't match condition 3: 
+```python
+{
+    'text': "John's investment account",
+    'allowed_intents': ['banking.money_transfer.account_type']
+}
+```
+
+However, if the client knew the list of nicknames to the end-user's account, for eg,
```
vembar, line 84: Nice! This is definitely useful!

rfcs/allowed_entities.md
```diff
@@ -0,0 +1,132 @@
...
+```
+Here, `domain=banking`, `intent=money_transfer` and `entity_type=account_type`.
+
+MindMeld will do the following logic which is different from the current behavior:
+1. Since an additional entity type is provided in the `allowed_intents` field, it will 
+first run the intent's entity classifier against the query.
+1. If the classifier returns `account_type`, so it matches the allowed_intent's entity type,
+MindMeld will return `Checkings` as `account_type` entity, along with all other entities detected.
+1. If the classifier doesnt detect any entities, we attempt to exact match the input phrase 
+against the allowed_intent's entity type's gazetteer entries and all training phrases annotated
```
vembar, line 53: Unsure if we'd want to match against training phrases. For cases like the assistant's call person intent, we have a lot of names that are ASR mistranscriptions added as training phrases which could lead to errors. 

Seems like it's better to be slightly conservative and match against gazetteer + whitelist data for now.

vembar:
Looks great! Very useful feature to have.

rfcs/allowed_entities.md
vijay120: `names that are ASR mistranscriptions added as training phrases which could lead to errors`: Given that the allowed_entities was passed in, doesn't the false positive rate go down since the client explicitly wants to match against such entity annotations?

PR #286, opened 2021-03-26
[RFC] Multi Package Refactor 

jjjacksn:
A proposal for separating mindmeld into many packages.

### DIFF
--- a/rfcs/multiple-packages.md
+++ b/rfcs/multiple-packages.md
@@ -1,0 +1,82 @@
+# MindMeld Multi Package Refactor
+
+The MindMeld framework is a powerful solution to building natural language applications. Over time more and more functionality has been added, at times leveraging third-party dependencies. There are now situations in which this growing list of dependencies is an obstacle. Here are some examples of use cases where it would be preferable to include only a subset of MindMeld functionality (and therefore of the dependencies which it includes):
+
+- developers of a large application decide they want to deploy their NLP models separately from their dialogue management logic so that they can scale them independently.
+- developers of a pre-existing application want to migrate to MindMeld for NLP models but would like to keep their existing dialogue handling system.
+- When starting a small application or proof of concept, developers don't want to bother with gathering training data but would prefer to leverage MindMeld's dialogue handling. (e.g. for a Webex Assistant Skill).
+
+This is a proposal to subdivide MindMeld into multiple packages to allow MindMeld projects to easily be deployed as multiple processes, each minimizing thier dependencies and footprint.
+
+## Overview
+
+To support the aforementioned use cases we will separate the `mindmeld` package into multiple packages which can be added as dependencies directly.
+
+- `mindmeld-core` will contain base data structures used by all MindMeld packages
+- `mindmeld-nlp` will contain code for building and running natural language processing models
+- `mindmeld-dialogue` will contain code for implementing the dialogue logic of MindMeld applications
+- `mindmeld-qa` will contain code for implementing MindMeld's question answerer component.
+- `mindmeld` will contain the above packages as dependencies, and will continue to work as it does now
+
+## Package Names
+There are two options for naming our new mindmeld packages after this refactor: traditional packages or namespace packages.
+
+#### Traditional Packages
+
+- With traditional packages we would give the new packages a `mindmeld_` prefix, e.g. `mindmeld_core` and `mindmeld_nlp`.
+- To use a mindmeld package other than the primary package, a user would use it's specific name, i.e. `from mindmeld_nlp import NaturalLanguageProcessor`
+- It is possible to implement the change while maintaining the existing `mindmeld` package API.
+
+#### Namespace Packages
+
+- With namespace packages all packages would use the `mindmeld` namespace.
+- To use, for example, the nlp package a user would use `from mindmeld.nlp import NaturalLanguageProcessor` -- even if they have not installed the primary `mindmeld` package.
+- It is unlikely that this can be implemented while maintaining the existing package API. This would require a major version bump.
+
+## Code Migration Destinations
+
+Here is an early plan for what code will belong in which of the new packages
+
+- MindMeld Core
+    - `mindmeld.core`
+    - `mindmeld.components._config`
+    - `mindmeld.components.request`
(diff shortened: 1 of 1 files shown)


ritvikshrivastava:
Really like this idea! 😄  

A couple of early thoughts:
1. In addition to these packages, we can have another one called ``mindmeld_beta`` or ``mindmeld_extras`` for any new developments that are a) not necessary to run the MM components and b) are additional useful functionalities to help in application development outside the core components. Such as the annotators, paraphrasers or active learning (on the top of my head). This could also give us an extra space to push out more beta/new research features into the codebase without coupling with the core functionality too much.

2. I think namespace packages could be a good idea even if it leads to a major version bump with the design. This could allow us to present this changed MindMeld package as a new, decoupled version to our existing users as a major release (re-branding the structure and potentially increased decoupled use-cases in a way?) and include any other major design changes we have considered over time (say moving the config and its usage from the current setup to a different format).

jre21:
Thanks for putting this together, it absolutely feels like the right direction to be moving in.

1. I tend to think namespace packages are a cleaner approach than having separate top-level packages, and that if there's going to be a shakeup we should take the time to get everything exactly the way we want it, even if that makes the transition a bit more painful.  My main intuition here is that adding more changes to an already-disruptive process can reduce the likelihood that we need to make further disruptive changes later, and navigating sequentially through multiple breaking changes is usually more of a hassle than just doing it all at once.
2. With that said, we really need to support users through the transition and have a clear migration process.  I'm going to use Tensor flow as

jjjacksn:
So far that's two votes for namespace packages. I would prefer them as well.

One more question is whether to keep everything in a single repo or move to multiple repos. I prefer multiple repos. If we do decide to use multiple repos we should also decide whether we want to migrate from the Cisco org to a MindMeld github org. 

Re @jre21's comment about dependencies -- in addition to bringing them up to date we could also use this as an opportunity to do some pruning. I think there a few things which are redundant.

rfcs/multiple-packages.md
```diff
@@ -0,0 +1,82 @@
...
+#### Traditional Packages
+
+- With traditional packages we would give the new packages a `mindmeld_` prefix, e.g. `mindmeld_core` and `mindmeld_nlp`.
+- To use a mindmeld package other than the primary package, a user would use it's specific name, i.e. `from mindmeld_nlp import NaturalLanguageProcessor`
+- It is possible to implement the change while maintaining the existing `mindmeld` package API.
+
+#### Namespace Packages
+
+- With namespace packages all packages would use the `mindmeld` namespace.
+- To use, for example, the nlp package a user would use `from mindmeld.nlp import NaturalLanguageProcessor` -- even if they have not installed the primary `mindmeld` package.
```
vijay120, line 33: How does installing namespace packages work? Is it something like this:
```
pip install mindmeld.nlp
```

vijay120:
> In addition to these packages, we can have another one called mindmeld-beta or mindmeld-extras for any new developments that are a) not necessary to run the MM components and b) are additional useful functionalities to help in application development outside the core components. Such as the annotators, paraphrasers or active learning (on the top of my head). This could also give us an extra space to push out more beta/new research features into the codebase without coupling with the core functionality too much. This could be similar to the extras we have right now.

Love this idea. My opinion is to name is `mindmeld.research` for experimental features

> With that said, we really need to support users through the transition and have a clear migration process. I'm going to use TensorFlow as the bogeyman here. The TF 1.x -> 2.0 transition was rough enough that many users decided it was simpler to abandon the project entirely and migrate to pytorch. Granted, TF has many other structural problems that also influenced that trend, but we certainly want to think about what we could do to make sure we aren't setting up a similar dynamic.

Yes, we should provide enough fore-warning to the community before shipping this. Documentation for the migration and maybe an early warning banner might be useful? Also circulating this change, along with other breaking changes like scikitlearn update, tf deprecation etc might be helpful.

> About namespace packaging

Count me in as well. Its more intuitive. Would like some clarity on installation instructions and how many package distributions we would have to maintain. Looks like we would need to ship 5-6 `mindmeld.*` distributions to pypi for this to work.

jre21:
If we're serious about pruning dependencies, then migrating to poetry could help with the analysis.  I've often found its lockfile a useful tool for understanding a package's dependency graph.  Also, as of poetry 1.1 and the poetry-core namespace package, the runtime dependencies have become much more lightweight.

rfcs/multiple-packages.md
jjjacksn: I believe it's determined by the package name in setup.py. We could choose `mindmeld-core` or `mindmeld.core`.

One of these package distributions could actually contain more than one subpackage if desired so there it may be better to use `mindmeld-core` as to not imply that we are literally only installing the `mindmeld.core` subpackage.

jjjacksn:
> Would like some clarity on installation instructions and how many package distributions we would have to maintain. Looks like we would need to ship 5-6 mindmeld.* distributions to pypi for this to work.

I was thinking we would keep `mindmeld` as a grab bag of all the "default" mindmeld functionality. This means that `mindmeld-core`, `mindmeld-nlp`, and `mindmeld-dialogue` would be dependencies and installation would be quite simple.

#### Install all the (default) mindmeld things
```
pip install mindmeld
```

#### Install just dialogue related stuff
```
pip install mindmeld-dialogue
```
    
#### Install just nlp related stuff 
```
pip install mindmeld-nlp
```

> In addition to these packages, we can have another one called mindmeld-beta or mindmeld-extras for any new developments that are a) not necessary to run the MM components and b) are additional useful functionalities to help in application development outside the core components. Such as the annotators, paraphrasers or active learning (on the top of my head). This could also give us an extra space to push out more beta/new research features into the codebase without coupling with the core functionality too much. This could be similar to the extras we have right now.

jjjacksn:
I've started some initial work on this and the nitty gritty details are starting to emerge. I've opened a PR (#288) so folks can follow the progress and comment if they like.

vijay120:
@jjjacksn 

Do you think style if better:
```
pip install mindmeld[dialogue]
```
It is similar to the optional dependencies we have to use certain mindmeld features

PR #289, opened 2021-04-06
Bug fix/entity token spans

vijay120:
This PR fixes the token span index of a query from being calculated incorrectly. It also updates the system entity features which would be affected from collisions based on the token span index issue being fixed by creating two features:
1. log count: the log of the total number of system entities that overlap with the current token
2. log average length: the log of the average length of the system entities that overlap with the current token

Most of the mindmeld tests passed with this bug fix, however some of the no-context system entities are not being detected unless some additional no-context training data is added.


mindmeld/models/query_features.py
```diff
@@ -537,13 +537,25 @@ def _extractor(query, resources):
...
+                            feat_seq[i - j][feat_name].append(len(entity.normalized_text))
+                        else:
+                            feat_seq[i - j][feat_name] = [len(entity.normalized_text)]
+
+        for token_features in feat_seq:
+            for feature, value in token_features.items():
+                if feature.endswith('log_len'):
+                    token_features[feature] = math.log(float(sum(value)) / len(value))
+                else:
+                    token_features[feature] = math.log(value)
```
vembar, line 558: Haven't really thought through how this would would affect things, but since we use a feature scaler (defaults to [MaxAbsScaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MaxAbsScaler.html#sklearn.preprocessing.MaxAbsScaler)) anyway, maybe we can use the raw counts?

vembar:
This was a nice catch! Should we also add unit tests for token spans returned for system entities that have non alpha numeric characters in them? (3456$, $23.45, 7:30, 2:30pm, 2.00 am)

mindmeld/core.py
```diff
@@ -37,6 +37,8 @@
     "milliseconds": 0,
 }
 
+WHITESPACE_CHAR = ' '
```
jjjacksn, line 40: Perhaps we should be robust other types of whitespace (tabs, line breaks)

mindmeld/models/query_features.py
```diff
@@ -537,13 +537,25 @@ def _extractor(query, resources):
                                 entity.entity.type, entity.entity.value.get("grain"), j
                             )
                         )
-                        feat_seq[i - j][feat_name] = 1
+                        if feat_name in feat_seq[i - j]:
+                            feat_seq[i - j][feat_name] += 1
+                        else:
+                            feat_seq[i - j][feat_name] = 1
```
jjjacksn, line 543:
```suggestion
                        feat_seq[i - j][feat_name] = feat_seq[i - j].get(feat_name, 0) + 1
```

mindmeld/models/query_features.py
```diff
@@ -537,13 +537,25 @@ def _extractor(query, resources):
...
                         feat_name = "sys_candidate|type:{}|granularity:{}|pos:{}|log_len".format(
                             entity.entity.type, entity.entity.value.get("grain"), j
                         )
-                        feat_seq[i - j][feat_name] = math.log(
-                            len(entity.normalized_text)
-                        )
+                        if feat_name in feat_seq[i - j]:
+                            feat_seq[i - j][feat_name].append(len(entity.normalized_text))
+                        else:
+                            feat_seq[i - j][feat_name] = [len(entity.normalized_text)]
```
jjjacksn, line 551:
```suggestion
                        feat_value = feat_seq[i - j][feat_name] = feat_seq[i - j].get(feat_name, [])
                        feat_value.append(len(entity.normalized_text))
```

mindmeld/models/query_features.py
jjjacksn: Just wanted to note that for the features ending in `log_len` we were already taking the log.

jjjacksn (approved):
Generally looks okay to me. I had some code style suggestions (take 'em or leave 'em).

I think we may need to handle other whitespace characters (tabs, line break) -- unless we are already normalizing those elsewhere into spaces.

mindmeld/core.py
KKsharma99: To make this more robust we can use `string.whitespace` from the `string` library.

`"In Python3, string.whitespace is a pre-initialized string used as string constant. In Python, string.whitespace will give the characters space, tab, linefeed, return, formfeed, and vertical tab."`

[Source](https://www.geeksforgeeks.org/string-whitespace-in-python/#:~:text=whitespace%20is%20a%20pre%2Dinitialized,%2C%20formfeed%2C%20and%20vertical%20tab.&text=Parameters%20%3A%20Doesn't%20take%20any,since%20it's%20not%20a%20function.)

mindmeld/core.py
```diff
@@ -529,9 +531,15 @@ def _get_form_details(query_span, offset, form_in, form_out):
...
             full_text = query.get_text_form(form_out)
             text = span_out.slice(full_text)
-            tok_start = len(full_text[: span_out.start].split())
+            tok_start = 0
+            # The span range is till the span_out or max to the second last char
+            span_range = min(span_out.start, len(full_text) - 1)
+            for idx, current_char in enumerate(full_text[:span_range]):
+                # Increment the counter only if a whitespace follows a non-whitespace
+                next_char = full_text[idx + 1]
+                if current_char != WHITESPACE_CHAR and next_char == WHITESPACE_CHAR:
```
KKsharma99, line 540: Consider using `string.isspace()` from the `string` library. This would check for tabs, spaces, newline etc and returns True if there are only such whitespace characters in the string. With this, you may not need to store the 'WHITESPACE_CHAR' as a constant. [Source](https://www.programiz.com/python-programming/methods/string/isspace)

mindmeld/models/query_features.py
vijay120: @vembar Good point, the maxabs is a default option for the entity rec features. I will try with the raw counts

mindmeld/core.py
vijay120: Great points. @KKsharma99 @serapio  Do we have to worry about what we define as a "token" here when indexing these tokens? For example, for non-whitespace token delims in east asian languages, this method of finding "empty space" chars wouldnt work.

mindmeld/models/query_features.py
vijay120: great! Added.

mindmeld/models/query_features.py
vijay120: thanks, done

mindmeld/models/query_features.py
vijay120: @jjjacksn Maybe I am misunderstanding, but in this line of code, we have taken the log of the expression exactly once, ie the log for this feature has not been taken before this line of code.

mindmeld/core.py
vijay120: Great suggestion Kunal. I am still not sure how this would affect non-empty space tokenized languages though.

vijay120: I implemented a way of accounting for token spans based on the normalized tokens, that should be resilient to non-space tokenizers. Thoughts?

mindmeld/core.py
vijay120: I implemented a way of accounting for token spans based on the normalized tokens, that should be resilient to non-space tokenizers. Thoughts?

KKsharma99: That's a good point. Using the normalized tokens is definitely more robust here as compared to a whitespace delimiter. This should work with the upcoming changes of `character tokenization` and `spacy tokenization` as each individual token can be found in `query._normalized_tokens` as you have in your PR.

mindmeld/models/query_features.py
serapio: MaxAbsScaler does linear scaling onto the interval, whereas the `log(length)` is just meant to balance small differences vs large differences, e.g. the difference between `1` and  `100` scales the odds more like the difference between `100` and `100000000` (three times as long) than the difference between `100` and `10000` (2 chars longer).

serapio (approved):

mindmeld/core.py
KKsharma99: Good point. Using the normalized tokens is definitely more robust here as compared to a whitespace delimiter. This should work with the upcoming changes of `character tokenization` and `spacy tokenization` as each individual token can be found in `query._normalized_tokens` as you have in your PR

vijay120: So after much debugging, I had to revert the normalized token changes. The reason for this is as follows:
1. This function operates on raw, processed and normalized text. So the token index from the raw text is different to that of a normalized text. For example, the text `meet me from 9am...10am` is normalized as `meet me from 9am 10am`, the token index of `10am` in the raw text is 3 and the token index of the normalized text is 4.
2. The split function used here is done here since once we tokenize, we join using a whitespace to represent the normalized text, so we do this `meet me from 9am...10am` -> `[meet, me, from, 9am, 10am]` -> `meet me from 9am 10am`. So any non-empty space tokenization operation will add whitespace by the time this function operates on it.

For the above two reasons, I have reverted to operating on empty space tokens.

mindmeld/models/query_features.py
vijay120: @serapio Thanks for that comment. I think it makes sense to continue using log_len, but for the other feature, the log count , where the count refers to the total number of system entities that overlap with a token, I think the log scaling doesnt help much since the count distribution looks pretty linear.

I did an analysis on the hr_blueprint and banking_assistant between master vs log vs no-log here: https://cisco-my.sharepoint.com/:x:/p/vijayrk/EQJzxhsBPWhHipofNShGEZwBv9rAXW9GlywB4gqTzEvlBQ?e=uBSmke. I don't think there are any clear patterns in the data ie not much difference between the three systems.

vijay120: I am fine with either log or no log for the counts. The only issue if the developer doesnt use maxabs, they might be in trouble, since the log operation is in itself a type of scaler function. Thoughts? @vembar @serapio

vembar: Looking at the results, it does seem like the results are varying slightly based on log or no_log for numeric entities, but yeah, there's no clear pattern. It's also a good point that the developer may choose remove the scaler. So all things considered, makes sense to keep the log operation.

vembar (approved):

PR #293, opened 2021-04-13
Feature/Spacy-Integrated Tokenizer

KKsharma99:
This PR integrates Spacy Language models to be used for tokenization. (This is the 1st of 2 PRs aimed at improving the configurability and usability of MindMeld's tokenizer).

In this PR:
- Users can select from 3 different tokenizers in the Tokenizer config: "WhiteSpaceTokenizer", "CharacterTokenizer" and "SpacyTokenizer".
- Users can specify a specific normalizer including Unicode character normalizers ("ASCIIFold", "NFC", "NFD", "NFKC", "NFKD")
- A "text_processing" module is introduced and contains "spacy_model_factory.py" which handles the creation of Spacy Language models for both the annotators and tokenizers.

Usage:
```
TOKENIZER_CONFIG = {
    "allowed_patterns" : [],
    "tokenizer": "SpacyTokenizer",
    "normalizer": "ASCIIFold"
}

```

"WhiteSpaceTokenizer" is the default tokenizer if one isn't selected. Moreover, if "SpacyTokenizer" is selected then the correct language model will automatically be chosen. If it doesn't already exist in the environment, it will be downloaded automatically.


mindmeld/auto_annotator.py
```diff
@@ -116,7 +111,7 @@ def _get_file_entities_map(self, action: AnnotatorAction):
             raise AssertionError(f"{action} is an invalid Annotator action.")
 
         for rule in rules:
-            pattern = get_pattern(rule)
+            pattern = Annotator._get_pattern(rule)
```
vijay120, line 114: Make `_get_pattern ` -> `get_pattern ` since its publicly accessed here

mindmeld/text_preparation/normalizers.py
```diff
@@ -0,0 +1,222 @@
...
+        Return the ASCII character corresponding to the folding token.
+
+        Args:
+            char: ASCII folding token
+
+        Returns:
+            char: a ASCII character
+        """
+        char_ord = ord(char)
+        if char_ord < ASCII_CUTOFF:
```
vijay120, line 83: Checking if this should be `if char_ord <= ASCII_CUTOFF:`

mindmeld/text_preparation/normalizers.py
```diff
@@ -0,0 +1,222 @@
...
+            return char
+
+        try:
+            return self.ascii_folding_table[char_ord]
+        except KeyError:
```
vijay120, line 88: If we dont have the value in the ascii table, we just use the same char? So technically, some unicode chars wouldnt be folded?

mindmeld/text_preparation/normalizers.py
```diff
@@ -0,0 +1,222 @@
...
+        ) as mapping_file:
+            for line in mapping_file:
+                codepoint, ascii_char = line.split()
+                ascii_folding_table[ord(codepoint)] = ascii_char
+
+        return ascii_folding_table
+
+
+class NFD(Normalizer):
+    """Unicode NFD Normalizer Class."""
```
vijay120, line 127: Well done! Love these classes!

mindmeld/text_preparation/spacy_model_factory.py
```diff
@@ -0,0 +1,133 @@
...
+        return f"{language}_core_{model_type}_{spacy_model_size}"
+
+    @staticmethod
+    def _download_spacy_model(spacy_model_name):
+        """Download Spacy Model.
+
+        Args:
+            spacy_model_name (str): Name of the Spacy NER model (Ex: "en_core_web_sm")
+        """
+        os.system("python -m spacy download " + spacy_model_name)
```
vijay120, line 116: Use subprocesses: https://docs.python.org/3/library/subprocess.html#replacing-os-system (os.system is deprecated)

mindmeld/text_preparation/tokenizers.py
```diff
@@ -0,0 +1,199 @@
...
+            if char.isspace():
+                # if we hit a space, close the token and empty the buffer
+                if token and token_text:
+                    token["text"] = token_text
+                    tokens.append(token)
+                token = {}
+                token_text = ""
+                continue
+            if (
+                token_text and cat[0] != unicodedata.category(token_text[-1])[0]
```
vijay120, line 89: Please make `cat[0] != unicodedata.category(token_text[-1])[0]` more readable. What exactly is  `unicodedata.category` and why do we do this check? I see the statement about non-latin chars below, but its non-obvious how the condition causes this.

mindmeld/text_preparation/normalizers.py
KKsharma99: Yes, that's right since not all characters have an ascii compatible character to replace. This is consistent with the previous implementation

mindmeld/tokenizer.py
```diff
@@ -29,21 +29,52 @@ class Tokenizer:
...
+    def _get_tokenizer(self):
+        """Gets the 'tokenizer' specified in the config.
+
+        Returns:
+            tokenizer (Tokenizer): Tokenizer specified in the config.
+        """
+        tokenizer = self.config.get("tokenizer")
+        if not tokenizer:
+            raise AssertionError("'tokenizer' cannot be None.")
+        elif tokenizer == NoOpTokenizer.__name__:
```
vijay120, line 56: Why do we do this?

mindmeld/tokenizer.py
```diff
@@ -268,8 +276,7 @@ def tokenize(self, text, keep_special_chars=True):
             else:
                 norm_token_text = self.multiple_replace(norm_token_text, self.compiled)
 
-            # fold to ascii
-            norm_token_text = self.fold_str_to_ascii(norm_token_text)
+            norm_token_text = self._normalizer.normalize(norm_token_text)
```
vijay120, line 279: To confirm, this is phase #1, where normalization happens in tokenization. But this will be changed in the future? If so, please add a TODO here about that.

mindmeld/tokenizer.py
KKsharma99: Just to check if "NoOpTokenizer" is being passed in. Should I skip this check instead?

mindmeld/text_preparation/normalizers.py
KKsharma99: The cut off of `\u0080` is hex for 128 and since ASCII only goes up to 127 the current check is correct. 

https://www.garykessler.net/library/ascii.html

mindmeld/text_preparation/tokenizers.py
```diff
@@ -0,0 +1,199 @@
...
+            tokens.append(token)
+            start_index += len(token_text)
+        return tokens
+
+
+class TokenizerFactory:
+    """Tokenizer Factory Class"""
+
+    @staticmethod
+    def get_tokenizer(tokenizer, language=None, spacy_model_size="lg"):
```
serapio, line 180: Why is "lg" the default? Did tokenization quality depend on model size?

mindmeld/tokenizer.py
KKsharma99: Removed

mindmeld/tokenizer.py
```diff
@@ -29,21 +29,52 @@ class Tokenizer:
...
+        """Gets the 'normalizer' specified in the config.
+
+        Returns:
+            normalizer (Normalizer): Normalizer specified in the config.
+        """
+        normalizer = self.config.get("normalizer")
+        if not normalizer:
+            logger.warning("'normalizer' not found in the config.")
+            return NoOpNormalizer()
+        elif normalizer == NoOpNormalizer.__name__:
```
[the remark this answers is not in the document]
KKsharma99: Removed

mindmeld/text_preparation/tokenizers.py
KKsharma99: Good point, looks like Spacy's tokenization is rule based so I'll change this to "sm"

mindmeld/text_preparation/tokenizers.py
```diff
@@ -0,0 +1,199 @@
...
+            tokens (List[str]): List of tokens.
+        """
+        return [text]
+
+
+class CharacterTokenizer(Tokenizer):
+    """A Tokenizer that splits text at the character level."""
+
+    def __init__(self):
+        """Initializes the CharacterTokenizer."""
```
serapio, line 64: Looking at what other tools are calling their tokenizers, I think this should be called a LetterTokenizer, comparable to (though different than) [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-letter-tokenizer.html) and [AllenNLP](https://docs.allennlp.org/main/api/data/tokenizers/letters_digits_tokenizer/). As you suggested, a CharacterTokenizer (which may also be useful) has single characters in each token.

mindmeld/tokenizer.py
```diff
@@ -250,9 +252,7 @@ def tokenize(self, text, keep_special_chars=True):
         Returns:
             list: A list of normalized tokens
         """
-
-        raw_tokens = self.tokenize_raw(text)
-
```
serapio, line 255: Do we intend to keep `tokenize_raw()` defined in this class for backwards compatibility? It is public, but it seems unlikely that people would actually invoke it directly.

mindmeld/text_preparation/spacy_model_factory.py
KKsharma99: Good point, now using subprocess.run()

`        subprocess.run(["python", "-m", "spacy", "download", spacy_model_name], check=True)`

mindmeld/tokenizer.py
```diff
@@ -134,28 +158,6 @@ def __deepcopy__(self, memo):
         # TODO: optimize this
         return Tokenizer(exclude_from_norm=self.exclude_from_norm)
```
serapio, line 159: This isn't going to produce a faithful copy anymore.

mindmeld/text_preparation/tokenizers.py
```diff
@@ -0,0 +1,199 @@
...
+    def tokenize(self, text):
+        """
+        Identify tokens in text and create normalized tokens that contain the text and start index.
+        Args:
+            text (str): The text to normalize
+        Returns:
+            tokens (List[str]): List of tokenized tokens.
+        """
+        tokens = []
+        token = {}
```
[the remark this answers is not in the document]
KKsharma99: Added the following documentation:

`tokens (List[Dict]): List of tokenized tokens which a represented as dictionaries.
                Keys include "start" (token starting index), "end" (token ending index), and
                "text" (token text). For example: [{"start": 0, "text":"hello", "end":4}]`

mindmeld/text_preparation/tokenizers.py
```diff
@@ -0,0 +1,199 @@
...
+                token_text and cat[0] != unicodedata.category(token_text[-1])[0]
+            ) or cat == "Lo":
+                # if we hit a non-Latin character, close the token and restart buffer
+                if token and token_text:
+                    token["text"] = token_text
+                    tokens.append(token)
+                token = {"start": i}
+                token_text = char
+                continue
+            if not token_text:
```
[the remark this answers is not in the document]
KKsharma99: Refactored this function and broke it down into 3 separate steps/functions. This improves readability of how the tokens are being processed:

text = "hi Joe"

1. For each character determine whether it is a "separator" (start to an independent token) "skip" (White space character) or "no_action" (a character following another within a token)

"hi Joe" -> ["separator", "no_action", "skip", "separator", "no_action", "no_action"]

2. With the information in step 1 you can determine the individual token number each character belongs to. (Skip is represented as -1)

"hi Joe" -> [1, 1, -1, 2, 2, 2]

3. Finally these token numbers are used to generate the final tokens.

tokens =[
{"start": 0, "text": "hi", "end":1},
{"start": 3, "text": "Joe", "end":5}
]

mindmeld/text_preparation/tokenizers.py
KKsharma99: Refactored this function and broke it down into 3 separate steps/functions to improve readability.

A new token is start if the character in sequence is one of the following:
1. It is a non-Latin Token
2. The previous token is a different unicode category. We do this because unicode categories tell us if characters are math symbols, numbers, or letters etc. We want to keep adjacent characters with the same type together. For example, numbers like "1234" in Japanese. More details at: https://www.compart.com/en/unicode/category
3. The character follows a space.

The refactor makes these steps more intuitive.

mindmeld/text_preparation/tokenizers.py
KKsharma99: Thanks for these resources! Sure, I'll go ahead and rename this to a LetterTokenizer and also create a separate CharacterTokenizer

mindmeld/tokenizer.py
KKsharma99: In Phase 2, we will be removing tokenize_raw(). Instead each Tokenizer class (Eg "CharacterTokenizer") will have a tokenize() method. The tokenize() method will be called by the (new) TextPreparationPipeline Class.

mindmeld/text_preparation/tokenizers.py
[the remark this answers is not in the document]
KKsharma99: Added "end" keys for all Tokenizers and updated tests accordingly

mindmeld/text_preparation/tokenizers.py
```diff
@@ -0,0 +1,258 @@
...
+            text (str): The text to process and get actions per character.
+        Returns:
+            actions_by_char (List[str]): List of actions corresponding to each character.
+                For example: ["separate", "separate", "skip", "no_action"]
+        """
+        category_by_char = [unicodedata.category(x) for x in text]
+        actions_by_char = []
+        for index, category in enumerate(category_by_char):
+            same_category_as_previous = (
+                index > 0 and category[0] == category_by_char[index - 1][0]
```
vijay120, line 98: Not immediately clear what `category[0]` represents. Seems like it represents a character like `Z`, but not sure why that makes the categories the same. A better check is needed here

mindmeld/text_preparation/tokenizers.py
```diff
@@ -0,0 +1,258 @@
...
+            )
+            if category == UNICODE_SPACE_CATEGORY:
+                action = "skip"
```
vijay120, line 101: I think we can add a enum class for these actions in the `LetterTokenizer `

mindmeld/text_preparation/tokenizers.py
```diff
@@ -0,0 +1,258 @@
...
+    @staticmethod
+    def get_token_num_by_char(actions_by_char):
+        """
+        Get the token number that each character belongs to.
+        Args:
+            actions_by_char (List[str]): List of actions corresponding to each character.
+                For example: ["separate", "separate", "skip", "no_action"]
+        Returns:
+            token_num_by_char (List[str]): Token number that each character belongs to.
+                Spaces are represented as -1. For example: [1,2,2,3,-1,4,-1,5,5,5]
```
vijay120, line 120: I think `None` is a better representation for Space than `-1`

mindmeld/text_preparation/tokenizers.py
```diff
@@ -0,0 +1,258 @@
...
+            tokens (List[Dict]): List of tokenized tokens which a represented as dictionaries.
+                Keys include "start" (token starting index), "end" (token ending index), and
+                "text" (token text). For example: [{"start": 0, "text":"hello", "end":4}]
+        """
+        actions_by_char = LetterTokenizer.get_actions_by_char(text)
+        token_num_by_char = LetterTokenizer.get_token_num_by_char(actions_by_char)
+        return LetterTokenizer.create_tokens(text, token_num_by_char)
+
+    @staticmethod
+    def get_actions_by_char(text):
```
vijay120, line 83: I think `get_actions_by_char ` and `get_token_num_by_char ` can be composed into one function that returns the `token_num_by_char ` representation. I think the action representation doesnt provide further information ie it has a direct map to `token_num_by_char`

mindmeld/text_preparation/tokenizers.py
```diff
@@ -0,0 +1,258 @@
...
+        Identify tokens in text and create normalized tokens that contain the text and start index.
+        Args:
+            text (str): The text to normalize
+        Returns:
+            tokens (List[str]): List of tokenized tokens.
+        """
+        tokens = []
+        token = {}
+        token_text = ""
+        for i, char in enumerate(text):
```
vijay120, line 182: You could do this `for i, char in enumerate(text + ' '):` and then 194-197 will not be required

vijay120 (approved):
LGTM

mindmeld/text_preparation/tokenizers.py
KKsharma99: Good point! Made this change and added a comment explaining the added space.

mindmeld/text_preparation/tokenizers.py
KKsharma99: Makes sense, went ahead and combined these functions into one. This simplified the code and removed the need to maintain "actions" at all. So "actions" have been removed.

mindmeld/text_preparation/tokenizers.py
KKsharma99: Done

mindmeld/text_preparation/tokenizers.py
KKsharma99: After combining the `get_actions_by_char` and `get_token_num_by_char` functions together as suggested, we no longer need to maintain any representation of "actions".

mindmeld/text_preparation/tokenizers.py
KKsharma99: Unicode defines a set of general category values, which is the first letter in the category code. http://www.unicode.org/reports/tr44/#General_Category_Values

The check here is comparing those first letter values to ensure that the letters are of the same category. Should I add a comment to explain this or add variables like "general_category_current_char"?

KKsharma99: Changed to the following:            

```
# General Category is represented by the first letter of a Unicode category.
general_category_current_char = category[0]
general_category_previous_char = (
    category_by_char[index - 1][0] if index > 0 else None
)
same_category_as_previous = (
    general_category_current_char == general_category_previous_char
)
```

serapio:
With these changes, what is the expected method for setting the exclude_from_norm characters? e.g. If the default list is fine except that I need "%" to not be normalized?

mindmeld/text_preparation/tokenizers.py
KKsharma99: Added a separate CharacterTokenizer

mindmeld/tokenizer.py
KKsharma99: Good point. As you mentioned this is because the app_path is not being passed in upon instantiation. To resolve this, I've made two changes. Firstly, the app_path is now being saved as a class attribute and secondly, the app_path is being passed on instantiation.

I tested this out in a notebook and this is now creating the correct internal tokenizer and normalizer objects

`return Tokenizer(app_path=self.app_path, exclude_from_norm=self.exclude_from_norm)`

mindmeld/components/_config.py
```diff
@@ -1102,6 +1104,12 @@ def get_tokenizer_config(app_path=None, exclude_from_norm=None):
         tokenizer_config = getattr(
             _get_config_module(app_path), "TOKENIZER_CONFIG", DEFAULT_TOKENIZER_CONFIG
         )
+        if not tokenizer_config.get("allowed_patterns"):
+            # If allowed_patterns are not provided, use default
+            tokenizer_config["allowed_patterns"] = []
+            tokenizer_config["default_allowed_patterns"] = _get_default_regex(
+                exclude_from_norm
+            )
```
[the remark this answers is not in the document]
KKsharma99: Good point. In this PR, `exclude_from_norm` should continue to work as it did before. In `_init_regex(self)` the `exclude_from_norm` symbols are added to the compiled regex.

`to_exclude = CURRENCY_SYMBOLS + "".join(self.exclude_from_norm)`
  
This now leaves the remaining normalization methods "ASCII Folding", "NFC", "NFD", "NFKC", "NFKD". Since these all normalize at the character level, I have added the following code which skips normalization if the character is in exclude_from_norm:

```
        final_norm_token_text = []
        for c in regex_norm_token_text:
            c = c if c in self.exclude_from_norm else self._normalizer.normalize(c)
            final_norm_token_text.append(c) 
```

This code is temporary since in Phase 2 it will be the user's responsibility to ensure that their Regex rules do not normalize a particular character. Although, this would work for the RegexNormalizer. I might be able to alter the pipeline to normalize around exclude_from_norm characters, for example:

"I got 80% on the test"
normalize("I got 80") +"%" normalize("on the test")

serapio (approved):

PR #294, opened 2021-04-13
Bugfix/Adds missing 'verbose' field to RequestSchema

ritvikshrivastava:
This PR adds the missing 'verbose' attribute to the RequestSchema class. 

Current error, testing on the MindMeld server:
```
error: Bad request {'text': 'when is the movie being telecast', 'context': {}, 'frame': {}, 'history': [], 'params': {}, 'verbose': True} caused error {'verbose': ['Unknown field.’]}
```


vijay120:
Could you please add a quick test case to catch these?

vijay120 (approved):

Issue #220, opened 2020-09-01
QuestionAnswerer load_kb fails on compact JSON file


davidmfrey:
The following situation demonstrates the problem

Consider a file `data.json`
```[{"id":"1","name":"data1"},{"id":"2","name":"data2"}]```

Loading this file with `QuestionAnswerer.load_kb()` will fail because the JSON is compact.  I would suggest adding an option to load a knowledge base by passing in an iterator to allow the user of the library to load the data in whatever method works best for them.  The JSON file loader should also handle compact JSON format.

Issue #219, opened 2020-09-01
QuestionAnswerer interface could be tightened up


davidmfrey:
I was trying out the QuestionAnswerer class and noticed that the interface can be difficult to use.  For example
```
from mindmeld.components import QuestionAnswerer
qa = QuestionAnswerer('.')
qa.load_kb('my_app', 'stores', './data/stores.json')
qa.get(index='stores')
```
The call to `load_kb()` allows the `app_namespace` parameter to be different than the `app_path` parameter for the construction of a `QuestionAnswerer`.  However, the `get()` function does not have a corresponding `app_namespace` parameter and instead relies on resolved namespace from the `app_path` used in the constructor.  This requires the user to have internal knowledge of how the `app_path` parameter is resolved in order to use the QuestionAnswerer functionality.

Harirupa:
I am one of those people who intend to use the Mindmeld blueprint app as a baseline and attempting to customise by adding data and logic specific to our use case. As mentioned by @davidmfrey, do not have internal knowledge of how the app_path parameter is resolved. Local blueprint data was modified as a test but the elesticsearch index still points to the stored elesticsearch index (in spite of using the option clean = True).

murali1996:
Hi @Harirupa, this issue will be resolved as part of PR #297 . For the time being, you can assume that the `app_namespace` is nothing but the app's name in its absolute directory path. Meaning, if your app path is something like `/Users/<username>/Projects/mindmeld_blueprints/food_ordering`, the `app_namespace` is nothing but `food_ordering`. You can pass-in this name as `qa.load_kb('food_ordering', index_name='stores', './data/stores.json')` while loading a KB.

In case you wish to delete an Elasticsearch index and not just do a _clean_ fit, you can use commands like `curl -X DELETE 'http://localhost:9200/_all'` or `curl -XDELETE localhost:9200/food_ordering*`.

Issue #291, opened 2021-04-09
Add a new feature for instructed KB : multiple questions & answers


warrior13reg:
Good afternoon dear colleagues,

I am writing to you with a proposal to make your solution a little bit flexible, especially when we make out "kb" with instructed data.

In your current version, for one id I can add one question and one answer like:

{
"id": "001",
"question": "What is the weather like today?",
"answer"  : "It's fine."
}

But we know that each of us can give different variants of the same question and possible answers to that question. 

If you deploy the following feature it will make your solution more flexible. The example:

{
"id": "001",
"question": (["What is the weather like today?", "Can you tell me about the weather in your place?"]), and so on
"answer": (["The weather is fine", "It's been unbroken sunshine"]) and so
}

Of course, there will be situations that we won't manage to put all possible Q&A for one topic but it will give to a bot more opportunities to maintain topics by giving different variants of answers and they will be more relevant without making out a huge "kb" of instructed data.

I hope you will look through my proposal and find it adequate. 

Have a nice day.

vijay120:
@murali1996 Wanted to get your inputs on this as you work on the entity resolution/QA features

warrior13reg:
Hi, what I have got to do?

murali1996:
Hi @warrior13reg, thanks for the feedback and the proposal. The idea of allowing lists of strings in knowledge base fields definitely can aid in making the database compact. Will try to take this up in [PR 297](https://github.com/cisco/mindmeld/pull/297) or in a follow-up PR.

murali1996:
Support added through PR #297. Note that this functionality cannot be used when using Elasticsearch backed question answerer but can be used with other backend choice. To use this feature, clone the latest MindMeld through [git clone](https://github.com/cisco/mindmeld). For docs related to different QuestionAnswerer backends, [make docs from the clone](https://github.com/cisco/mindmeld) and checkout the docs section- _Working with the Knowledge Base and Question Answerer_.

PR #297, opened 2021-04-22
Feature/Adding QuestionAnswerer usability without Elasticsearch

murali1996:
# Updates

#### New config format

To make the config format more generic, the following format will be used:
```
QUESTION_ANSWERER_CONFIG = {
   "model_type": <native/elasticsearch>
   "model_settings": {
       "query_type": "embedder_text",
       "embedder_type": "bert",
       "embedding_fields": {
           ...
       }
   }
}
```

instead of the old format (although we still support the old format!):
```
QUESTION_ANSWERER_CONFIG = {
   "model_type": "embedder_text"
   "model_settings": {
       "embedder_type": "bert",
       "embedding_fields": {
           ...
       }
   }
   "use_elastic_search": True/False
}
```

#### Naming convenctions
- `NonElasticsearchQuestionAnswerer` now renamed to `NativeElasticSearch`

# Quick start code snippets
#### structured text
```
from mindmeld.components import QuestionAnswerer

QUESTION_ANSWERER_CONFIG = {
    "model_type": "native",
    "model_settings": {
           "query_type":"keyword"
     }
}

qa = QuestionAnswerer(app_path='food_ordering', config=QUESTION_ANSWERER_CONFIG)
qa.load_kb(index_name="restaurants", data_file='food_ordering/data/restaurants.json')
qa.load_kb(index_name="menu_items", data_file='food_ordering/data/menu_items.json', clean=True)

qa.get(index='menu_items', name='fish and chips')
qa.get(index='menu_items', name='fish and chips', _sort='price', _sort_type='asc')

s = qa.build_search(index='menu_items')
s.query(name='fish and chips')
s.sort(field='price', sort_type='asc')
s.execute(size=15)
```

#### unstructured text
```
from mindmeld.components import QuestionAnswerer

QUESTION_ANSWERER_CONFIG = {
    "model_type": "native",
    "model_settings": {
           "query_type": "text"
     }
}

qa = QuestionAnswerer(app_path='hr_assistant')
qa.load_kb(app_namespace='hr_assistant', index_name='faq_data', data_file='hr_assistant/data/hr_faq_data.json')
qa.get(index='faq_data')

query = 'what is phase 2 of the review cycle'
qa.get(index='faq_data', query_type='text', question=query, answer=query, size=1)
```
`TIP`: pass in `elasticsearch` instead of `native` for `"model_type"` and compare the results

#### leveraging semantic embeddings
```
from mindmeld.components import QuestionAnswerer

QUESTION_ANSWERER_CONFIG = {
    "model_type": "native",
    "model_settings": {
        "query_type": "embedder_text",
        "embedder_type": "bert",
        "embedding_fields": {
            "faq_data": ["question", "answer"]
        }
    },
    "use_elastic_search": False
}

qa = QuestionAnswerer(app_path='hr_assistant', config=QUESTION_ANSWERER_CONFIG)
qa.load_kb(app_namespace='hr_assistant', index_name='faq_data', data_file='hr_assistant/data/hr_faq_data.json')
query = 'what is phase 2 of the review cycle'
qa.get(index='faq_data', question=query, answer=query)
```

#### old config format still supported
```
from mindmeld.components import QuestionAnswerer

QUESTION_ANSWERER_CONFIG = {
    "model_type": "embedder",
    "model_settings": {
        "embedder_type": "bert",
        "embedding_fields": {
            "faq_data": ["question", "answer"]
        }
    }
}

qa = QuestionAnswerer(app_path='hr_assistant', config=QUESTION_ANSWERER_CONFIG)
qa.load_kb(app_namespace='hr_assistant', index_name='faq_data', data_file='hr_assistant/data/hr_faq_data.json')
query = 'what is phase 2 of the review cycle'
qa.get(index='faq_data', question=query, answer=query)
```

# What's in this PR?

This PR contains code modifications to allow users to use a QuestionAnswerer without invoking Elastic Search. To achieve that while, we do the following:

- Create a base class called `BaseQuestionAnswerer` and rename currently existing `QuestionAnswerer` class as a derived class with name `ElasticsearchQuestionAnswerer`.
- Create a new derived class for creating a question answerer without Elastic Search (named `NonElasticsearchQuestionAnswerer`)
- Create a factory class named `QuestionAnswerer` with a `.create_question_answerer()` method to facilitate moving towards a factory based model creation with input configs. To maintaining backwards compatibility, we implement two additional methods in the new `QuestionAnswerer` class-- (1) `__new__(...)` and class method `load_kb(...)`. While the former is used to return an appropriate `XxxQuestionAnswerer`, the latter is just used to maintain using the `load_kb` functionality of previous `QuestionAnswerer` through  `load_kb` of renamed `ElasticsearchQuestionAnswerer`.

This PR also contains code modifications to the class methods, variables used within them, allowing users to better access `.get()`, `.build_search()` and `.load_kb()` of either question answerer classes.

# What files change and what are the change?
- `mindmeld/exceptions.py`, `mindmeld/_util.py`, `mindmeld/cli.py` and `mindmeld/components/nlp.py`: Change KnowledgeBaseConnectionError nomenclature to ElasticsearchKnowledgeBaseConnectionError to make it more explicit that it is a ES based error. Similarly EntityResolverConnectionError to ElasticsearchConnectionError.
- `mindmeld/components/_config.py`: Move all ES based configuration vars to `_elastic_helpers.py` so as to keep all ES related resources at one place. This is done mainly as part of the _decoupling_ aspect of the PR. Additionally, did a very minor improvement to _get\_classifier\_config_ method.
- `mindmeld/components/_elasticsearch_helpers.py`: Made elasticsearch python package usage a conditional usage, in the sense, even if someone do not want to install it while not using an ES based question answerer or entity resolver, no errors are thrown. Previously, because the package is loaded globally, ModuleNotFound errors will be thrown when one fails to have this package. 
- `mindmeld/components/entity_resolver.py` and `mindmeld/models/embedder_models.py`: Since the idea for non-ES QA is to re-use resolvers, an Embedder based resolver is to be implemented and not just sentence-BERT resolver. So the major changes here include code modifications to allow for any embedder based entity resolution, with all embedder caching and resource management moved to `embedder_models.py` file. Additionally improved some aspects of the resolver classes, like being able to pass-in entity maps directly as inputs for the _.fit()_ method. 
- `mindmeld/components/question_answerer.py`: This contains the main changes for this PR, wherein a new question Answerer without Elasticsearch dependency is implemented. Simply move from an ES to non-ES QA by specifying `"use_elastic_search": False` in the QA config.
- `mindmeld/models/_util.py`: This is simply a duplicate of _mindmeld/components/\_util.py_ file.
- ` mindmeld/models/helpers.py`: Minor improvement to the _create\_embedder\_model_ method.
- `mindmeld/path.py`: Adding a method to facilitate path creation for saving indices in question answerer module.
- `tests/components/test_question_answerer.py`: Recreates all test that exist for ES based QA for the non-ES scenario too!

So, all in all, the main files modified are:
- [mindmeld/components/question_answerer.py](https://github.com/cisco/mindmeld/blob/master/mindmeld/components/question_answerer.py)
- [mindmeld/components/entity_resolver.py](https://github.com/cisco/mindmeld/blob/master/mindmeld/components/entity_resolver.py)
- [mindmeld/models/embedder_models.py](https://github.com/cisco/mindmeld/blob/master/mindmeld/models/embedder_models.py)

# Phased implementation of the PR

### Phase 1
- [x] existing QuestionAnswerer class re-naming and factory class implementation
- [x] bug-fixes and resolve some existing Issues #219, #220, and #299.
                 #219:  To tighten it up, we now allow `app_namespace` argument in both `.get()` and `.build_search()` methods, along with its already existing allowance in `.load_kb()`. Instead of specifying at multiple places, a developer can now also input it at QuestionAnswerer initialization time so that the initialized `app_namespace` gets used in all the above methods. By passing `app_namespace` as arguments in above methods, the default `app_namespace` value gets overwritten. If `app_namespace` is never provided, it is derived from the `app_path`.
                 #220  Fixed it by modifying some conditions while reading files. Specifically, we do `line.strip().startswith("[")` instead of `line.strip() == "[":` to resolve compact json issue.
                 #299 Fixed the third by adding `TypeError` in the try-catch.
        
### Phase 2:
- [x] Tighten up `self._qa_config` usage. Required because it is repeatedly used in different places although properties such as `._query_type` exist.
- [x] Some improvements in already existing QA code base to eliminate loose ends. Also allow `embedding_fields` to be passed in directly to `.load_kb()` making it easier in cases if you do not wish to pass a whole new config in input but want specific embedding fields to be encoded.
- [x] Modify `entity_resolver.py` and `embedder_models.py` in a way that allows using any embedder model in Entity Resolver and also re-use resolvers instances in QA module for text or embedder based search. This is also a first step towards refactoring ER and QA code-bases with clubbed commonalities.
- [x] Complete implementing `NonElasticsearchQuestionAnswerer` with instance methods `.get()`, `.build_search()`  and `.load_kb()`. This can be called by setting config as follows:
     ```
      QUESTION_ANSWERER_CONFIG = {
          "model_type": "embedder_text",
          "model_settings": {
              "embedder_type": "bert",
              "embedding_fields": {
                  ...
              }
          }
          "use_elastic_search": False
      }
     ```
       - [x] .load_kb()
       - [x] .build_search() -> 
                  - [x] .query()
                  - [x] .filter()
                  - [x] .sort()
       - [x] .get()
We basically try to re-use TFIDF based resolver for resolving each field in the knowledge base. For fields with embedding requirements, we use additional Embedder based entity resolver instance for those. During search execution time, we combine prediction scores of all involved resolvers and rank the documents based on average scores. For combining scores, we first normalize each resolvers' score predictions across all kb objects using min-max normalization and then average out the scores across all resolvers (for each kb).
- [x] Add required unit tests to test the new class

### Phase 3:
- [x] Phase 1 implementation recognizes `date` and `location` fields in KB as simple strings. Also, the only support available is for matching fields which are strings data type and for doing mathematical operations for number data types (such as >=, <=, etc.). In other words, the Phase 1 implementation just takes the fields in KB and process them as strings or numbers. In this phase, we plan to make this efficient like how Elastic search does.
- [x] Add docs for location calculation and date scoring
- [x] Update docs

# Some Next Steps
- Add support for non-English QA, by allowing user to pass in a language argument


mindmeld/models/embedder_models.py
```diff
@@ -14,119 +14,738 @@
...
+
+        similarity_scores_2d = _sim_func(src_vecs, tgt_vecs, as_numpy=True)  # a 2d numpy array
+
+        results = []
+        for similarity_scores in similarity_scores_2d:
+            similarity_scores = similarity_scores.reshape(-1)
+            # Rounding sometimes helps to bring correct answers on to the list of top scored results
+            similarity_scores = np.around(similarity_scores, decimals=2)
+
+            if scores_normalizer:
```
vembar, line 358: For my understanding, is the normalization of similarity scores here just to make the scores sum to 1 i.e, make it more interpretable?

mindmeld/models/embedder_models.py
```diff
@@ -14,119 +14,738 @@
...
+                        _no_sort=False):
+        """Computes the cosine similarity
+
+        Args:
+            src_texts (Union[str, list]): string or list of strings to obtain matching scores for.
+            tgt_texts (list, optional): list of strings that will be matched to.
+                if None, existing cache is used as target strings
+            top_n (int, optional): maximum number of results to populate. if None, equals length
+                of tgt_texts
+            scores_normalizer (str, optional): normalizer type to normalize scores. Allowed values
```
vembar, line 322: Should there be a default for the normalizer?

mindmeld/models/embedder_models.py
```diff
@@ -14,119 +14,738 @@
...
+            convert_to_tensor (bool): If true, you get one large tensor as return. Overwrites any
+                setting from convert_to_numpy
+            device: Which torch.device to use for the computation
+            concat_last_n_layers (int): number of hidden outputs to concat starting from last layer
+            normalize_token_embs (bool): if the (sub-)token embs are to be individually normalized
+        """
+        super().__init__(app_path, **kwargs)
+
+        # init runtime configs for the model
+        self._batch_size = kwargs.get("batch_size", 16)
```
vembar, line 432: Should we standardize the default `batch_size` across all mindmeld features? Currently the default `batch_size` for the paraphraser is 8.

mindmeld/models/embedder_models.py
murali1996: Thats correct! However, we do not use this feature directly in a QA use-case. Because, this feature normalizes across all the synonyms' (whitelist+cnames') scores, however, we need normalization across all doc ids. I left this feature be present so that it can be useful for any future work.

mindmeld/models/embedder_models.py
murali1996: That's a good point. We should. Also, we should standardize the way a developer inputs/modifies batchsize as well as GPU usage. Can a developer do that as part of paraphraser config?

mindmeld/models/embedder_models.py
```diff
@@ -14,119 +14,731 @@
...
+                    isinstance(data["synonyms"], dict)
+                ):  # deprecated format; backwards compatible with ER module code
+                    self.data = {key: data["synonyms_embs"][j] for key, j in data["synonyms"]}
+
+                else:  # deprecated format; backwards compatible with QA module code
+                    if not isinstance(data, dict):
+                        msg = "Unknown data format while loading cache embeddings. " \
+                              "Ignoring loading ..."
+                        logger.error(msg)
+                    self.data = data
```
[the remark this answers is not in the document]
murali1996: So because this `load()` method belongs to a cache class, i think it would make sense to call a load and expect that a cache file is loaded into memory; and for that reason we assign here.

mindmeld/models/_util.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    Loads an attribute from a module or a module itself
+    (check if the module exists before calling this function)
+    """
+    m = importlib.import_module(module_name)
+    if not func_name:
+        return m
+    if func_name not in dir(m):
+        raise ImportError(f"Cannot import {func_name} from {module_name}")
+    return getattr(m, func_name)
```
vijay120, line 40: Where is `getattr` defined?

mindmeld/models/embedder_models.py
```diff
@@ -14,119 +14,738 @@
...
+        msg = f"DeprecationWarning: Use {self.__class__.__name__}.dump_cache() instead of " \
+              f"{self.__class__.__name__}.dump()"
+        logger.warning(msg)
+        self.dump_cache()
+
+    def get_encodings(self, text_list, add_to_cache=True):
         """Fetches the encoded values from the cache, or generates them.
         Args:
             text_list (list): A list of text strings for which to get the embeddings.
+            add_to_cache (bool): If True, adds te encodings to self.cache and returns embeddings
```
vembar, line 227:
```suggestion
            add_to_cache (bool): If True, adds the encodings to self.cache and returns embeddings
```

mindmeld/models/embedder_models.py
```diff
@@ -14,119 +14,738 @@
...
+
         return encoded
 
-    def dump(self):
-        """Dumps the cache to disk."""
-        with open(self.cache_path, "wb") as fp:
-            pickle.dump(self.cache, fp)
+    def add_to_cache(self, superficial_data):
+        """Manually add some embeddings to cache. This method is created to entertain storing
+        superficial text-encoding pairs wherein the encodings are not the encodings of the text
```
vembar, line 246: Can't we be more specific and call it average/max pooled whitelist embeddings? Don't think we plan to modify or provide alternate methods to do this operation anyway.

mindmeld/models/embedder_models.py
```diff
@@ -14,119 +14,738 @@
...
+        itself but a combination of encodings of some list of texts from same embedder model.
+        For example, to add superficial entity embeddings as average of whitelist embeddings
+        in Entity Resolution.
+
+        Args:
+            superficial_data (dict): texts and their correspoonding superficial embeddings as
```
vembar, line 252:
```suggestion
            superficial_data (dict): texts and their corresponding superficial embeddings as
```

mindmeld/components/entity_resolver.py
```diff
@@ -468,44 +150,51 @@ def create_resolver(cls, app_path, entity_type, **kwargs):
...
 
-    def __init__(self, app_path, entity_type, resource_loader=None):
+    def __init__(self, app_path, entity_type, resource_loader=None, **_kwargs):
         """Initializes an entity resolver"""
         self.app_path = app_path
         self.type = entity_type
-        self._resource_loader = (
-            resource_loader or ResourceLoader.create_resource_loader(app_path=self.app_path)
-        )
+        self._resource_loader = resource_loader
```
vijay120, line 171: `self._resource_loader = resource_loader or ResourceLoader.create_resource_loader(app_path=self.app_path)`

mindmeld/components/entity_resolver.py
```diff
@@ -542,43 +231,120 @@ def _process_entity_map(entity_type,
...
+        new_entities = []
+        id_counter = 0
+        for ent in entity_map.get("entities", []):
+            cname = ent.get("cname", None)
+            if not cname:
+                continue
+            _id, whitelist = ent.get("id", None), ent.get("whitelist", [])
+            if not _id:
+                _id = str(id_counter)
+                id_counter += 1
```
vijay120, line 273: does the id have some relative ordering? I ask since there might be instances when `ent.get("id")` is not None and return 42 while the next ent is None and the `id_counter ` is 42. In that case, we will have two entities with the same id. If ordering is not important, can we just gen a UUID?

mindmeld/components/entity_resolver.py
```diff
@@ -542,43 +231,120 @@ def _process_entity_map(entity_type,
...
+            logger.warning(msg)
+            return results
+
+        # Obtain top scored for each doc id, if scores field exist in results
+        best_results = {}
+        for result in results:
+            if "score" not in result:
+                return results
+            _id = result["id"] if "id" in result else result["cname"]
+            if _id not in best_results or result["score"] > best_results[_id]["score"]:
```
vijay120, line 308: its not exactly clear why `_id not in best_results` returning `True` would lead to `best_results[_id] = result`

mindmeld/components/entity_resolver.py
```diff
@@ -1544,40 +1121,136 @@ class TfIdfSparseCosSimEntityResolver(EntityResolverBase):
...
+                        _no_sort=False):
+        """Computes sparse cosine similarity
+
+        Args:
+            src_texts (Union[str, list]): string or list of strings to obtain matching scores for.
+            top_n (int, optional): maximum number of results to populate. if None, equals length
+                of self._syn_tfidf_matrix
+            scores_normalizer (str, optional): normalizer type to normalize scores. Allowed values
+                are: "min_max_scaler", "standard_scaler"
+           _return_as_dict (bool, optional): if the results should be returned as a dictionary of
```
vijay120, line 1186: Nit: Whats the standard you use to add the prefix `_` to the arg?

mindmeld/components/entity_resolver.py
murali1996: That's a good catch. I didn't realize IDs can be simple integer numbers. It is important that ids are unique and are not same across two docs. I'll make required changes to tackle this.

mindmeld/components/entity_resolver.py
```diff
@@ -542,43 +231,120 @@ def _process_entity_map(entity_type,
...
+        results = [*best_results.values()]
+        n_scores = len(results)
+        if n_scores < top_n:
+            msg = f"Retrieved only {len(results)} entity resolutions instead of asked " \
+                  f"number {top_n} for entity type {self.type}"
+            logger.info(msg)
+            return results
+        #  n_scores >= top_n
+        _sim_scores = np.asarray([val["score"] for val in results])
+        _top_inds = _sim_scores.argpartition(n_scores - top_n)[-top_n:]
```
vijay120, line 321: Its hard to parse this operation. If one wants to sort top_n, I would think we could do this:

```
_top_inds = _sim_scores.partition(top_n)

or 

_top_inds = _sim_scores[_sim_scores.argpartition(top_n)]
```

mindmeld/components/entity_resolver.py
```diff
@@ -603,18 +372,17 @@ def fit(self, clean=False):
         self.ready = True
         return
 
-    @abstractmethod
-    def _predict(self, nbest_entities, top_n):
-        raise NotImplementedError
-
-    def predict(self, entity, top_n: int = 20):
+    def predict(self, entity, top_n=20):
```
vijay120, line 375: Not sure why this line was changed `def predict(self, entity, top_n: int = 20):` seems ok

mindmeld/components/entity_resolver.py
murali1996: I added it to differentiate it from other args in the sense that the args with underscore are not to be used generally except under special use-cases (underscore to emphasize their non-generic usage) and might possibly lead to unexpected outputs.

mindmeld/components/entity_resolver.py
```diff
@@ -1146,231 +919,94 @@ def _load(self):
...
+        settings = er_config.get("model_settings", {})
+        self._aug_lower_case = settings.get("augment_lower_case", False)
+        self._aug_title_case = settings.get("augment_title_case", False)
+        self._aug_normalized = settings.get("augment_normalized", False)
+        self._aug_avg_syn_embs = settings.get("augment_average_synonyms_embeddings", True)
 
-        # [n_syns, emd_dim] -> [n_entities, n_syns, emd_dim]
-        t_syn_enc = _torch("as_tensor", synonyms_encodings)
-        t_syn_enc = t_syn_enc.expand([n_entities, *t_syn_enc.shape])
+        cache_path = settings.get("cache_path", None)
```
vijay120, line 954: None is redundant

mindmeld/components/entity_resolver.py
```diff
@@ -1381,28 +1017,12 @@ def _predict(self, nbest_entities, top_n):
...
-                    normalize_token_embs=self._runtime_configs["normalize_token_embs"],
-                )
-            )
-        top_entity_emb = top_entity_emb.reshape(1, -1)
 
         try:
-            sorted_items = self._compute_cosine_similarity(
-                synonyms, synonyms_encodings, top_entity_emb, top_n
-            )
+            sorted_items = self._embedder_model.find_similarity(top_entity.text, _no_sort=True)
```
vijay120, line 1025: Is it better to add a double negative like `_no_sort` vs `sort`?

mindmeld/components/entity_resolver.py
```diff
@@ -1544,40 +1121,136 @@ class TfIdfSparseCosSimEntityResolver(EntityResolverBase):
...
+                their scores, else a list of sorted synonym names paired with their
+                similarity scores (descending order)
+        """
+
+        is_single = False
+        if isinstance(src_texts, str):
+            is_single = True
+            src_texts = [src_texts]
+
+        top_n = self._syn_tfidf_matrix.shape[0] if not top_n else top_n
```
vijay120, line 1202: `top_n = len(self._syn_tfidf_matrix) if not top_n else top_n` should suffice, no?

mindmeld/components/entity_resolver.py
```diff
@@ -1693,6 +1368,63 @@ def _load(self):
...
+        }
+
+        # determine cache path and pass it in `model_settings`
+        er_config = copy.deepcopy(kwargs.get("er_config", {}))
+        model_settings = er_config.get("model_settings", {})
+        for key, value in model_settings.items():
+            if key in defaults:
+                defaults.update({key: value})
+        string = json.dumps(defaults, sort_keys=True)
+        hashid = f"{Hasher(algorithm='sha1').hash(string=string)}$synonym_{kwargs['entity_type']}"
```
vijay120, line 1419: lets use sha256 since it provides a lesser attack surface for hash collisions

mindmeld/components/question_answerer.py
```diff
@@ -15,77 +15,148 @@
...
+                collisions between the indices of this app and those of other apps.
+            config (dict, optional): The QA config if passed directly rather than loaded from the
+                app config
+            resource_loader (ResourceLoader, optional): An object which can load resources for the
+                question answerer.
+        """
+        self.app_path = app_path
+        self.app_namespace = kwargs.get("app_namespace", None)
+        self._resource_loader = kwargs.get("resource_loader", None)
+        self.__qa_config = kwargs.get("config", None)
```
vijay120, line 90: Why a double under?

mindmeld/components/question_answerer.py
```diff
@@ -94,14 +165,55 @@ def _es_client(self):
...
+        Args:
+            index (str): index name.
+        """
+
+        # load field info from local cache
+        index_info = self._es_field_info.get(index, {})
+
+        if not index_info:
+            try:
+                # TODO: move the ES API call logic to ES helper
```
vijay120, line 180: It would be good to finish this todo in this PR

mindmeld/components/question_answerer.py
```diff
@@ -201,138 +316,117 @@ def build_search(self, index, ranking_config=None):
...
-                .get(index_name, [])
-            )
 
-        def _doc_data(data_file):
+        def _doc_data_count(data_file):
             with open(data_file) as data_fp:
                 line = data_fp.readline()
                 data_fp.seek(0)
-                if line.strip() == "[":
+                # fix related to Issue 220: https://github.com/cisco/mindmeld/issues/220
```
vijay120, line 428: great!

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +528,2178 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+
+            # dump
+            metadata_dump = self._create_metadata_for_dumping(index_name)
+            metadata_dump.update({"__all_ids": index_all_ids})
+            opfile = open(cache_path, "wb")
+            pickle.dump(metadata_dump, opfile)
+            opfile.close()
+
+        def is_available(self, index_name):
+            if (
```
vijay120, line 1664: maybe `return index_name in self._indices or os.path.exists(self._get_index_cache_path(index_name))`?

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +528,2178 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+            if query_type:
+                query_type = None
+
+            field = kwargs.pop("field", None)
+            gt = kwargs.pop("gt", None)
+            gte = kwargs.pop("gte", None)
+            lt = kwargs.pop("lt", None)
+            lte = kwargs.pop("lte", None)
+            et = kwargs.pop("et", None)  # NEW! support equals-to filter; similar to above
+            boolean = kwargs.pop("boolean", None)  # NEW! support boolean filter; input True/False
```
vijay120, line 1719: 😎

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +528,2178 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+            for field_name, kwargs in self._search_queries.items():
+                query_type = kwargs["query_type"]
+                value = kwargs["value"]
+                field_resource = index_resources.get(field_name, None)
+                if not field_resource:
+                    msg = f"The field `{field_name}` is not available in index `{self.index_name}`."
+                    logger.error(msg)
+                    raise ValueError("Invalid knowledge base field '{}'".format(field_name))
+                this_field_scores = field_resource.do_search(query_type, value)
+                scores = {_id: (scores.get(_id, 0.0) + _score) / (n_scores + 1)
```
vijay120, line 1780: Seems like a Counter dict collection might be more intuitive here

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +528,2178 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+                has_text_resolver=self.has_text_resolver,
+                has_embedding_resolver=self.has_embedding_resolver
+            )
+            return cache_object
+
+        @staticmethod
+        def auto_string_processor(string_or_strings, query_type):
+
+            try:
+                english_stop_words = set(nltk_stopwords.words('english'))
```
vijay120, line 1893: this function assumes english language, but this wouldnt suffice for spanish/japanese, no?

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +528,2178 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+            def keyword_processor(string):
+                # TODO: can add char_filters like Elasticsearch; see `keyword_match_analyzer` in
+                #       `components/_elasticsearch_hepers.py`
+                return strip_accents(lowercase(str(string)))
+
+            def text_processor(string):
+                string = strip_accents(lowercase(str(string)))
+                word_tokens = nltk_word_tokenize(string)
+                filtered_string = " ".join(
+                    [english_stemmer.stem(w) for w in word_tokens if w not in english_stop_words])
```
vijay120, line 1921: assumes english here

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +528,2178 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+            # assumes both args are of format "37.77,122.41" where comma seperated lat and lon
+
+            R = 6373.0
+            lat1, lon1 = [float(ii) for ii in some_location.split(",")]
+            lat2, lon2 = [float(ii) for ii in origin_location.split(",")]
+
+            dlon = radians(lon2) - radians(lon1)
+            dlat = radians(lat2) - radians(lat1)
+
+            a = sin(dlat / 2) ** 2 + cos(lat1) * cos(lat2) * sin(dlon / 2) ** 2
```
vijay120, line 2004: This is a bit hard to understand, documentation here would be good.

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +528,2178 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+            """
 
-        def validate(self):
+            if not id2value and not self.data_type:
+                return  # else, if required, update resolvers
+
+            # update id2cname and compute hash
+            for _id, value in id2value.items():
+                # ignore null values
+                if not isinstance(value, bool) and not value:
```
vijay120, line 2107:
```suggestion
                if not value and not isinstance(value, bool):
```

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +528,2178 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+        @staticmethod
+        def min_max_normalizer(list_of_numbers):
+            _min = min(list_of_numbers)
+            list_of_numbers = [num - _min for num in list_of_numbers]
+            _max = max(list_of_numbers)
+            _max = 1.0 if not _max else _max
+            list_of_numbers = [num / _max for num in list_of_numbers]
+            return list_of_numbers
+
+        def determine_data_type(self, value):
```
vijay120, line 2019: update_data_type

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +528,2178 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+                #    in docs and sort them
+                # _scores will all be zeros in case no similarity match is computed before sort
+                inds, common_ids, _scores = zip(
+                    *[(ii, doc["_id"], doc["_score"]) for ii, doc in enumerate(curated_docs) if
+                      doc["_id"] in self.id2value])
+                field_value_scores = [self.number_scorer(self.id2value[_id]) for _id in common_ids]
+
+                _scores = self.min_max_normalizer(_scores)
+                field_value_scores = self.min_max_normalizer(field_value_scores)
+                final_scores = [0.5 * a + 0.5 * b for a, b in zip(_scores, field_value_scores)]
```
vijay120, line 2481: why do we do this?

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +528,2178 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+
+                curated_docs = [curated_docs[ii] for ii in inds]
+                for jj, (doc, final_score) in enumerate(zip(curated_docs, final_scores)):
+                    doc.update({"_sort_score": final_score})
+                    curated_docs[jj] = doc
+
+                return sorted(curated_docs, key=lambda x: x["_sort_score"],
+                              reverse=sort_type != "asc")
+
```
vijay120, line 2490: I think a lot of this code can be consolidated as follows:

```


                transform_func = self.date_scorer if self.data_type == "date" else ...

                inds, common_ids, _scores = zip(
                    *[(ii, doc["_id"], doc["_score"]) for ii, doc in enumerate(curated_docs) if
                      doc["_id"] in self.id2value])
                field_value_scores = [transform_func(self.id2value[_id]) for _id in common_ids]

                _scores = self.min_max_normalizer(_scores)
                field_value_scores = self.min_max_normalizer(field_value_scores)
                final_scores = [0.5 * a + 0.5 * b for a, b in zip(_scores, field_value_scores)]






```

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +528,2178 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+                list[dict]: compiled docs
+            """
+            docs = {}
+
+            ids_are_curated = False
+            all_ids = _ids
+            if not all_ids:
+                ids_are_curated = True
+                all_ids = set()
+                for field_name, field_resource in index_resources.items():
```
vijay120, line 2561:
```suggestion
                for _, field_resource in index_resources.items():
```

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +528,2178 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+                    all_ids.update(field_resource.doc_ids_in_this_field)
+
+            if (not ids_are_curated) and _scores:
```
vijay120, line 2564:
```suggestion
            if not ids_are_curated and _scores:
```

mindmeld/models/embedder_models.py
murali1996: Good point. I'll do that change.

mindmeld/components/entity_resolver.py
murali1996: Replaced this snippet with following in QA module as well
```python
            _id = doc.get("id", None)
            if _id in all_ids:
                msg = f"Found a duplicate id {_id} while processing {data_file}. "
                _id = uuid.uuid4()
                msg += f"Replacing it with a new id: {_id}"
                logger.warning(msg)
            if not _id:
                _id = uuid.uuid4()
                msg = f"Found an entry in {data_file} without a corresponding id. " \
                      f"Creating a random new id ({_id}) for this object."
                logger.warning(msg)
```

mindmeld/components/entity_resolver.py
murali1996: When `_id not in best_results`, it means that `_id` is seen for the first time in the for-loop and hence, a field with that `_id` does not already exist in  `best_results`, thereby leading to KeyError if we directly do `best_results[_id] = result`.

mindmeld/components/entity_resolver.py
murali1996: `top_n` (integer) here actually means number of top items to be collected. Not an array! I think it make sense to change this variable to `size` to avoid confusion. 

Because the partition and argpartition funcs move smaller values to left side of array index=top_n-1, I move all scores less than top_n to left side and retain all top_n scores from array index top_n to end of array. And then simply collect the last top_n indices of array. Let me know if it makes sense; else I'll breakdown the op.

mindmeld/components/entity_resolver.py
murali1996: So the use-case is to avoid sorting; sorting happens by default unless _no_sort is called. I felt it makes sense to call it _no_sort for that reason only.

mindmeld/components/entity_resolver.py
murali1996: So sparse matrices don't have `len` attribute. So used shape instead.

mindmeld/components/entity_resolver.py
murali1996: Good point. Will do that change!

mindmeld/components/question_answerer.py
murali1996: Just to ensure it is not used in derived classes. One must use `_query_type` or `_query_settings` properties of base class instead. Helps to keep things tractable I feel. Also, using `_query_type` is safer as it avoids unnecessary errors due to unknown query_type strings.

mindmeld/components/question_answerer.py
murali1996: It's a TODO from previous version of code. Sure, I'll try to see how to resolve it.

mindmeld/components/question_answerer.py
murali1996: Very good point. Even in Elasticsearch, we assume text is in English. I think it's good idea to mention this note, as we can look into extending support for non-english QA in future. I think it is better not to add this feature in this PR to just keeps things tractable!

mindmeld/components/question_answerer.py
murali1996: ref.: <https://github.com/cisco/mindmeld/pull/297#discussion_r632737580>

mindmeld/components/question_answerer.py
murali1996: Consciously put it in reverse order so as to first check if it is a bool, then check for `None`. If we make it other way round and if `value=False` (a meaningful input!), the data is not recorded and ignored. In the way I initially wrote, the input is only ignored if the input is not a boolean and then if it's a null value. Does that make sense?

mindmeld/components/question_answerer.py
murali1996: @mhuang2173, @vijay120, @vembar :  I am not fully sure what part of code is required to me moved to helpers from here. Can you take a quick look and give a suggestion. Thanks!

mindmeld/models/embedder_models.py
vembar: Currently the user can only modify the batch size

murali1996: @vembar I modified it in this PR. Just for future reference, there is a `batch_size` param in lstm.py that uses a value of 20!

mindmeld/components/question_answerer.py
murali1996: Just to confirm, did you suggest to use [Counter](https://docs.python.org/3/library/collections.html#collections.Counter). If so, it does not support floats as values i think. let me know if I got it wrong.

mindmeld/components/question_answerer.py
murali1996: Edited, and we no more do such averaging.

mindmeld/components/question_answerer.py
murali1996: https://github.com/cisco/mindmeld/pull/297#discussion_r635701465

mindmeld/components/question_answerer.py
murali1996: Added!

mindmeld/components/question_answerer.py
vembar: Currently, we use `english analyzer` in the ES version of the QuestionAnswerer for unstructured text which does a similar processing. If we want to support more languages, we'd have to look into this on the ES side as well.

mindmeld/components/question_answerer.py
```diff
@@ -451,803 +529,2174 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+                          f"({self.processor_type}) vs inferencing ({processor_type}) for the " \
+                          f"field {self.field_name} in index {self.index_name}"
+                    logger.warning(msg)
+                new_value = self.auto_string_processor(value, processor_type)
+                this_field_scores, n_scores = \
+                    update_scores(self._text_resolver, new_value, this_field_scores, n_scores)
+            elif ("text" in query_type or "keyword" in query_type):
+                msg = f"No text based resolver configured for field {self.field_name} " \
+                      f"in index {self.index_name}."
+                logger.warning(msg)
```
vembar, line 2344:
```suggestion
                if ("text" in query_type or "keyword" in query_type):
                     if self._text_resolver:
                         processor_type = "text" if "text" in query_type else "keyword"
                         if processor_type != self.processor_type:
                             msg = f"Using different text processings during loading KB " \
                                         f"({self.processor_type}) vs inferencing ({processor_type}) for the " \
                                         f"field {self.field_name} in index {self.index_name}"
                             logger.warning(msg)
                           new_value = self.auto_string_processor(value, processor_type)
                           this_field_scores, n_scores = \
                                             update_scores(self._text_resolver, new_value, this_field_scores, n_scores)
                       else:
                           msg = f"No text based resolver configured for field {self.field_name} " \
                                       f"in index {self.index_name}."
                           logger.warning(msg)                      
```

vembar, line 2344:
```suggestion
            if ("text" in query_type or "keyword" in query_type):
                if self._text_resolver:
                    # get processor type, process the value and then obtain similarities
                    processor_type = "text" if "text" in query_type else "keyword"
                    if processor_type != self.processor_type:
                        msg = f"Using different text processings during loading KB " \
                              f"({self.processor_type}) vs inferencing ({processor_type}) for the " \
                              f"field {self.field_name} in index {self.index_name}"
                        logger.warning(msg)
                    new_value = self.auto_string_processor(value, processor_type)
                    this_field_scores, n_scores = \
                        update_scores(self._text_resolver, new_value, this_field_scores, n_scores)
            else:
                msg = f"No text based resolver configured for field {self.field_name} " \
                      f"in index {self.index_name}."
                logger.warning(msg)
```

vembar:
Will do another pass soon, but otherwise looking good. Nice work! Personally felt some variable names could be change to something more specific like `dummy_encoding` -->`average_encoding` or `origin_location` --> `query_location`. That's just my opinion, not a change requirement.

mindmeld/components/question_answerer.py
murali1996: Yes. I added TODO tag in both question_answerer.py as well as in elasticsearch_helpers.py files about multilingual support. Will also add it as _Next Steps_ at the end of this PR's description.

mindmeld/components/question_answerer.py
```diff
@@ -449,805 +591,2172 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+            scores = {}
+            for field_name, kwargs in self._search_queries.items():
+                query_type = kwargs["query_type"]
+                value = kwargs["value"]
+                field_resource = index_resources.get(field_name)
+                if not field_resource:
+                    msg = f"The field '{field_name}' is not available in index '{self.index_name}'."
+                    logger.error(msg)
+                    raise ValueError("Invalid knowledge base field '{}'".format(field_name))
+                this_field_scores = field_resource.do_search(query_type, value)
```
mhuang2173, line 1831: not sure if i'm missing something but we probably should apply filters first (if specified) so that we only need to compute scores on a narrower set?

mindmeld/components/question_answerer.py
```diff
@@ -449,805 +591,2172 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+            return get_classifier_config("question_answering", app_path=app_path)
+        return config
+
+    @staticmethod
+    def _get_question_answerer(config):
+
+        # TODO: Both QA classes assume input text is in English.
+        #  Going forward, this should be configurable and multilingual support should be added!
+
+        use_elastic_search = config.get("use_elastic_search", True)
```
mhuang2173, line 2668: i'm wondering if we should make this config more generic? This is related to another comment i have on the naming of the non-Elasticsearch based implementation. For example, if we were to add another implementation/approach this config param could become a bit confusing?

mindmeld/components/entity_resolver.py
```diff
@@ -1693,6 +1381,63 @@ def _load(self):
...
+            "pretrained_name_or_abspath": "distilbert-base-nli-stsb-mean-tokens",
+            "bert_output_type": "mean",
+            "quantize_model": True,
+            "concat_last_n_layers": 4,
+            "normalize_token_embs": True,
+        }
+
+        # determine cache path and pass it in `model_settings`
+        er_config = copy.deepcopy(kwargs.get("config", {}))
+        model_settings = er_config.get("model_settings", {})
```
mhuang2173, line 1427: my understanding is that it would be possible to specify other embedder model to use in the `embedder_type` param and not necessarily BERT for the resolver, while this resolver model class is BERT specific?

mindmeld/components/entity_resolver.py
```diff
@@ -593,7 +377,7 @@ def fit(self, clean=False):
             return
 
         # load data: list of canonical entities and their synonyms
-        entity_map = self._load_entity_map()
+        entity_map = self._format_entity_map(entity_map) or self._load_entity_map()
```
mhuang2173, line 380: looks like we would only format entities when they come from input arg but not when they are loaded from file?

mindmeld/components/entity_resolver.py
```diff
@@ -1544,40 +1135,136 @@ class TfIdfSparseCosSimEntityResolver(EntityResolverBase):
...
+        """
+        string = string.strip()
+        if len(string) == 1:
+            return [string]
+
         results = []
-        # give more importance to starting and ending characters of a word
-        string = f" {string.strip()} "
+        # give importance to starting and ending characters of a word
+        string = f" {string} "
```
vijay120, line 1176: why no strip?

mindmeld/components/entity_resolver.py
```diff
@@ -1659,18 +1348,10 @@ def _predict(self, nbest_entities, top_n):
...
-
-        similarity_scores = self._syn_tfidf_matrix.dot(top_entity_vector.T).toarray().reshape(-1)
-        # Rounding sometimes helps to bring correct answers on to the top score as other
-        # non-correct resolutions
-        similarity_scores = np.around(similarity_scores, decimals=4)
-        sorted_items = sorted(list(zip(self._unique_synonyms, similarity_scores)),
-                              key=lambda x: x[1], reverse=True)
 
         try:
+            sorted_items = self.find_similarity(top_entity.text, _no_sort=True)
```
vijay120, line 1354: If `_no_sort ` is true, why is the var named `sorted_items `?

mindmeld/components/question_answerer.py
```diff
@@ -15,77 +15,231 @@
...
+                  f"distinctly identify the Knowledge Base indices being created. Using the " \
+                  f"default 'app_path' as the current working directory path: '{os.getcwd()}'."
+            logger.warning(msg)
+
+        self._app_path = kwargs.get("app_path") or os.getcwd()  # app_path can be NoneType as well!
+        self.app_namespace = kwargs.get("app_namespace") or get_app_namespace(self._app_path)
+        self._resource_loader = (
+            kwargs.get("resource_loader") or ResourceLoader.create_resource_loader(self._app_path)
+        )
+        self.__qa_config = (
```
vijay120, line 104: why the two dunders?

mindmeld/components/question_answerer.py
```diff
@@ -15,77 +15,231 @@
...
+            _sort (str): Specify the knowledge base field for custom sort.
+            _sort_type (str): Specify custom sort type. Valid values are 'asc', 'desc' and
+                              'distance'.
+            _sort_location (dict): The origin location to be used when sorting by distance.
+
+        Returns:
+            list: A list of matching documents.
+        """
+
+        index_name = self._resolve_deprecated_index_name(index_name, kwargs.pop("index", None))
```
vijay120, line 164: `kwargs.pop("index", None)` -> `kwargs.pop("index")`

mindmeld/components/question_answerer.py
vijay120: Ah got it, if so, thats OK (ie doesnt need to be completed this PR)

mindmeld/components/question_answerer.py
```diff
@@ -449,805 +591,2172 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
-                e.status_code,
-                e.info,
+            return cache_object
+
+        @staticmethod
+        def auto_string_processor(string_or_strings, query_type, language='english'):
+
+            if language != 'english':
+                msg = "Only allowed language for processing in QA is 'english'. " \
+                      "More support coming soon!!"
```
vijay120, line 1956: Nit: Remove `More support coming soon!!`

mindmeld/components/question_answerer.py
```diff
@@ -449,805 +591,2172 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+                >>> point_3, point_4 = "52.2296756,21.0122287", "52.406374,16.9251681"
+                >>> location_scorer(point_3, point_4)
+                >>> # 278.54 kms
+            """
 
-        @abstractmethod
-        def validate(self):
-            """Validate the clause."""
-            raise NotImplementedError("Must override validate()")
+            R = 6373.0
```
vijay120, line 2089: Whats the significance of this constant?

mindmeld/components/question_answerer.py
```diff
@@ -449,805 +591,2172 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+                    if lte and not (value <= lte):
+                        return False
+                    if et and not (value != et):
+                        return False
+                    return True
+
+            elif self.data_type in ["date"]:
+
+                def is_valid(value):
+                    value = abs(self.date_scorer(value))  # returns no. of days
```
vijay120, line 2479: seems like date and number have similar logic that can be collapsed. Also, could you explain what the conditional logic is doing?

vijay120 (approved):
Small nits. Please address Marvin's and Varsha's concerns but LGTM from me.

mindmeld/components/question_answerer.py
```diff
@@ -449,805 +591,2172 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+            return self.type in self.TEXT_TYPES
+
+        def is_vector_field(self):
+            """Returns True if the knowledge base field is a vector field, otherwise returns False
+            """
+
+            return self.type in self.VECTOR_TYPES
+
+
+class NonElasticsearchQuestionAnswerer(BaseQuestionAnswerer):
```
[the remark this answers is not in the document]
murali1996: After some discussion with Varsha, I am thinking to rename `NonElasticsearchQuestionAnswerer` to `QuestionAnswerer`. So `ElasticsearchQuestionAnswerer` will be like a variant of Question Answerer that uses Elasticsearch.

mindmeld/components/question_answerer.py
```diff
@@ -449,805 +591,2172 @@ def _generate_mapping_data(embedder_model, embedding_fields):
...
+        app_namespace = app_namespace or self.app_namespace
+
+        # If an id was passed in, simply retrieve the specified document
+        if doc_id:
+            logger.info(
+                "Retrieve object from KB: index= '%s', id= '%s'.", index, doc_id
+            )
+            s = self.build_search(index, app_namespace=app_namespace)
+
+            if NonElasticsearchQuestionAnswerer.FieldResource.is_number(doc_id):
```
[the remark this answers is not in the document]
murali1996: We expect doc_id to be unique, however, it can be a string or a number. In either case, a filter operation that looks for exact match is possible. In case of string doc_ids, it is also possible to do a fuzzy match instead of exact match. 

For reference, please see [this docstring of filter op in Elasticsearch based QA](https://github.com/cisco/mindmeld/blob/8d21d3dd372c7ef6ac5c55025d2f9273487d9b61/mindmeld/components/question_answerer.py#L676) which mentions `If the entity is not in the canonical form, a fuzzy filter can be applied by setting the query_type to 'text'.` 

So in case of non-Es QA, this fuzzy match is not implemented irrespective of the query type.

mindmeld/components/question_answerer.py
murali1996: Thanks for pointing it out. Makes sense to do filtering first. Will make this change in code.

mindmeld/components/entity_resolver.py
murali1996: modified, we are now formatting both during loading as well as when it is inputted

mindmeld/components/question_answerer.py
murali1996: it is just a constant based on the _Haversine formula_

mindmeld/components/entity_resolver.py
murali1996: Thats correct. Although we advertise BERT based embedder in docs, a generic embedder based class is created to aid QA non-ES class.

source/userguide/kb.rst
```diff
@@ -716,6 +716,38 @@ In the example below, we search for restaurants whose names best match ``firetra
...
+
+  QUESTION_ANSWERER_CONFIG = {
+      'model_type': 'keyword'
+  }
+
+The ``model_type`` can be one of 'keyword' or 'text'. While specifying the former optimizes your QuestionAnswerer search for keywords or short spans of text, the latter optimizes for searching on larger paragraphs or passages of unstructured text. In order to leverage embeddings based semantic matching along with surface-level text features, you can specify one of three model types- 'embedder', 'embedder_text' and 'embedder_keyword'. For leveraging appropriate embeddings, you can specify embedder configurations in the :data:`QUESTION_ANSWERER_CONFIG` with an additional key 'model_settings'. For full details on using embedders in QuestionAnswerer, check the :ref:`Leveraging semantic embeddings <semantic_embeddings>` section below. Note that the specified ``model_type`` is also the default ``query_type`` for all QuestionAnswerer calls in your application, but you can always pass in a different query type to your ``qa.get()`` command as desired.
+
+.. _non_elasticsearch_question_answerer:
+
+QuestionAnswerer without Elasticsearch backend
```
vembar, line 736:
```suggestion
QuestionAnswerer without Elasticsearch 
```

source/userguide/kb.rst
```diff
@@ -716,6 +716,38 @@ In the example below, we search for restaurants whose names best match ``firetra
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Search operations of QuestionAnswerer are by default backed by Elasticsearch. While it is recommended to keep using Elasticsearch backend for QuestionAnswerer, it might not always be feasible to use it due to some or the other constraints. In such cases, one can easily disable Elasticsearch usage by setting a disable-key ``"use_elastic_search": False`` in the ``QUESTION_ANSWERER_CONFIG`` as follows.
```
vembar, line 739:
```suggestion
Search operations of QuestionAnswerer are by default backed by Elasticsearch. While it is recommended to use the Elasticsearch based QuestionAnswerer, it might not always be feasible to use it due to some constraints. In such cases, one can easily disable Elasticsearch by setting a disable-key ``"use_elastic_search": False`` in the ``QUESTION_ANSWERER_CONFIG`` as follows.
```

source/userguide/kb.rst
```diff
@@ -716,6 +716,38 @@ In the example below, we search for restaurants whose names best match ``firetra
...
+
+.. code-block:: python
+
+  QUESTION_ANSWERER_CONFIG = {
+      'model_type': 'keyword',
+      'use_elastic_search': False
+  }
+
+You can set this key irrespective of leveraging embedders or not. If unspecified, this key will be set to ``True``. It is noteworthy that the search results obtained with and without Elasticsearch might have minor differences between them.
```
vembar, line 748: Can we also briefly mention what happens in the case of non-elasticsearch question answerer?

mindmeld/components/question_answerer.py
```diff
@@ -133,1121 +162,2704 @@ def get(self, index, size=10, query_type=None, **kwargs):
...
+            ("app_path" in kwargs and kwargs["app_path"])
+        ):
+            msg = f"Passing 'config' or 'app_path' to '.load_kb()' method is no longer " \
+                  f"supported. Create a {self.__class__.__name__} instance with the required " \
+                  f"configurations and/or app path before calling '.load_kb()'."
+            raise ValueError(msg)
+        self._load_kb(index_name, data_file, **kwargs)
+
+
+class ResolverBasedQuestionAnswerer(BaseQuestionAnswerer):
```
vembar, line 222: I think ResolverBased is confusing to the end-user (if they ever have to debug something) compared to NonElasticSearch. Should we rename it to NativeQuestionAnswerer or something that indicates the ranking happens locally in some sense? @mhuang2173 @vijay120 @murali1996

vembar (approved):
Some small documentation nits. Once we resolve the naming issue, this PR is good to go as far as i'm concerned.

mindmeld/components/question_answerer.py
murali1996: Good point! I'm in for naming it as `NativeQuestionAnswerer`; if we don't have any other suggestions I can go ahead and rename it that way. @mhuang2173 @vijay120 @vembar

source/userguide/kb.rst
murali1996: Sure. Can do that. I have few questions though:    
1) Do you want me to mention that we re-purpose TF-IDF and Embedder Resolvers to do the matching or would that confuse the reader? 
1) I can also mention in which folder are the resources cached for re-loading. That should be good right?
1) In this resolver, users can also pass KBs as list of strings which is currently not allowed for ES based QA. Should we mention that? (ref. [Issue 291](https://github.com/cisco/mindmeld/issues/291)) I can look into how we can entertain list of strings in ES based QA too but that might take bit more time.

@vembar

vembar: I think it's best to not bring in the resolver since that will only confuse the user. Yeah, just the details about where and how it is stored and a one liner or so about the ranking function should be good.

mindmeld/components/question_answerer.py
mhuang2173: I think `NativeQuestionAnswerer` is great since it reflects that it uses native ranking implementation!

mindmeld/components/question_answerer.py
murali1996: update:
the configs will now be passed in as follows:
```
QUESTION_ANSWERER_CONFIG = {
   "model_type": <native/elasticsearch>
   "model_settings": {
       "query_type": "embedder_text",
       "embedder_type": "bert",
       "embedding_fields": {
           ...
       }
   }
}
```

source/userguide/kb.rst
```diff
@@ -726,26 +726,33 @@ To override MindMeld's default QuestionAnswerer configuration with custom settin
...
 
 .. _non_elasticsearch_question_answerer:
 
-QuestionAnswerer without Elasticsearch backend
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+QuestionAnswerer without Elasticsearch
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Search operations of QuestionAnswerer are by default backed by Elasticsearch. While it is recommended to keep using Elasticsearch backend for QuestionAnswerer, it might not always be feasible to use it due to some or the other constraints. In such cases, one can easily disable Elasticsearch usage by setting a disable-key ``"use_elastic_search": False`` in the ``QUESTION_ANSWERER_CONFIG`` as follows.
+Search operations of QuestionAnswerer are by default backed by Elasticsearch. While it is recommended to use the Elasticsearch based for QuestionAnswerer, it might not always be feasible to use it due to some or the other constraints. In such cases, one can easily use a non-Elasticsearch backend by setting `"model_type": "native"` in the ``QUESTION_ANSWERER_CONFIG`` as follows.
```
vembar, line 742:
```suggestion
Search operations of QuestionAnswerer are by default backed by Elasticsearch. While it is recommended to use the Elasticsearch based QuestionAnswerer, it might not always be feasible to use it due to various constraints. In such cases, one can easily disable Elasticsearch by setting `"model_type": "native"` in the ``QUESTION_ANSWERER_CONFIG`` as follows.
```

source/userguide/kb.rst
```diff
@@ -726,26 +726,33 @@ To override MindMeld's default QuestionAnswerer configuration with custom settin
...
+      }
   }
 
-You can set this key irrespective of leveraging embedders or not. If unspecified, this key will be set to ``True``. It is noteworthy that the search results obtained with and without Elasticsearch might have minor differences between them.
+You can set this key irrespective of leveraging embedders or not. If unspecified, this key will be set to ``elasticsearch``. It is noteworthy that the search results obtained with and without Elasticsearch might have minor differences between them.
+
+The native QuestionAnswerer, similar to the Elasticsearch backed QuestionAnswerer, is capable of doing both surface-level text features matching as well as semantic matching based on embeddings. Upon loading a Knowledge Base, the native QuestionAnswerer stores the KB data at the location '/Users/<username>/.cache/mindmeld/.generated/question_answerer/<indexname>.pkl'. If using an embedder, the embeddings of the required data fields will also be saved under the same '.generated' folder. Note that the matching scores of the Elasticsearch backed QuestionAnswerer are larger in magnitude whereas the scores from the native backend are generally fractions less than 1.
```
vembar, line 755: Isn't the `.generated` folder within the app?

mindmeld/components/question_answerer.py
```diff
@@ -17,110 +17,146 @@
...
 logger = logging.getLogger(__name__)
 
-
 DEFAULT_QUERY_TYPE = "keyword"
-ALL_QUERY_TYPES = ["keyword", "text", "embedder", "embedder_keyword", "embedder_text"]
+ALL_QUERY_TYPES = ["keyword", "text", "embedder",
+                   "embedder_keyword", "keyword_embedder",
+                   "embedder_text", "text_embedder"]
 EMBEDDING_FIELD_STRING = "_embedding"
+NON_ELASTICSEARCH_INDICES_STORAGE_PATH = os.path.join(os.path.expanduser("~"), ".cache/mindmeld")
```
vembar, line 76: This should be moved to `path.py`

mindmeld/components/_config.py
```diff
@@ -104,10 +104,30 @@
 
 DEFAULT_ENTITY_RESOLVER_CONFIG = {
     "model_type": "resolver",
-    "model_settings": {"resolver_type": "text_relevance",},
+    "model_settings": {"resolver_type": "text_relevance", },
```
mhuang2173, line 107: minor: is the added space intentional?

source/userguide/kb.rst
```diff
@@ -716,6 +716,45 @@ In the example below, we search for restaurants whose names best match ``firetra
...
+.. code-block:: python
+
+  QUESTION_ANSWERER_CONFIG = {
+      'model_type': 'elasticsearch',
+      'model_settings': {
+          'query_type': 'keyword'
+      }
+  }
+
+The ``query_type`` can be one of ``keyword`` or ``text``. While specifying the former optimizes your QuestionAnswerer search for keywords or short spans of text, the latter optimizes for searching on larger paragraphs or passages of unstructured text. In order to leverage embeddings-based semantic matching along with surface-level text features matching, you can specify one of the three embedder query types- ``embedder``, ``embedder_text``, or ``embedder_keyword``. For using embeddings of your choice, you can specify embedder configurations in the :data:`QUESTION_ANSWERER_CONFIG` within the key 'model_settings'. For full details about using embedders in QuestionAnswerer, check the :ref:`Leveraging semantic embeddings <semantic_embeddings>` section below. Note that the specified ``query_type`` is also the default ``query_type`` for all QuestionAnswerer calls in your application, but you can always pass in a different query type to your ``qa.get(query_type=...)`` command as desired.
```
mhuang2173, line 735: minor: maybe we should re-phrase the first sentence a little bit to indicate that `keyword` and `text` and embedder types can be specified in `query_type` param? Right now it sounds like only `keyword` and `text` are allowed before we talk about embedder types later in the paragraph.

mindmeld/components/question_answerer.py
```diff
@@ -17,110 +17,148 @@
...
+        resolve_es_config_for_version,
+    )
 
 logger = logging.getLogger(__name__)
 
-
 DEFAULT_QUERY_TYPE = "keyword"
-ALL_QUERY_TYPES = ["keyword", "text", "embedder", "embedder_keyword", "embedder_text"]
+ALL_QUERY_TYPES = ["keyword", "text", "embedder",
+                   "embedder_keyword", "keyword_embedder",
```
mhuang2173, line 76: do we need both `embedder_keyword` and `keyword_embedder`? (same for `embedder_text`)

mhuang2173 (approved):
Some final minor comments, otherwise LGTM. Great work!

PR #301, opened 2021-05-07
Training optimizations for memory and time reduction

davidmfrey:
# What does this PR do?
Reworks the storage and flow of data through the MindMeld nlp building process.  This was instigated by a new build restriction that limits the amount of RAM available to the build process in our pipeline.  The end result of this work was tested for memory use and processing time and the results are below.
___
# What are the results?
Memory usage would peak at around 4GB previously.  With these changes, it grows to about 1GB.
Training time has also been positively impacted. 
Previous training times:
- Full build - 22:12
- Incremental build with no changes - 1:40
- Incremental build with a minor change - 6:06

Training times with this PR:
- Full build - 19:21
- Incremental build with no changes - 1:02
- Incremental build with a minor change - 2:45

The decrease in training time can be mostly attributed to more efficient loading and unloading of data from disk to memory.

# What are the major changes?
1. The `QueryCache` has been moved from an in-memory dictionary that is saved to disk to a sqlite3 backed database.
  a. Keys for the queries are now a sha256 hash of the domain, intent, and query string.
  b. ProcessedQuery and all of the classes it contains now implement a `to_cache()` and `from_cache()` function.  These must be updated if new properties are added.
  c. There is no `load()` and `dump()` function for the `QueryCache`.  Every write to the cache is immediately committed to disk and is available for lookup.
2. Accessing a list of queries and labels is done through `resource_loader.ProcessedQueryList`.  This is done in order to minimize the number of queries loaded into memory at any point in time.
  a. `ResourceLoader.file_to_query_info` is now a list of ids rather than the actual `ProcessedQuery` objects.
  b. `ProcessedQueryList` provides a set of iterators for efficient access to various commonly used properties of the `ProcessedQuery` object.  These iterators load the required data from the `QueryCache` as needed.
  c. Some labels are still being resolved into in-memory lists if the data is small and they are iterated over more than once.
  d. Resource extraction has been reworked to only iterate over the `Query` objects in the set once.
  e. Raw file loading no longer exists.  Instead, there is a `ProcessedQueryList.RawQueryIterator` provided.
3. Some particularly memory-hungry features used by the CRF model now use a `FileBackedList` to aggregate and iterate over these features.
  a. A temporary file is created and the `append()` function writes a JSON serialization of the data to the file.
  b. A `FileBackedList.Iterator` is provided to iterate over the contents of the list.  It loads the data from the temporary file as requested.  
  c. The temporary file is removed when the `FileBackedList` is garbage collected.
4. Models are no longer kept in memory during training.   
  a. An `unload()` method has been added to models which is responsible for clearing any references to memory.
  b. The `fit()` method now returns `True` if the model was loaded into memory, `False` if a cached model is available.  The cached model is no longer loaded into memory by the `fit()` method.
  c.  Models are built, then written to disk and unloaded during training so only the model that is currently training is in memory.
  d. All model are loaded into memory at the end of `nlp.build()` so the `NaturalLanguageProcessor` is ready for inference.
5. I removed the `queries` parameter from the classifiers.  This parameter is not documented externally and was leading to confusion in the data flow.
6. Unit tests have been updated to match the changes.


mindmeld/components/classifier.py
```diff
@@ -181,6 +179,10 @@ def fit(self, queries=None, label_set=None, incremental_timestamp=None, **kwargs
                 feature extractor function, or a callable which will be used as to
                 extract features.
 
+        Returns:
+            True if model was loaded and fit, False if a valid cached model exists.  The cached
+            model must be loaded if fit() returns False.
```
vijay120, line 184: What happens when valid cache model exists but the model is not loaded into mem?

mindmeld/components/classifier.py
```diff
@@ -216,15 +218,14 @@ def fit(self, queries=None, label_set=None, incremental_timestamp=None, **kwargs
...
 
-        new_hash = self._get_model_hash(model_config, queries, label_set)
+        new_hash = self._get_model_hash(model_config, label_set)
         cached_model = self._resource_loader.hash_to_model_path.get(new_hash)
 
         if incremental_timestamp and cached_model:
-            logger.info("No need to fit. Loading previous model.")
-            self.load(cached_model)
-            return
+            logger.info("No need to fit.  Previous model is cached.")
```
vijay120, line 225: Does this mean in order to parse queries, one has to do this:
```
nlp.fit()
nlp.load()
nlp.process("xyz")
```

instead of:
```
nlp.fit()
nlp.process("xyz")
```
If this is true, just wondering if the gains we see here can be done without a change in the API behavior since this would be a backwards incompatible change.

mindmeld/components/domain_classifier.py
```diff
@@ -101,49 +105,50 @@ def inspect(self, query, domain=None, dynamic_resource=None):
...
         """
-        query_tree = self._get_query_tree(queries, label_set=label_set)
+        query_tree = self._get_query_tree(label_set=label_set)
         queries = self._resource_loader.flatten_query_tree(query_tree)
 
         if not queries:
             return [None, None]
-        return list(zip(*[(q.query, q.domain) for q in queries]))
+        return (queries.queries(),
+                list(queries.domains()))
```
vijay120, line 142: Dont we want an iterator here to yield a tuple of (query, domain)?

mindmeld/components/entity_recognizer.py
```diff
@@ -141,6 +140,15 @@ def dump(self, model_path, incremental_model_path=None):
         )
         super().dump(model_path, incremental_model_path)
 
+    def unload(self):
+        logger.info(
+            "Unloading entity recognizer: domain=%r, intent=%r", self.domain, self.intent
+        )
+        self.entity_types = None
```
vijay120, line 147: Do we also want to unload the reference to the resource loader?

mindmeld/components/domain_classifier.py
davidmfrey: No, this is returning a tuple of iterators rather than an iterator of tuples.  This behaviour hasn't changed.

mindmeld/components/classifier.py
```diff
@@ -150,12 +149,11 @@ def __init__(self, resource_loader):
         self.config = None
         self.hash = ""
 
-    def fit(self, queries=None, label_set=None, incremental_timestamp=None, **kwargs):
```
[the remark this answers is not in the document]
davidmfrey: Keeping queries ends up adding complexity with the new lazy loading ProcessedQueryList.  Since queries could be a list that a user passes in there would need to be code to check the type before performing operations on it that might fail.

Also, the code path to load the dataset for a classifier is easier to reason about without the queries param being there.  It's always loaded from the resource_loader by the label_set.

mindmeld/components/classifier.py
davidmfrey: I guess I'm not sure what interfaces are external.  The nlp.build function loads all of the models into memory once it completes.   If people use fit() on individual classifiers independently of the nlp.build() method, we could have an option to load it regardless to make it backwards compatible.

mindmeld/components/classifier.py
davidmfrey: I'm not sure what you mean.  The training process will load this later via load()

We could load this, but we end up with extra work
1. load the cached model
2. unload the cached model when training the next one.
3. load the model again when nlp.build completes
4. dump the model again from the main process training it.

vs

1. verify a cached model exists
2. load it when nlp.build() finishes
3. dump it from the main process.

mindmeld/components/entity_recognizer.py
davidmfrey: There are probably some properties retaining memory that I've missed.  Please let me know if you notice any.

mindmeld/core.py
```diff
@@ -229,6 +239,29 @@ def __init__(
         self._timestamp = timestamp
         self.stemmed_tokens = stemmed_tokens or tuple()
 
+    def to_cache(self):
+        return {
+            'raw_text': self.text,
+            'processed_text': self.processed_text,
+            'normalized_tokens': self._normalized_tokens,
+            'char_maps': {},
```
vijay120, line 247: Why are the char maps empty here?

mindmeld/components/classifier.py
davidmfrey: I committed a change to handle this.

mindmeld/core.py
davidmfrey: TODO: find out if char maps are used in training

mindmeld/components/classifier.py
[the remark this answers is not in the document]
vijay120: @KKsharma99  Is there anywhere in the AL codepath that involved directly passing in query objects to fit?
@serapio Similarly, do we have any scripts that directly call NLP functions on query objects?

mindmeld/components/domain_classifier.py
vijay120: For further clarity:
```
>>> list(zip(*[(i,i) for i in range(10)]))
[(0, 1, 2, 3, 4, 5, 6, 7, 8, 9), (0, 1, 2, 3, 4, 5, 6, 7, 8, 9)]
```

mindmeld/components/entity_recognizer.py
vijay120: As we discussed, since there are many refs to the resource loader, it doesnt make sense to unload it

mindmeld/components/role_classifier.py
```diff
@@ -98,24 +98,26 @@ def fit(self, queries=None, label_set=None, incremental_timestamp=None, **kwargs
...
+        new_hash = self._get_model_hash(model_config, label_set)
         cached_model = self._resource_loader.hash_to_model_path.get(new_hash)
 
         if incremental_timestamp and cached_model:
-            logger.info("No need to fit. Loading previous model.")
-            self.load(cached_model)
-            return
+            logger.info("No need to fit. Previous model is cached.")
+            if load_cached:
+                self.load(cached_model)
```
vijay120, line 107: Shouldn't we set the following here?
```
        self.ready = True
        self.dirty = True
```

Likewise for the other classifiers where we do:
```
            if load_cached:
                self.load(cached_model)
                return True
```

mindmeld/core.py
vijay120: There are quite a few places that are not in the `nlp.train` flow that rely on `char_maps`. For example:
`nlp.inspect -> process_markup -> _process_annotations -> NestedEntity.from_query -> transform_span -> transform_index -> _unprocess_index -> _char_maps`

mindmeld/markup.py
```diff
@@ -105,37 +105,28 @@ def load_query_file(
         query_cache (QueryCache): A container containing cache query objects
```
vijay120, line 105: Could we add a description here of `is_gold `

mindmeld/models/helpers.py
```diff
@@ -463,3 +466,60 @@ def add_resource(func):
...
 
     return add_resource
+
+
+class FileBackedList:
+    """
+    FileBackedList implements an interface for simple list use cases
+    that is backed by a temporary file on disk.  This is useful for
+    simple list processing in a memory efficient way.
+    """
```
vijay120, line 476: Why couldnt we use `https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile` instead of implementing our own version?

mindmeld/models/tagger_models.py
```diff
@@ -131,9 +131,14 @@ def fit(self, examples, labels, params=None):
         # Shuffle to prevent order effects
         indices = list(range(len(labels)))
         random.shuffle(indices)
-        examples = [examples[i] for i in indices]
-        labels = [labels[i] for i in indices]
-
+        try:
+            examples.reorder(indices)
```
vijay120, line 135: Since this assumes a non-list datatype, can we update this functions description for the type for `examples `

mindmeld/models/taggers/crf.py
```diff
@@ -126,7 +127,8 @@ def _preprocess_data(self, X, fit=False):
         if fit:
             self._feat_binner.fit(X)
 
-        new_X = []
+        # We want to use a list for in-memory and a LineGenerator for disk based
+        new_X = X.__class__()
```
vijay120, line 131: nice!

mindmeld/resource_loader.py
```diff
@@ -40,7 +40,105 @@
...
+            Override this function for getting other types of cached data.
+            '''
+            return self.source.cache.get(self.row_ids[key])
+
+        @staticmethod
+        def passthrough(processed_query):
+            return processed_query
+
+        def reorder(self, indices):
+            self.row_ids = [self.row_ids[i] for i in indices]
```
vijay120, line 121: Is the expectation that the indices passed in are sorted according to the values to match the alternate `.sorted()` operation? I am struggling to find where this is called in this UI, but from my memory, we passed in `.reoder([i in range(len(list))])` and the alternate was `list.sorted()`. I could be mistaken.

mindmeld/resource_loader.py
```diff
@@ -537,131 +629,121 @@ def _update_query_file_dates(self, query_tree):
...
                             if stemmed_ngram_query != ngram_query:
                                 ngram_tokens.append(stemmed_ngram_query)
-                word_freq_dict.update(ngram_tokens)
-        return word_freq_dict
+                    self.word_freq_dict.update(ngram_tokens)
 
-    def _build_query_freq_dict(self, **kwargs):  # pylint: disable=no-self-use
-        """Compiles frequency dictionary of normalized and stemmed query strings
+        def get_resource(self):
+            return self.word_freq_dict
```
vijay120, line 703: wouldnt this also need a counter?

mindmeld/resource_loader.py
```diff
@@ -537,131 +629,121 @@ def _update_query_file_dates(self, query_tree):
...
-        enable_stemming = kwargs.get(ENABLE_STEMMING_ARGS)
-        word_freq_dict = Counter()
-        for length, threshold in zip(kwargs.get("lengths"), kwargs.get("thresholds")):
-            if threshold > 0:
-                ngram_tokens = []
-                for query in kwargs.get("queries"):
+                    self.char_freq_dict.update(character_tokens)
+
+        def get_resource(self):
+            return self.char_freq_dict
```
vijay120, line 675: Wouldnt this also need a counter?

mindmeld/query_cache.py
```diff
@@ -14,131 +14,197 @@
...
+            processed_query(ProcessedQuery): The ProcessedQuery generated for this example
+        Returns:
+            integer: The unique id of the query in the cache.
+        '''
+
+        cursor = self.connection.cursor()
+        cursor.execute('''
+        INSERT OR IGNORE into queries values (?, ?, ?, ?, ?)
+        ''', (key,
+              json.dumps(processed_query.to_cache()),
```
vijay120, line 138: We could use ultrajson here (https://github.com/ultrajson/ultrajson), based on these benchmarks: https://medium.com/@shmulikamar/python-serialization-benchmarks-8e5bb700530b since it has a faster serialization/deserialization than JSON stdlib. But the performance is not a significant improvement, so maybe we can punt on it.

vijay120 (commented):
Awesome improvements, a few comments but nothing major sticks out.

mindmeld/components/role_classifier.py
davidmfrey: Right, however, I don't know that we should set `dirty = True`.  There are no changes that require saving back to disk, right?

mindmeld/query_cache.py
davidmfrey: My general opinion is that we need fairly significant reasons to bring in another dependency.  This doesn't seem like one since we'd probably not see a major difference in training time.

mindmeld/resource_loader.py
davidmfrey: This is called from the `fit()` method in tagger_models and text_models. They create a list of integers from 0-len(list), randomize it, then use it to randomize the examples and labels in the same order.

mindmeld/resource_loader.py
davidmfrey: self.work_freq_dict is a `Counter`.  Maybe I should change the name of the variable.  I just kept them the same as what they were.

mindmeld/components/classifier.py
```diff
@@ -180,6 +178,11 @@ def fit(self, queries=None, label_set=None, incremental_timestamp=None, **kwargs
                 values are either a kwargs dict which will be passed into the
                 feature extractor function, or a callable which will be used as to
                 extract features.
+            load_cached (bool): If the model is cached on disk should it be loaded?
+
+        Returns:
+            True if model was loaded and fit, False if a valid cached model exists.  The cached
+            model must be loaded if fit() returns False.
```
vijay120, line 185:
```suggestion
            True if model was loaded and fit, False if a valid cached model exists but not loaded (controlled by the load_cached 
            arg).  The cached model must be loaded if fit() returns False. 
```

mindmeld/components/nlp.py
```diff
@@ -136,23 +136,27 @@ def build(self, incremental=False, label_set=None):
...
         for child in self._children.values():
             # We pass the incremental_timestamp to children processors
             child.incremental_timestamp = self.incremental_timestamp
-            child.build(incremental=incremental, label_set=label_set)
-            if incremental:
-                child.dump()
-
-        self.resource_loader.query_cache.dump()
-        self.ready = True
-        self.dirty = True
```
vijay120, line 155: Shouldnt we set `self.dirty = True` here?

mindmeld/components/role_classifier.py
vijay120: Understood, makes sense!

mindmeld/core.py
```diff
@@ -406,7 +439,9 @@ class ProcessedQuery:
         confidence (dict): A dictionary of the class probas for the domain and intent classifier
```
vembar, line 439: TIL: We keep track of whether a query is human labelled or predicted! Do we use `is_gold` anywhere?

mindmeld/markup.py
```diff
@@ -105,37 +105,28 @@ def load_query_file(
...
-                query = load_query(
-                    query_text,
-                    query_factory=query_factory,
-                    domain=domain,
-                    intent=intent,
-                    is_gold=is_gold,
-                )
-                query_cache.set_value(domain, intent, query_text, query)
-        else:
+        key = query_cache.get_key(domain, intent, query_text)
```
vijay120, line 117: It would be great if we can abstract out the details of the cache key gen and access through an API that preserves the previous call: `query_cache.get_value(domain, intent, query_text)`

mindmeld/models/helpers.py
```diff
@@ -463,3 +466,60 @@ def add_resource(func):
...
+
+        def __len__(self):
+            return len(self.source)
+
+        def __next__(self):
+            try:
+                line = next(self.file_handle)
+                return json.loads(line)
+            except Exception:
+                self.file_handle.close()
```
vijay120, line 519: Log an error here

mindmeld/resource_loader.py
```diff
@@ -537,131 +629,121 @@ def _update_query_file_dates(self, query_tree):
...
+        Compiles n-gram character frequency dictionary of normalized query tokens
+        '''
+        def __init__(self, lengths, thresholds):
+            self.lengths = lengths
+            self.thresholds = thresholds
+            self.char_freq_dict = Counter()
+
+        def add(self, query):
+            for length, threshold in zip(self.lengths, self.thresholds):
+                if threshold > 0:
```
vijay120, line 666: @vembar It looks like these threshold are 0 for ANLP. Is that by design?

mindmeld/resource_loader.py
vijay120: No worries, this is fine.

mindmeld/resource_loader.py
vembar: Don't think we're using char_ngrams in ANLP right now, but setting threshold to 0 will essentially include every character ngram in the data as a feature [L501](https://github.com/cisco/mindmeld/blob/2ce240eb8d85cb979969e10f944a26c1f22679a9/mindmeld/models/query_features.py#L501). Trying to recall why we set the default threshold to 0 rather than 1. Seems unintuitive and also creates a feature set mismatch (We're still fine since sklearn vectorizer will ignore features not seen during training)? @serapio Do you remember?

mindmeld/core.py
```diff
@@ -742,6 +839,30 @@ def __init__(
...
+            'entity_type': self.type,
+            'role': self.role,
+            'value': self.value,
+            'display_text': self.display_text,
+            'confidence': self.confidence
+        }
+
+    @staticmethod
+    def from_cache(obj):
+        return Entity(**obj)
```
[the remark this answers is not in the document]
davidmfrey: This is in the `Entity` class.  It should be an entity, right?

mindmeld/models/text_models.py
```diff
@@ -158,9 +158,14 @@ def fit(self, examples, labels, params=None):
         # Shuffle to prevent order effects
         indices = list(range(len(labels)))
         random.shuffle(indices)
-        examples = [examples[i] for i in indices]
-        labels = [labels[i] for i in indices]
-
+        try:
```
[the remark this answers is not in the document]
davidmfrey: I ended up enforcing ProcessedQueryList.Iterator types here instead of allowing lists.

mindmeld/components/role_classifier.py
davidmfrey: My understanding of `dirty` is "This model in memory has changes made to it that should be written to disk".  That's not true if we've just loaded the model from disk.  I may not understand what exactly the `dirty` flag is used for here though.

mindmeld/components/entity_recognizer.py
```diff
@@ -71,51 +72,55 @@ def _get_model_config(self, **kwargs):  # pylint: disable=arguments-differ
...
-        if not label_set:
-            label_set = self._model_config.train_label_set
-            label_set = label_set if label_set else DEFAULT_TRAIN_SET_REGEX
+        if not queries:
+            if not label_set:
+                label_set = self._model_config.train_label_set
+                label_set = label_set if label_set else DEFAULT_TRAIN_SET_REGEX
+            queries = self._get_queries_from_label_set(label_set)
+        elif not isinstance(queries, ProcessedQueryList):
+            queries = ProcessedQueryList.from_in_memory_list(queries)
```
vijay120, line 94: I find this code block repeated across most `Classifier ` sub-classes. Can we create a helper function in the `Classifier` class that will return the queries based on `label_set` and optional `queries `?

davidmfrey: Great idea!  I'll see what that looks like.

mindmeld/components/role_classifier.py
vijay120: Make sense, no changed needed here

mindmeld/core.py
```diff
@@ -229,6 +240,41 @@ def __init__(
         self._timestamp = timestamp
         self.stemmed_tokens = stemmed_tokens or tuple()
 
+    def char_maps_to_cache(self):
+        return {json.dumps(k) : v for k,v in self._char_maps.items()}
```
vijay120, line 244: Since the keys are tuples, why cant we do this (since json.dump has some unsavory mem footprint):

Serialization: `return {str(k) : v for k,v in self._char_maps.items()}`

Deserialization of the tuple:`tuple(map(int, str.split(', ')))`

mindmeld/models/model.py
```diff
@@ -1185,27 +1189,35 @@ def initialize_resources(self, resource_loader, examples=None, labels=None):
...
         # get list of resources required by feature extractors
         required_resources = self.config.required_resources()
-
         enable_stemming = ENABLE_STEMMING in required_resources
-
-        # load required resources if not present in model resources
+        resource_builders = {}
         for rname in required_resources:
-            if rname == ENABLE_STEMMING:
+            if rname in self._resources:
```
vijay120, line 1195: I think the intention of this line is not the override a key in `self._resources`, but judging by the current code, we do override the key even if it exists in the map. Maybe we should remove this check?

mindmeld/query_cache.py
```diff
@@ -14,131 +14,201 @@
...
+            cursor.execute('''
+            DROP TABLE IF EXISTS queries;
+            ''')
+            cursor.execute('''
+            DROP TABLE IF EXISTS version;
+            ''')
+        # Create table to store queries
+        cursor.execute('''
+        CREATE TABLE IF NOT EXISTS queries
+        (hash_id TEXT PRIMARY KEY, query TEXT, raw_query TEXT, domain TEXT, intent TEXT)
```
vijay120, line 56: I propose also adding a composite primary key containing app_path and text as the primary key. Since we have users who use MindMeld in a multi-application mode on the same python process, we need to be able to distinguish same text queries which are in different domains/intents in the different applications. So the app_path + text would be a unique composite key in such a case.

mindmeld/query_cache.py
```diff
@@ -14,131 +14,201 @@
...
+
+    def get_raw_query(self, row_id):
+        '''
+        Get the raw text only from a cached example.  See notes on get().
+        '''
+        cursor = self.connection.cursor()
+        cursor.execute('''
+        SELECT raw_query FROM queries WHERE rowid=(?);
+        ''', (row_id,))
+        return cursor.fetchone()[0]
```
vijay120, line 182: Is there a risk of retuning an empty list and we referencing the index 0, creating an out of index exception?

mindmeld/resource_loader.py
```diff
@@ -40,7 +40,176 @@
...
+        def __getitem__(self, key):
+            return self.source.cache.get_query(self.row_ids[key])
+
+    class EntitiesIterator(Iterator):
+        def __getitem__(self, key):
+            return self.source.cache.get_entities(self.row_ids[key])
+
+    class DomainIterator(Iterator):
+        def __init__(self, source):
+            cached = not isinstance(source.cache, ProcessedQueryList.MemoryCache)
```
vijay120, line 157: Its not immediately clear why DomainIterator and IntentIterator have the special __init__ function but `EntitiesIterator ` does ot

mindmeld/query_cache.py
davidmfrey: I'm not sure I understand the problem.  The `hash_id` primary key is derived from domain, intent, and raw_text.  The only time a conflict would occur is if two apps have the same example in the same domain and intent.  I expect that case would actually function correctly as it is now?

mindmeld/resource_loader.py
```diff
@@ -40,7 +40,176 @@
...
+            self.iter_idx = -1
+            self.elements = elements
+
+        def __getitem__(self, key):
+            return self.elements[key]
+
+        def __len__(self):
+            return len(self.elements)
+
+        def reorder(self, indices):
```
vijay120, line 185: Cant we just use the parent class's `reorder ` since `result_cache ` is None anyway?

mindmeld/query_cache.py
davidmfrey: No, this is why I have the access method using the row_id.  It's guaranteed to be there unless there is corruption of the db file.

mindmeld/resource_loader.py
```diff
@@ -400,12 +564,24 @@ def flatten_query_tree(query_tree):
...
-        flattened = []
+        flattened = ProcessedQueryList()
         for _, intent_queries in query_tree.items():
             for _, queries in intent_queries.items():
-                flattened.extend(queries)
+                if not flattened.cache:
+                    flattened.cache = queries.cache
+                elif flattened.cache != queries.cache:
+                    logger.error('query_tree is built from incompatible query_caches')
+                    raise ValueError()
```
vijay120, line 574: We can add the `'query_tree is built from incompatible query_caches'` to the `ValueError `

mindmeld/resource_loader.py
davidmfrey: It's because domain and intent are small bits of data that are cached in memory.  When they are run through multiple times, the data is used from memory instead of reading from the db again.

vijay120 (commented):
The only outstanding issue I raised is the primary key change to take account of the app_name. Apart from that, the other issues are questions or nits. This PR is very close to being shipped.

mindmeld/models/model.py
davidmfrey: There was a check before as well.  
https://github.com/cisco/mindmeld/blob/master/mindmeld/models/model.py#L1197
This is slightly different, but it should end up equal from what I saw.

davidmfrey: This check is actually required to load models.  If we try to reload the resources when the models are loaded from cache it fails because there  are no examples or labels at that time.

vijay120 (approved):
Great work @davidmfrey , this change is great and we welcome more of such advanced designs being incorporated to the platform to improve performance.

PR #321, opened 2021-06-08
Adding deep neural networks support to MindMeld (Phase 1) (P1)

murali1996:
`Tag`: [`Phase 1/3 of NN support`](https://github.com/cisco/mindmeld/pull/321)

# What does this PR do?
This PR is the first of three PRs aimed at adding deep neural models support to MindMeld. MindMeld currently supports using Sklearn based models and an LSTM model through Tensorflow. Given the rapidly evolving state of the art, adding support for deep neural models in MindMeld facilitates more options for developers and researchers. However, to leverage deep architectures, simply adding model architectures wouldn't suffice as those models are not serializable and do not exactly fit in the regime of dumping/loading models as done for the sklearn based models.

# What are the changes in this PR?
This PR addresses some refactoring that is required in the sub-directory `./mindmeld/models` in order to facilitate adding pytorch based deep neural models while still maintaining backwards compatability. The following are the files that are changed in this PR, marked with some overview of what's the change.

- Files wherein some code is modified:
```
.
├── mindmeld/
│   ├── components/
│   │   ├── classifier.py, role_classifier.py, entity_recognizer.py
│   ├── models/
│   │   ├── model.py, text_models.py, tagger_models.py, helpers.py __init__.py
```
The changes here are mainly to move model loading/dumping functionality from `classifier.py`, `role_classifier.py` and `entity_recognizer.py` classes to the respective models' classes, so that these classifier modules can be used as-is even if we add a new models (eg. non-serializable models), without worrying about creating if-else conditions at the classifiers abstraction. This also decouples and creates better abstraction between classifiers and models classes-- classifiers classes are high level abstraction that only care about passing relevant resources and data to models classes abstraction. Additionally, to make the classes structure in `*_model.py` files bit more vivid, a new base class has been implemented in `model.py` file.

- Files which are newly added
```
.
├── mindmeld/
│   ├── models/
│   │   ├── auto_models.py
```
The auto model creates or loads models with just the model configs dictionary or from a dumped pickle path. This is useful as the classifier classes need no worry about what's happening while creating/dumping/loading models and leave those decisions to the models classes. This adds a clean layer of abstraction for classifiers and makes them re-usable across different kinds of models. While the creating models functionality is already decoupled through `helpers.py/create_model()` method, the dumping/loading aren't and hence using 'auto_model.py/AutoModel' class helps to do creating/dumping/loading without any coupling with classifiers classes. All in all, the classifiers only care about `helpers.py/create_model()` and `helpers.py/load_model()` methods.

- ~~Files that were moved across directories to bring all embedder models code under common roof~~
~~```mindmeld/models/taggers/embeddings.py → mindmeld/models/dense_featurizers/embeddings.py```~~
Update: This edit is reverted so as to maintain backwards compatibility for models that were serialized and dumped.

- Files that were moved across directories to bring all features' methods under a common roof
```
mindmeld/models/query_features.py → mindmeld/models/features/query_features.py
mindmeld/models/entity_features.py → mindmeld/models/features/entity_features.py
```
This is done with a futuristic idea in mind. The following is a design plan which can be looked into in the future:
```
.
├── models/
│   ├── features/
│   │   ├── sparse_features
│   │   │   ├── query_features.py, entity_features.py
│   │   ├── dense_features
│   │   │   ├── embedder_models.py
```

- Split code in `model.py` file to keep labels and evaluation classes separately from models classes 
```
mindmeld/models/model.py → {mindmeld/models/evaluation.py, mindmeld/models/labels.py, mindmeld/models/model.py}
```
As the label encoder classes and evaluation classes are not related to models classes, it makes sense to group them separately and hence this change.

# Upcoming PRs:
- [`Phase 2/3 of NN support`](https://github.com/cisco/mindmeld/pull/325): Pytorch based neural architectures will be made available to train/test. 
- `Phase 3/3 of NN support`: Add Joint modeling support in MindMeld, where intent/domain/entity classifiers can be trained together as global models


mindmeld/components/classifier.py
```diff
@@ -422,13 +420,7 @@ def _get_model_config(loaded_config=None, **kwargs):
...
         return ModelConfig(**model_config)
 
-    def _data_dump_payload(self):
-        return self._model
-
-    def _create_and_dump_payload(self, path):
-        joblib.dump(self._data_dump_payload(), path)
-
-    def dump(self, model_path, incremental_model_path=None):
+    def dump(self, model_path, incremental_model_path=None, metadata=None):
```
vijay120, line 423: what is metadata?

mindmeld/components/entity_recognizer.py
```diff
@@ -176,24 +169,11 @@ def load(self, model_path):
         logger.info(
             "Loading entity recognizer: domain=%r, intent=%r", self.domain, self.intent
         )
-        try:
-            er_data = joblib.load(model_path)
-
-            self.entity_types = er_data["entity_types"]
-            self._model_config = er_data.get("model_config")
+        metadata = load_model(model_path)
```
vijay120, line 172: `er_metadata`

mindmeld/models/taggers/lstm.py
```diff
@@ -954,25 +954,28 @@ def _predict_proba(self, X):
...
 
         if not os.path.isdir(path):
             os.makedirs(path)
 
         if not self.saver:
-            # This conditional happens when there are not entities for the associated
-            # model
-            return
+            # This conditional happens when there are not entities for the associated model
+            return path
```
vijay120, line 978: I dont understand why we return the path when the path is passed into the args

mindmeld/models/taggers/taggers.py
```diff
@@ -181,22 +181,22 @@ def _predict_proba(X):
...
+        return True
+
     @staticmethod
-    def dump(model_path, config):
+    def dump(model_path):
         """
-        Since traditional SKLearn models are easily serializable, we can
-        use JobLib to serialize them. So we alter the context object to make
-        this explicit.
+        Dumps the model to memory. This is a no-op since we do not
```
vijay120, line 192: Why are we dumping to memory? Shouldnt we dumping to disk?

mindmeld/models/text_models.py
```diff
@@ -29,33 +29,31 @@
...
-RANDOM_FOREST_TYPE = "rforest"
-SVM_TYPE = "svm"
-SUPER_LEARNER_TYPE = "super-learner"
-BASE_MODEL_TYPES = [LOG_REG_TYPE, DECISION_TREE_TYPE, RANDOM_FOREST_TYPE, SVM_TYPE]
+logger = logging.getLogger(__name__)
 
-# default model scoring type
-ACCURACY_SCORING = "accuracy"
 
+class TextModel(Model):
```
vijay120, line 44: Unclear why we need to inherit from Model. These fields can be represented as Enums

mindmeld/models/text_models.py
```diff
@@ -472,5 +383,124 @@ def _get_feature_scaler(self):
...
+        return self._extract_features(
+            example, dynamic_resource=dynamic_resource, tokenizer=self.tokenizer
+        )
+
+    def dump(self, path, metadata=None):
+        metadata = metadata or {}
+        metadata.update({
+            "model": self,
+            "model_config": self.config,
+            "serializable": True}
```
vijay120, line 482: Why is `serializable ` field necessary? What models are not serializable?

mindmeld/components/entity_recognizer.py
```diff
@@ -156,7 +143,13 @@ def dump(self, model_path, incremental_model_path=None):
         logger.info(
             "Saving entity classifier: domain=%r, intent=%r", self.domain, self.intent
         )
-        super().dump(model_path, incremental_model_path)
+        metadata = {
```
ritvikshrivastava, line 146: A suggestion: instead of initializing different metadata dictionaries at various points in the code: in individual classifiers (with these keys), in the base model (with the model and model config keys) etc., a more ideal solution might be to create a Metadata/Metadata manager class separately and define all possible attributes for the class explicitly there. 

* Instance(s) can then be utilized at different points of this refactor and attributes tracked more easily in one place
* It would be cleaner to update instances through classmethods instead of standalone dictionary updates
* validation for metadata attributes can be done (using marshmallow).

mindmeld/models/model.py
```diff
@@ -1223,104 +654,54 @@ def initialize_resources(self, resource_loader, examples=None, labels=None):
...
-        self.system_entity_recognizer = SystemEntityRecognizer.get_instance()
+    def dump(self, path, metadata=None):
+        # do featurizer specific dumping
 
-    def _get_tag_scheme(self):
-        return self.config.model_settings.get("tag_scheme", "IOB").upper()
+        # dump model configs
+        # metadata = metadata or {}
+        # metadata.update({'model_config': self.featurizer.config})
+        super().dump(path)
```
vembar, line 679: Why are we calling this if we are expecting the specific model classes to implement `dump()`?

mindmeld/components/classifier.py
vijay120: I would prefer if metadata is a "ModelMetadata" object that can be standardized read/write APIs across all classifiers

mindmeld/models/auto_models.py
```diff
@@ -0,0 +1,133 @@
...
+class AutoModel:
+    """Auto class that identifies appropriate text/tagger model from text_models.py/tagger_models.py
+    to load based on the inputted configs or from the loaded configs file.
+
+    The .from_config() methods allows to load the appropriate model when passed in with ModelConfig.
+    The .from_path() method uses BaseModel's load method to load a dumped config, which is them used
+    to load appropriate model and return it through a metadata dictionary object.
+
+    """
+    ALLOWED_MODEL_TYPES = ["text", "tagger"]
```
vijay120, line 39: these can be enums

mindmeld/models/text_models.py
vembar: Since we have different base classes for classical and neural models, do we still need this key?

mindmeld/models/features/entity_features.py
```diff
@@ -12,8 +12,9 @@
 # limitations under the License.
 
 """This module contains feature extractors for entities"""
-from .helpers import GAZETTEER_RSC, get_ngram, register_entity_feature, requires
-from ..core import QueryEntity, Query, NestedEntity
+from mindmeld.core import QueryEntity, Query, NestedEntity
+
```
vijay120, line 16: remove extra space

mindmeld/models/text_models.py
```diff
@@ -472,5 +383,124 @@ def _get_feature_scaler(self):
         }.get(scale_type)
         return scaler
 
+    ##################
+    # abstract methods
```
vijay120, line 387: what does this mean? the below methods are not abstract methods.

mindmeld/components/role_classifier.py
```diff
@@ -127,16 +125,21 @@ def fit(self,
...
             self._model = model
             self.config = ClassifierConfig.from_model_config(self._model.config)
+        else:
+            # This _else_ conditional is created to support moving .dump() and .load()
+            # to self._model.dump() and self._model.load() respectively.
+            #
+            # Any calls to abstract methods other than .fit(), .load(), .unload(), and .dump() are
+            # to be carefully used by identifying if role classification is valid or not with the
+            # conditional`if len(self.role_classifier.roles) > 1: ...; else: ...`
+            self._model = model
```
vembar, line 135: If we have no valid examples to train on, shouldn't `self._model` be set to `None`?

vijay120 (approved):
Overall, PR looks good. Please address the outstanding comments.

mindmeld/models/taggers/lstm.py
murali1996: Good question. So the value of `path` is modified in this method of LSTM and in [line 392 in tagger_models.py](https://github.com/cisco/mindmeld/blob/95b22e18eaae85e5102ab558b871c8653bd4c7af/mindmeld/models/tagger_models.py#L392), once the modified path value is returned, you can see that this modified path is used to save `tagger_vars`. This logic is existing in current version of MindMeld as well ([see this line](https://github.com/cisco/mindmeld/blob/11478f302c835347ca9a558679b5e85fea669564/mindmeld/models/tagger_models.py#L368)).

This scenario is happening because there are items to be dumped for (1) entity_recognizer (2) tagger_model (3) underlying clf algo. And the dump path for (2) is being created from the path decided by (3). 

Proposal: We can merge (1) and (2) into `ModelMetadata`, and when loading, the (2) tagger_model first obtains all its required items from dump and then passes the remainders to (1). Let me know if thats good to implement.

mindmeld/components/classifier.py
murali1996: @vijay120 I'm slightly confused where should this metadata class reside: in `models/` sub-directory or the `components/`.  The reason for my question is that, while creating a metadata instance for dumping, classifiers have to create an instance of the metadata class, fill it up with metadata and pass it to models. But when loading, models have to create an instance and pass it to classifiers layer. Any suggestions here?

mindmeld/models/text_models.py
murali1996: We just need it in `TaggerModel`, as it currently also supports LSTM model.

mindmeld/models/taggers/taggers.py
murali1996: typo corrected

mindmeld/models/model.py
murali1996: So all models are supposed to dump `model_config` as it is required to determine the right class to instantiate when loading. So calling super().dump(..) validates that and does the dumping of metadata.

vembar: Once the specific model overrides this method, doesn't it become the responsible of that version of `dump()` to write the model config? This method will just raise a `NotImplementedError`

mindmeld/models/auto_models.py
murali1996: @vijay120 I'm not sure if enum is required here as i am using this var `ALLOWED_MODEL_TYPES` just to register models and nothing else. Do you still think its required?

mindmeld/components/role_classifier.py
murali1996: tldr: Sure, will make the required change.

On top of the comment in previous lines, I realized we can do `self._model=None` and put a try-catch in the load method wherein if there exists no model, the load_model() (based on AutoModel) throws an error and upon catching the error, the model is set to None in load method as well.

mindmeld/components/classifier.py
```diff
@@ -469,11 +464,11 @@ def load(self, model_path):
...
             model_path (str): The location on disk where the model is stored
         """
-        try:
-            self._model = joblib.load(model_path)
-        except (OSError, IOError) as e:
-            msg = "Unable to load {}. Pickle at {!r} cannot be read."
-            raise ClassifierLoadError(msg.format(self.__class__.__name__, model_path)) from e
+
+        model = load_model(model_path)
+        self._model = model
```
ritvikshrivastava, line 469:
```suggestion
        self._model = load_model(model_path)
```

mindmeld/components/role_classifier.py
```diff
@@ -151,6 +149,15 @@ def dump(self, model_path, incremental_model_path=None):
             self.intent,
             self.entity_type,
         )
+        # classifier specific dump
+        rc_data = {"roles": self.roles}
+        pickle.dump(rc_data, open(self._get_classifier_resources_save_path(model_path), "wb"))
```
ritvikshrivastava, line 154: Same comment as ER

mindmeld/models/model.py
```diff
@@ -239,615 +224,116 @@ def required_resources(self):
...
-            expected.append(vec)
-            expected_flat.extend(vec)
-        return RawResults(
-            predicted=predicted,
-            expected=expected,
-            text_labels=text_labels,
-            predicted_flat=predicted_flat,
-            expected_flat=expected_flat,
-        )
+        # every XxxModel derived from baseModel has one default .pkl dump file that contains a
```
ritvikshrivastava, line 289:
```suggestion
        # every ABCModel derived from baseModel has one default .pkl dump file that contains a
```

mindmeld/models/auto_models.py
```diff
@@ -0,0 +1,134 @@
...
+                  f"Found path={path} of type({type(path)})"
+            raise ValueError(msg)
+
+        if not path.endswith(".pkl"):
+            msg = "Model Path must end with .pkl for AutoModel to be able to identify the model"
+            raise ValueError(msg)
+
+        try:
+
+            # if loading from path, determine the XxxModel type & return after doing xxxModel.load()
```
ritvikshrivastava, line 78:
```suggestion
            # if loading from path, determine the ABCModel type & return after doing ABCModel.load()
```

mindmeld/models/tagger_models.py
```diff
@@ -324,78 +357,91 @@ def evaluate(self, examples, labels):
...
+        # In TaggerModel, unlike TextModel, two dumps happen,
+        # one, the underneath classifier and two, the model metadata
 
-        Args:
-            path (str): The path to dump the model to
-            config (dict): The config containing the model configuration
-        """
-        self._clf.dump(path, config)
+        os.makedirs(os.path.dirname(path), exist_ok=True)
+        metadata = {"serializable": self._clf.is_serializable}
```
ritvikshrivastava, line 366:
```suggestion
        tagger_data = {"serializable": self._clf.is_serializable}
```
Since you're removing metadata altogether, this would be better aligned I think?

mindmeld/models/tagger_models.py
```diff
@@ -324,78 +357,91 @@ def evaluate(self, examples, labels):
...
-        self._clf.load(path)
-        variables_to_load = joblib.load(os.path.join(config["model"], ".tagger_vars"))
-        self._current_params = variables_to_load["current_params"]
-        self._label_encoder = variables_to_load["label_encoder"]
-        self._no_entities = variables_to_load["no_entities"]
 
-    def get_feature_matrix(self, examples, y=None, fit=False):
-        raise NotImplementedError
+        # load model metadata
+        metadata = joblib.load(path)
```
ritvikshrivastava, line 395: same for the load functionality consistency as well
```suggestion
        tagger_data = joblib.load(path)
```

mindmeld/models/taggers/lstm.py
```diff
@@ -954,25 +954,28 @@ def _predict_proba(self, X):
...
-        config["serializable"] = False
 
         if not os.path.isdir(path):
             os.makedirs(path)
 
         if not self.saver:
-            # This conditional happens when there are not entities for the associated
-            # model
-            return
+            # This conditional happens when there are not entities for the associated model
```
ritvikshrivastava, line 977:
```suggestion
            # This conditional happens when there are no entities for the associated model
```

mindmeld/models/taggers/lstm.py
ritvikshrivastava: We won't need this anymore if we decouple the dumps, right?

ritvikshrivastava (commented):
The dump functionality decoupling changes look good! Have a few thoughts though.

mindmeld/models/model.py
murali1996: I modified this code snippet in the latest version of code. So the subclasses get to define a method called `_dump()` which is in-turn called in a publicly exposed method `dump()` residing in the base class.

mindmeld/models/tagger_models.py
murali1996: We are still including the class instance itself in the dictionary along with some bunch of data related to tagging, so I'm not sure if just calling it tagger_data is meaningful 🤔

mindmeld/components/entity_recognizer.py
```diff
@@ -156,6 +146,20 @@ def dump(self, model_path, incremental_model_path=None):
...
+        # classifier specific dump
+        er_data = {
+            "entity_types": self.entity_types,
+            "w_ngram_freq": self._model.get_resource("w_ngram_freq"),
+            "c_ngram_freq": self._model.get_resource("c_ngram_freq"),
+            "model_config": self._model_config,
+        }
+        pickle.dump(er_data, open(self._get_classifier_resources_save_path(model_path), "wb"))
+        if incremental_model_path:
+            pickle.dump(
```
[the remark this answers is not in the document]
murali1996: Unified

mindmeld/models/taggers/lstm.py
murali1996: So i went and checked, looks like this is required for backwards compatibility for LSTM model. So based on the returned <path> from the LSTM.dump(), another file with path "<path>.tagger_vars" is created in TaggerModel.dump(). Although such dependency is cleared in this PR's proposed way of dumping, the backwards compatibility requires this path be returned at least for the LSTM model

murali1996: True, will make that change.

mindmeld/models/auto_models.py
vijay120: No its fine.

mindmeld/components/entity_recognizer.py
```diff
@@ -176,25 +175,19 @@ def load(self, model_path):
...
-                self._model = er_data["model"]
-            else:
-                self._model = create_model(self._model_config)
-                self._model.load(model_path, er_data)
-        except (OSError, IOError) as e:
-            msg = "Unable to load {}. Pickle file cannot be read from {!r}"
-            raise ClassifierLoadError(msg.format(self.__class__.__name__, model_path)) from e
+        # classifier specific load
+        try:
+            er_data = pickle.load(open(self._get_classifier_resources_save_path(model_path), "rb"))
```
vijay120, line 184: Do we have to use pickle or can we still use joblib?

mindmeld/resource_loader.py
```diff
@@ -446,7 +446,7 @@ def _load_cached_models(self):
                 hash_val = open(file_path, "r").read()
                 classifier_file_path = file_path.split(".hash")[0]
                 if not os.path.exists(classifier_file_path):
-                    logger.warning("Could not find the serialized model")
+                    logger.info("Could not find the serialized model")
```
vijay120, line 449: I think warning is better here.

mindmeld/components/entity_recognizer.py
murali1996: I did some quick benchmarking and found that pickle is better time-wise, although using either makes no difference otherwise.

mindmeld/resource_loader.py
murali1996: @vijay120 So the warning is leading to failure of a test case. And also, for RoleClassifiers, when no model is required, we do not save any `entity-role.pkl` file which is causing this error.

mindmeld/models/auto_models.py
```diff
@@ -32,26 +31,26 @@ class AutoModel:
     to load based on the inputted configs or from the loaded configs file.
 
     The .from_config() methods allows to load the appropriate model when passed in with ModelConfig.
-    The .from_path() method uses BaseModel's load method to load a dumped config, which is them used
-    to load appropriate model and return it through a metadata dictionary object.
+    The .from_path() method uses AbstractModel's load method to load a dumped config, which is them
```
vembar, line 34:
```suggestion
    The .from_path() method uses AbstractModel's load method to load a dumped config, which is then
```

vembar (approved):

ritvikshrivastava (approved):

mindmeld/components/entity_recognizer.py
```diff
@@ -134,16 +135,14 @@ def fit(self,
         self.dirty = True
         return True
 
-    def _data_dump_payload(self):
-        return {
+    def _create_and_dump_payload(self, path):
```
ritvikshrivastava, line 138: I can see the abstract method being overwritten here, but instead of this, consider dumping everything by passing it to the super.dump() directly? That way we'll have a single place for dumping all model specific data instead of 3 separate functions.

ritvikshrivastava (approved):

mindmeld/components/entity_recognizer.py
murali1996: I think if we do that, the _load_ process might bit become convoluted. Meaning, while loading, it would be a good practice if the base class only loads the dumps the path. But once loaded, base class has no idea what to do with the loaded contents. So it has to pass back the loaded contents to the sub-classes. And doing so feels bit unfaithful while using the name `load()`. What do you think?  @ritvikshrivastava

mindmeld/components/classifier.py
```diff
@@ -454,6 +450,16 @@ def dump(self, model_path, incremental_model_path=None):
             if path == model_path:
                 self.dirty = False
 
+    def _dump(self, path):
```
vijay120, line 453: maybe we can make this into an abstract method?

murali1996: @vijay120 not all classifiers have dump-able data, so i think making it into abstract method might not be necessary and the default action of the method being not to dump anything.

PR #323, opened 2021-06-15
Fix: ngram frequency dict is empty due to wrong threshold check

vembar:
This PR fixes the following bugs:

1) ngram frequency dict is not populated since we check for threshold values greater than the default. Also, the default value should be 1 and not 0 since every ngram in the dict will appear atleast once in the training files. These bugs led to extracting all ngrams irrespective of whether they were in vocabulary or not. The current implementation extracts an OOV token if the ngram was not seen during training or does not have a count equal or greater than the threshold value provided.
With current mindmeld version, looking at the contents gives the following
```
(Pdb) x = resources[WORD_NGRAM_FREQ_RSC]
(Pdb) len(x.keys())
0
(Pdb) x = resources[WORD_FREQ_RSC]
(Pdb) len(x.keys())
5328
```
With this fix, the word ngram dict gets populated
```
(Pdb) x = resources[WORD_NGRAM_FREQ_RSC]
(Pdb) len(x.keys())
23431
(Pdb) x = resources[WORD_FREQ_RSC]
(Pdb) len(x.keys())
5328
(Pdb)
```
2) There were some inconsistencies in preprocessing numbers in text during the dictionary building phase and ngram extraction phase. We mask numbers with 0s during the extraction phase, but not when we populate the dictionaries. This results in all numeric tokens being treated as OOV tokens. This fix masks numerics during the dictionary population stage as well.
3) We now check whether the stemmed token ngrams are present in the dictionary before adding them which is consistent with ngram extraction for entities.
4) While extracting edge ngrams for short queries, we missed out on some ngrams due to an incorrect length comparison. Eg,  for a query "jeremy" and ngram length 1, we would not extract any edge ngrams since we look for ngram lengths greater than length of query.
5) Query cache no longer has a `dump()` which causes `nlp.evaluate()` to error out

Note: These changes may cause some regressions in NLP accuracies.

One additional question that I would like some feedback on - for ngrams (word/char) not found in our dictionary, we replace the whole ngram with OOV. Should we instead extract ngrams with the uni-grams masked as OOVs. Eg, For the query `and meaning` where the word meaning is not in vocabulary, for ngram of length 2, should we extract `OOV` or `and OOV`? (This only effects cases where threshold>1, since otherwise there are no OOV tokens during training)


serapio:
What would the memory usage be if the ngram freq dict only use OOV for individual tokens? That seems more consistent, but could be a lot when N is large

vembar:
> What would the memory usage be if the ngram freq dict only use OOV for individual tokens? That seems more consistent, but could be a lot when N is large

it can only reduce, right? more ngrams would map to the same key since we're collapsing them into OOVs?

mindmeld/models/query_features.py
```diff
@@ -378,10 +378,30 @@ def extract_bag_of_words_features(
         (function) The feature extractor.
     """
     threshold_list = list(thresholds)
-    word_thresholds = threshold_list + [0] * (
+    word_thresholds = threshold_list + [1] * (
         len(ngram_lengths_to_start_positions.keys()) - len(threshold_list)
     )
 
+    def remove_excess_out_of_bounds(n_gram):
```
vijay120, line 385: could you provide a type hint here?

mindmeld/models/query_features.py
```diff
@@ -592,36 +613,37 @@ def extract_char_ngrams(lengths=(1,), thresholds=(0,), **kwargs):
...
-                ):
-                    ngram_counter.update(
-                        [
-                            "char_ngram|length:{}|ngram:{}".format(
-                                len(char_ngram), " ".join(char_ngram)
-                            )
-                        ]
-                    )
+
+                freq = resources[CHAR_NGRAM_FREQ_RSC].get("".join(char_ngram), 0)
```
vijay120, line 629: shouldnt `"".join(char_ngram)` be `" ".join(char_ngram)` ?

mindmeld/models/query_features.py
```diff
@@ -650,39 +672,30 @@ def _extractor(query, resources):
...
-                        tok_stemmed = mask_numerics(stemmed_tokens[index])
+                        tok_stemmed = re.sub('\d','0',stemmed_tokens[index])
                         stemmed_ngram.append(tok_stemmed)
 
-                freq = resources[WORD_NGRAM_FREQ_RSC].get(" ".join(ngram), 1)
+                freq = resources[WORD_NGRAM_FREQ_RSC].get(" ".join(ngram), 0)
 
-                if freq > threshold:
-                    joined_ngram = " ".join(ngram)
+                joined_ngram = " ".join(ngram) if freq >= threshold else "OOV"
```
vijay120, line 684: Lets make `OOV` into a global var?

mindmeld/models/query_features.py
vijay120: To prevent such issues, lets make `" "` into a global var and call `SPACE_SEP.join(char_ngram)`

mindmeld/models/query_features.py
```diff
@@ -650,39 +672,30 @@ def _extractor(query, resources):
...
+                ngram_counter.update(
+                    [
+                        "bag_of_words|length:{}|ngram:{}".format(len(ngram), joined_ngram)
+                    ]
+                )
+                if kwargs.get(ENABLE_STEMMING, False):
+                    freq = resources[WORD_NGRAM_FREQ_RSC].get(" ".join(stemmed_ngram), 0)
```
vijay120, line 691: Lets try to reuse vars when we can like `" ".join(stemmed_ngram` is used twice here.

mindmeld/models/query_features.py
vembar: We store the char ngrams without a space, but add a space while featurizing. Will make it uniform.

mindmeld/models/query_features.py
vembar: Done!

mindmeld/models/query_features.py
```diff
@@ -891,7 +913,7 @@ def _extractor(query, resources):
         freq_features = defaultdict(int)
 
         for tok in tokens:
-            query_freq = "OOV" if resources[WORD_FREQ_RSC].get(tok) is None else "IV"
+            query_freq = OUT_OF_VOCABULARY if resources[WORD_FREQ_RSC].get(tok) is None else "IV"
```
serapio, line 916: "IV" should be a constant too, I guess.

serapio (approved):
Looks good otherwise

vijay120 (approved):

mindmeld/models/query_features.py
vembar: done!

PR #325, opened 2021-06-17
Adding deep neural networks support to MindMeld

murali1996:
`Tag`: [`Phase 2/3 of NN support`](https://github.com/cisco/mindmeld/pull/325) Pytorch based neural networks implementation for MindMeld

### [PPT Slides](https://cisco-my.sharepoint.com/:p:/p/saijayan/Efze5_S4ZWdHt5rPP2zle9gBQZranGELMLfdsx1MVEuY8Q?e=dIpihM)

# Checkpoints (latest on top)
- Params Checks and Benchmarking
    - [x] Ready-to-use default hyper-parameters
        - CNN and LSTM based models
        - BERT-based models from huggingface (warmup_steps, etc.)
    - [x] BERT optimizer (see end of this PR description for details)
    - [x] Test on GPU machine and fix any emerging issues
    - [x] Compare results of neural models on blueprints and in-house datasets
    - [x] Add type hints
- Fixes
    - [x] Modify nlp.py (specifically using entity_recognizer) when there are no entity_types 
    - [x] Clean up codebase, resolve pylint _too-many-local_ and _too-many-instance-attributes_
    - [x] Change _verbose_ terminology in nn base module. (Change naming convention for 'lstm-lstm' and 'cnn-lstm' if needed.)
- Implementations
    - [x] Add unittests
    - [x] Outputs of _input\_encoders.py_ to be a custom object rather than a dictionary.
    - [x] Merge Phase 1
    - [x] Modify requirements file and adopt conditional imports for torch functionality
    - [x] Implement PytorchTaggerModel
    - Implement token classification modules
        - [x] CNN and LSTM based models
        - [x] BERT-based models from huggingface
    - [x] Implement PytorchTextModel
    - Implement sequence classification modules
        - [x] CNN and LSTM based models
        - [x] BERT-based models from huggingface
    - [x] Implement Input text encoders that can be used with pretrained LMs like BERT
    - [x] Implement Input text encoders that can be used with word embedders like Glove
- Add-ons (can also be worked upon in separate PRs):
    - MM Documentation (see below)
    - Integrating mindmeld tokenizer (loaded through resource loader and input configs) into _input\_encoder.py_
    - CNN models can be deep, i.e. support more than 1 CNN layers. We allow only 1 layer currently
    - Model quantization can be allowed upon training
    - Pytorch's DataLoader can be used with collate_fn and pin_memory to improve training speed when using GPUs
    - Model's weights can be averaged across multiple best performing epoch-checkpoints
    - Adapters-like modeling when using large pretrained LMs like BERT

### MM Documentation
As discussed offline, docs for neural nets integration will be aded as a separate PR

### Usage
Similar to how we access sklearn based models, we can use neural models through the configs in `config.py` file of your app. Below are some example snippets.

```python
INTENT_CLASSIFIER_CONFIG = {
    "model_type": "text",
    "model_settings": {
        "classifier_type": "lstm",  # embedder, cnn, lstm
    },
    "params": {
        # "embedder_type": "glove",
        'emb_dim': 2,
        'number_of_epochs': 1
    },
}

ENTITY_RECOGNIZER_CONFIG = {
    'model_type': 'tagger',
    'model_settings': {
        'classifier_type': 'lstm-lstm',  # embedder, lstm-pytorch, lstm-lstm, cnn-lstm
    },
    "params": {
        "embedder_type": "glove",
        # 'emb_dim': 10,
        # 'char_emb_dim': 20,
        # 'use_crf_layer': True,
        #
        # "embedder_type": "bert",
        # "pretrained_model_name_or_path": "bert-base-cased",
    },
}
```

### Accessing default params or params that can be passed in for any of the nn modules
```python
from mindmeld.models import nn_utils
print(nn_utils.__all__)
# >>> ['EmbedderForSequenceClassification', 'CnnForSequenceClassification', 'LstmForSequenceClassification', 
# >>>  'BertForSequenceClassification', 'EmbedderForTokenClassification', 'LstmForTokenClassification', 
# >>>  'CharCnnWithWordLstmForTokenClassification', 'CharLstmWithWordLstmForTokenClassification', 
# >>>  'BertForTokenClassification'
# >>> ]
nn_utils.CnnForSequenceClassification().get_default_params()
# >>> {
# 	'device': 'cpu',
# 	'number_of_epochs': 100,
# 	'patience': 10,
# 	'batch_size': 32,
# 	'gradient_accumulation_steps': 1,
# 	'max_grad_norm': None,
# 	'optimizer': 'Adam',
# 	'learning_rate': 0.001,
# 	'validation_metric': 'accuracy',
# 	'verbose': True,
# 	'dev_split_ratio': 0.8,
# 	'output_keep_prob': 0.7,
# 	'padding_idx': None,
# 	'update_embeddings': True,
# 	'embedder_output_keep_prob': 0.7,
# 	'window_sizes': [3, 4, 5],
# 	'number_of_windows': [100, 100, 100]
# } 
```

### Notes related to AdamW optimizer for BERT
- Some references: 
    - https://arxiv.org/pdf/2006.05987.pdf#page=3
    - https://pytorch.org/docs/stable/generated/torch.optim.AdamW.html [[implementation]](https://github.com/pytorch/pytorch/blob/9a622f4cd9318d69896462823891588ef0bf719c/torch/optim/_functional.py#L101) [[resourceful blogpost]](https://johnchenresearch.github.io/demon/)
    - https://huggingface.co/transformers/custom_datasets.html,  https://huggingface.co/transformers/migration.html [[implementation]](https://github.com/huggingface/transformers/blob/9160d81c98854df44b1d543ce5d65a6aa28444a2/src/transformers/optimization.py#L271)
- The tagged paper empirically demonstrates how bias correction is useful. Huggingface now defaults to doing bias_correction which was not done originally in BERT paper. With the default settings, Pytorch's AdamW optimizer too does the same and we use the latter as optimizer for BERT related classifiers.

# Previous PRs:
- [`Phase 1/3 of NN support`](https://github.com/cisco/mindmeld/pull/321): Code refactors to accommodate adding new models. [PPT](https://cisco-my.sharepoint.com/:p:/p/saijayan/EXKyQF_zer1CiLdkkVk4OdUBu5OtFH08Xa2Q-0_y3PFpMA?e=2PYwbE)


mindmeld/components/classifier.py
```diff
@@ -68,15 +68,9 @@ def __init__(
...
-        for arg, val in {"model_type": model_type, "features": features}.items():
+        for arg, val in {"model_type": model_type}.items():
             if val is None:
                 raise TypeError("__init__() missing required argument {!r}".format(arg))
-        if params is None and (
-            param_selection is None or param_selection.get("grid") is None
-        ):
-            raise ValueError(
-                "__init__() One of 'params' and 'param_selection' is required"
-            )
```
murali1996, line 79: Moved model-specific validation of ModelConfig to respective sklearn or pytorch base models.

mindmeld/components/entity_recognizer.py
```diff
@@ -54,7 +54,6 @@ def __init__(self, resource_loader, domain, intent):
         self.domain = domain
         self.intent = intent
         self.entity_types = set()
-        self._model_config = None
```
murali1996, line 57: Looks like there is no much use to have this additional variable which we were previously dumping and loading. Happy to take suggestions on this change. Removing this variable necessitates changes to one unittest in _test\_nlp.py_ file. We do not have this variable saved for RoleClassifiers, wondering what could be the use-case for EntityReconizers.

mindmeld/components/entity_recognizer.py
```diff
@@ -119,16 +118,19 @@ def fit(self,
...
+            self.entity_types = set()
+            for label in labels:
+                for entity in label:
+                    self.entity_types.add(entity.entity.type)
+
+            model = create_model(model_config)
+            model.initialize_resources(self._resource_loader, examples, labels)
+            model.fit(examples, labels)
+            self._model = model
+            self.config = ClassifierConfig.from_model_config(self._model.config)
```
murali1996, line 133: Modified this snippet similar to RoleClassifiers. We don't have to instantiate a model unless there is some training data. Although this will never be the case with latest changes to _nlp.py_ in this PR. The latest changes ensure that we don't call `EntityRecognizer.fit()` when there are zero entity_types in training data.

mindmeld/components/entity_recognizer.py
vembar: In the config, you can specify what file name patterns you want to train the model on. For example, in the case of Webex Assistant, the call_person intent is trained on one subset of files whereas the entity recognizer for that intent is trained on a larger subset of files. As the test in `test_nlp.py` suggests, the config is dumped so that we have access to the actual train files to extract queries from so we can decipher whether it is an incremental build or not. 

I actually think we should be dumping the config for all models. This will help reproduce results since one can change train/test file patterns, feature sets, hyperparam values etc while building an application. So dumping the config along with the model will make things more maintainable.

mindmeld/components/nlp.py
```diff
@@ -1202,7 +1211,7 @@ def _load(self, incremental_timestamp=None):
             self._children[entity_type] = processor
 
     def _evaluate(self, print_stats, label_set="test"):
-        if len(self.entity_recognizer.entity_types) > 1:
+        if len(self.entity_recognizer.entity_types) > 0:
```
vembar, line 1214: Nice catch!

mindmeld/models/containers.py
```diff
@@ -0,0 +1,374 @@
...
+    @staticmethod
+    def _extract_and_map(glove_file):
+        word_to_embedding = {}
+        for line in glove_file:
+            values = line.split()
+            word = values[0]
+            if not isinstance(word, str):  # can be encoded as byte type
+                word = word.decode()
+            coefs = np.asarray(values[1:], dtype="float32")
+            word_to_embedding[word] = coefs
```
vembar, line 143: Should we lower case here? I don't remember if glove is always uncased.

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+
+    def _add_vocab(self, vocab: Iterable[str], target_dict=None):
+        if target_dict is None:
+            target_dict = self.token2idx
+        for token in vocab:
+            if token not in target_dict:
+                target_dict.update({token: len(target_dict)})
+
+
+class PretrainedLangModelBasedEncoder(AbstractEncoder):
```
vembar, line 345: Consider renaming to `PretrainedLMBasedEncoder`

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+                      f"used with the provided 'embedder_type':'{self.embedder_type}'. " \
+                      f"Discarding 'tokenizer_type' information"
+                logger.warning(msg)
+            self.pretrained_model_name_or_path = params.get("pretrained_model_name_or_path")
+            if not self.pretrained_model_name_or_path:
+                msg = "Must include a valid 'pretrained_model_name_or_path' param when using " \
+                      "embedder_type: 'bert'"
+                raise ValueError(msg)
+            self.params_keys.update(["pretrained_model_name_or_path"])
+            if not _disable_loading_embedder:
```
vembar, line 392: What is `_disable_loading_embedder` used for?

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+            new_labels = [self.encode_labels(label, padding_length, add_terminals) for label in
+                          labels]
+            return_dict.update({
+                "labels": torch.as_tensor(new_labels, dtype=torch.long)
+            })
+
+        return return_dict
+
+
+class SeqClsEncoderWithPlmLayer(PretrainedLangModelBasedEncoder):
```
vembar, line 605: `SeqClsEncoderWithPLMLayer`

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+                "seq_lengths": _inputs["attention_mask"].sum(dim=-1)
+            })
+
+        if labels:
+            return_dict.update({"labels": torch.as_tensor(labels, dtype=torch.long)})
+
+        return return_dict
+
+
+class TokenClsEncoderWithPlmLayer(PretrainedLangModelBasedEncoder):
```
vembar, line 642: `TokenClsEncoderWithPLMLayer`

mindmeld/components/entity_recognizer.py
```diff
@@ -151,18 +153,19 @@ def dump(self, model_path, incremental_model_path=None):
...
         er_data = {
             "entity_types": self.entity_types,
-            "w_ngram_freq": self._model.get_resource("w_ngram_freq"),
-            "c_ngram_freq": self._model.get_resource("c_ngram_freq"),
-            "model_config": self._model_config,
         }
+        if self._model:
+            er_data.update({
+                "w_ngram_freq": self._model.get_resource("w_ngram_freq"),
+                "c_ngram_freq": self._model.get_resource("c_ngram_freq"),
```
vijay120, line 160: Add `model_config ` if you agree with Varsha's comments

mindmeld/components/nlp.py
vijay120: (y)

mindmeld/components/nlp.py
```diff
@@ -1286,22 +1295,30 @@ def _recognize_entities(self, query, dynamic_resource=None, verbose=False):
                 return nbest_transcripts_entities
             else:
                 if verbose:
+                    if len(self.entities) == 0:
```
vijay120, line 1298: I think you can add one statement on top of the conditional. I dont think you need to worry about `[[]]` or `[()]`. lets stick with the latter.

mindmeld/components/nlp.py
```diff
@@ -1286,22 +1295,30 @@ def _recognize_entities(self, query, dynamic_resource=None, verbose=False):
...
                 else:
+                    if len(self.entities) == 0:
+                        return [()]
                     return [
                         self.entity_recognizer.predict(
                             query[0], dynamic_resource=dynamic_resource
                         )
                     ]
         if verbose:
+            if len(self.entities) == 0:
```
vijay120, line 1314: same here. lets return an empty tuple.

mindmeld/models/containers.py
```diff
@@ -0,0 +1,374 @@
...
+    """
+    CONTAINER_LOOKUP = {}
+
+    GLOVE_DOWNLOAD_LINK = "http://nlp.stanford.edu/data/glove.6B.zip"
+    EMBEDDING_FILE_PATH_TEMPLATE = "glove.6B.{}d.txt"
+    ALLOWED_WORD_EMBEDDING_DIMENSIONS = [50, 100, 200, 300]
+
+    def __init__(self, token_dimension=300, token_pretrained_embedding_filepath=None):
+        # validations
+        if token_dimension not in GloVeEmbeddingsContainer.ALLOWED_WORD_EMBEDDING_DIMENSIONS:
```
vijay120, line 66: I think we should fail explicitly here instead of downloading a default size.

mindmeld/models/containers.py
```diff
@@ -0,0 +1,374 @@
...
+                )
+                return
+
+            file_name = GloVeEmbeddingsContainer.EMBEDDING_FILE_PATH_TEMPLATE.format(
+                self.token_dimension)
+            zip_file_object = zipfile.ZipFile(EMBEDDINGS_FILE_PATH, "r")
+
+            if file_name not in zip_file_object.namelist():
+                logger.info(
+                    "Embedding file with %s dimensions " "not found",
```
vijay120, line 127: "Embedding file with %s dimensions not found in %s file path"

mindmeld/models/model.py
```diff
@@ -350,12 +366,16 @@ def register_resources(self, **kwargs):  # pylint: disable=no-self-use
...
-    def _get_model_config_save_path(path):
-        head, ext = os.path.splitext(path)
-        model_config_save_path = head + ".config" + ext
-        os.makedirs(os.path.dirname(model_config_save_path), exist_ok=True)
-        return model_config_save_path
+    @property
+    def tokenizer(self):
+        tokenizer = self._resources.get("tokenizer")
+        if not tokenizer:
+            logger.error(
```
vijay120, line 373: I think a warning here is suffient

mindmeld/models/containers.py
```diff
@@ -0,0 +1,374 @@
...
+        self.model_id = Hasher(algorithm="sha1").hash(string=string_to_hash)
+
+        self._reload = reload
+        self._model_bunch = self._extract_model()
+
+    def get_model_bunch(self):
+        return self._model_bunch
+
+    def _extract_model(self):
+        if self.model_id not in HuggingfaceTransformersContainer.CONTAINER_LOOKUP or self._reload:
```
vijay120, line 353: Would prefer if we did the opposite conditional:
```
if not self.model_id not in HuggingfaceTransformersContainer.CONTAINER_LOOKUP and not self._reload:
    return HuggingfaceTransformersContainer.CONTAINER_LOOKUP[self.model_id]
.
.

``

File: mindmeld/components/nlp.py (hunk 4 above)
Comment (murali1996) [reply mindmeld/components/nlp.py:1214]: :)

File: mindmeld/models/nn_utils/input_encoders.py [hunk 17]
```diff
@@ -0,0 +1,937 @@
...
+        # Add special vocab before actual vocab
+        special_vocab_dict_ = self.__class__.BASIC_SPECIAL_VOCAB_DICT
+        if self.special_vocab_dict:
+            special_vocab_dict_.update(self.special_vocab_dict)
+        self.special_vocab_dict = dict(special_vocab_dict_)
+        self._add_special_vocab(self.special_vocab_dict)
+
+        # add vocab from tokens inputted directly (useful when loading the encoder from a dump)
+        if tokens:
+            self._add_vocab(tokens)
```
vembar, line 230: What is the use-case for this?

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+        return self.pad_token_idx
+
+    def get_embedding_weights(self):
+        if not hasattr(self, "token2emb") or not self.token2emb:
+            msg = f"Encoder instance ({self.name}) does not contain any token-to-embeddings mapping"
+            logger.info(msg)
+            return None
+        embedding_weights = {}
+        for token, idx in self.token2idx.items():
+            if token in self.token2emb:
```
vembar, line 97: If a word is not present in the embedding dict, shouldn't we initialize a random vector? What happens in case of special tokens/tokens that are user defined?

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+                    msg = f"Number of tokens in a sentence ({len(ex_tokens)}) must be same as the" \
+                          f"number of tokens in the corresponding token labels " \
+                          f"({len(label_tokens)}) for sentence '{ex}' with labels '{labels}'"
+                    raise AssertionError(msg)
+
+        return examples, labels
+
+    @staticmethod
+    def _trim_combined(x: List[List[Any]], trim_len: int, y: List[Any] = None):
+        curr_len = 0
```
vembar, line 661: Can we get a docstring explaining what this function is doing?

mindmeld/models/nn_utils/nn_base_modules.py
```diff
@@ -0,0 +1,386 @@
...
+        "number_of_epochs": 100,
+        "patience": 10,
+        "batch_size": 32,
+        "gradient_accumulation_steps": 1,
+        "max_grad_norm": None,
+        "optimizer": "Adam",
+        "learning_rate": 0.001,
+        "validation_metric": "accuracy",  # or 'f1'
+        "verbose": True,  # to print progress to stdout/log file
+        "dev_split_ratio": 0.8
```
vembar, line 66: If we're naming the key `dev_split_ratio`, shouldn't the value be 0.2 ?

mindmeld/models/nn_utils/nn_base_modules.py
```diff
@@ -0,0 +1,386 @@
...
+        "device": "cuda" if is_cuda_available else "cpu",
+        "number_of_epochs": 100,
+        "patience": 10,
+        "batch_size": 32,
+        "gradient_accumulation_steps": 1,
+        "max_grad_norm": None,
+        "optimizer": "Adam",
+        "learning_rate": 0.001,
+        "validation_metric": "accuracy",  # or 'f1'
+        "verbose": True,  # to print progress to stdout/log file
```
vembar, line 65: I think the log level should handle this consistent with other models

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+            return_dict.update({
+                "labels": torch.as_tensor(labels, dtype=torch.long)
+            })
+
+        return return_dict
+
+
+class TokenClsEncoderForEmbLayer(WordEmbeddingsBasedEncoder):
+
+    def _reformat_and_validate(self, examples, labels=None):
```
vembar, line 525: Can't we move this to `PyTorchTaggerModels` since this only changes with the task and not the Encoder type?

mindmeld/components/entity_recognizer.py
vijay120: Looks like we can remove it.

mindmeld/components/entity_recognizer.py
vijay120: Disregard this.

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+        for token, idx in self.token2idx.items():
+            if token in self.token2emb:
+                embedding_weights[idx] = self.token2emb[token]
+        return embedding_weights
+
+    @staticmethod
+    def get_pad_label_idx():
+        return LABEL_PAD_TOKEN_IDX
+
+    def _reformat_and_validate(self, examples, labels=None):
```
vijay120, line 105: type hints

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+        os.makedirs(ckpt_folder, exist_ok=True)
+        self._bert_config.save_pretrained(ckpt_folder)
+        self._bert_tokenizer.save_pretrained(ckpt_folder)
+        self._bert_model.save_pretrained(ckpt_folder)
+
+
+# sub-classes that define a custom `batch_encode` on top of base encoders
+
+
+class SeqClsEncoderForEmbLayer(WordEmbeddingsBasedEncoder):
```
vijay120, line 475: I think we should spell the whole name out

vijay120, line 475: Its not immediately clear why SeqClsEncoderForEmbLayer inherit from WordEmbeddingsBasedEncoder. I thought it should be the other way, since SeqClsEncoderForEmbLayer outputs ints and WordEmbeddingsBasedEncoder converts those ints to vectors

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+
+        # if padding_length is None, it is computed as max(length of all seqs from inputted text)
+        padding_length = padding_length or self.padding_length
+        # add 2 for start and end tokens respectively if add_terminals is True
+        add_terminals = add_terminals or self.add_terminals
+        curr_max = max(sequence_lengths) + 2 if add_terminals else max(sequence_lengths)
+        padding_length = min(padding_length, curr_max) if padding_length else curr_max
+
+        # batchify by truncating or padding for each example
+        seq_ids, sequence_lengths = zip(*[
```
vijay120, line 505: We should try to use generators/iterators to do this since its memory inefficient

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+
+class TokenClsEncoderForEmbLayer(WordEmbeddingsBasedEncoder):
+
+    def _reformat_and_validate(self, examples, labels=None):
+        super()._reformat_and_validate(examples, labels)
+
+        # validation for labels' lengths too
+        if labels:
+            for ex, label_tokens in zip(examples, labels):
+                ex_tokens = ex.split(" ")
```
vijay120, line 531: Arent we assuming whitespace tokenization here?

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+    ):
+        examples, labels = self._reformat_and_validate(examples, labels)
+
+        return_dict = {}
+        padding_length = padding_length or self.padding_length
+
+        if self.embedder_type == "bert":
+            # tokenize each word of each input seperately
+            tokenized_examples = [
+                [self._bert_tokenizer.tokenize(word) for word in example.split(" ")]
```
vijay120, line 699: why do we use whitespace splitting here?

mindmeld/models/nn_utils/input_encoders.py
vijay120: More docs here please. Its not clear what PLM is here. Also expand the name

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+            else:
+                trimmed_tokenized_examples = [
+                    self._trim_combined(t_ex, _trim_length) for t_ex in tokenized_examples
+                ]
+            split_lengths = [[len(x) for x in ex] for ex in trimmed_tokenized_examples]
+            trimmed_tokenized_examples = [" ".join(sum(ex, [])) for ex in
+                                          trimmed_tokenized_examples]
+            _inputs = self._bert_tokenizer(
+                trimmed_tokenized_examples, padding=True, truncation=True, max_length=None,
+                return_tensors="pt"
```
vijay120, line 722: what does `pt` mean?

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+
+# sub-classes that define more complex custom `batch_encode` on top of derived encoders
+
+
+class TokenClsDualEncoderForEmbLayers(TokenClsEncoderForEmbLayer):
+    """Dual encoder that encodes both word level as well as character level tokens
+    """
+
+    ALLOWED_TOKENIZERS = [None, "whitespace-tokenizer"]
+    BASIC_SPECIAL_CHAR_VOCAB_DICT = {
```
vijay120, line 757: SPECIAL_CHAR_VOCAB_MAPPING

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+        "char_unk_token": "<CHAR_UNK>",
+        "char_start_token": "<CHAR_START>",
+        "char_end_token": "<CHAR_END>",
+    }
+    DEFAULT_PARAMS = {
+        "emb_dim": 300,
+        "char_emb_dim": 50
+    }
+
+    def fit(
```
vijay120, line 768: I think the fit method should be named something else here. its not really doing a fit in a ML sense

mindmeld/models/nn_utils/layers.py
```diff
@@ -0,0 +1,269 @@
...
+SEED = 7246
+
+logger = logging.getLogger(__name__)
+
+
+# utils
+
+def get_disk_space_of_model(pytorch_module):
+    filename = "temp.bin"
+    torch.save(pytorch_module.state_dict(), filename)
```
vijay120, line 39: use sys.getsizeof instead of writing to disk

mindmeld/models/nn_utils/layers.py
```diff
@@ -0,0 +1,269 @@
...
+    size = os.path.getsize(filename) / 1e6
```
vijay120, line 40: not sure why 1e6 is being used here

mindmeld/models/nn_utils/layers.py
```diff
@@ -0,0 +1,269 @@
...
+    os.remove(filename)
+    return size
+
+
+def get_num_weights_of_model(pytorch_module):
+    n_total = 0
```
vijay120, line 46: doc string with explaination on why we are quantifying both grad and total params. also doc strings with return params documented

mindmeld/models/nn_utils/layers.py
```diff
@@ -0,0 +1,269 @@
...
+                    self.embeddings.weight.data[idx] = torch.as_tensor(emb)
+                msg = f"Initialized {len(embedding_weights)} number of embedding weights " \
+                      f"from the embedder model"
+                logger.info(msg)
+            else:
+                # when weights are passed as an array or tensor
+                self.embeddings.load_state_dict({'weight': torch.as_tensor(embedding_weights)})
+        self.embeddings.weight.requires_grad = update_embeddings
+
+        self.embedding_for_coefficients = None
```
vijay120, line 85: add docs on what this var is

mindmeld/models/nn_utils/layers.py
```diff
@@ -0,0 +1,269 @@
...
+
+# Various nn layers
+
+
+class EmbeddingLayer(nn_module):
+    """A pytorch wrapper layer for embeddings that takes input as a batched sequence of ids
+    and outputs embeddings correponding to those ids
+    """
+
+    def __init__(self, num_tokens, emb_dim, padding_idx=None,
```
vijay120, line 66: add docstring on what each argparam is

mindmeld/models/nn_utils/layers.py
```diff
@@ -0,0 +1,269 @@
...
+                  f"Expected a list of same length as emb_dim ({len(emb_dim)})"
+            raise ValueError(msg)
+
+        self.convs = nn.ModuleList()
+        # Unsqueeze input dim [BS, SEQ_LEN, EMD_DIM] to [BS, 1, SEQ_LEN, EMDDIM] and send as input
+        # Each conv module output's dimensions are [BS, n, SEQ_LEN, 1]
+        for length, n in zip(kernel_sizes, num_kernels):
+            self.convs.append(
+                nn.Sequential(
+                    nn.Conv2d(1, n, (length, emb_dim), padding=(length - 1, 0),
```
vijay120, line 137: what does `padding=(length - 1, 0)` mean? The torch does say its the implicit padding on each side, I am not entirely sure what that means here

vijay120 (commented):
part 1 of reviews

mindmeld/models/model.py
```diff
@@ -350,12 +365,19 @@ def register_resources(self, **kwargs):  # pylint: disable=no-self-use
...
-        os.makedirs(os.path.dirname(model_config_save_path), exist_ok=True)
-        return model_config_save_path
+    @property
+    def text_preparation_pipeline(self):
+
+        text_preparation_pipeline = self._resources.get("text_preparation_pipeline")
+
+        if not text_preparation_pipeline:
+            logger.error(
+                "The text_preparation_pipeline resource has not been registered "
```
KKsharma99, line 375: This warning may get confusing to developers since they would think it is referring to the 'global' text_preparation_pipeline

mindmeld/models/model_factory.py
```diff
@@ -0,0 +1,186 @@
...
+                instance of ModelConfig
+
+        Returns:
+            model (Type[AbstractModel]): A text/tagger model instance
+
+        Raises:
+            ValueError: When the configs are invalid
+        """
+
+        if not (model_config and isinstance(model_config, (ModelConfig, dict))):
```
KKsharma99, line 68: Nit: Consider for improved readability

```
is_valid_model = model_config and isinstance(model_config, (ModelConfig, dict))
if not is_valid_model:
      // CODE
```

mindmeld/models/model_factory.py
```diff
@@ -0,0 +1,186 @@
...
+                model_config)
+            return model_class.load(path)
+
+        except FileNotFoundError:
+            # sometimes a model (and its config file) might not be dumped, eg. in role classifiers
+            # or even if dumped, can be of NoneType enclosed in a dictionary
+            msg = f"No model file found while trying to load model from path: {path}. It might " \
+                  f"be the case that the classifier didn't need a model due to one or no classes."
+            logger.warning(msg)
+            return None
```
KKsharma99, line 126: Consider removing the `return None`. Also should we throw an error if the file is not found instead of silently returning None?

mindmeld/models/nn_utils/layers.py
```diff
@@ -0,0 +1,259 @@
...
+
+        return outputs
+
+
+class CnnLayer(nn_module):
+
+    def __init__(self, emb_dim, kernel_sizes, num_kernels):
+        super().__init__()
+
+        if isinstance(num_kernels, list) and len(num_kernels) != len(kernel_sizes):
```
KKsharma99, line 95: Can you please clarify why we are taking the `len(num_kernels)`? Since 'num_var' normally refers to an int. Would be helpful here to add a comment with the type of `num_kernels` or rename it

mindmeld/models/model.py
```diff
@@ -729,27 +740,79 @@ def _validate_model_configs(self) -> Union[TypeError, ValueError]:
...
+        self._class_encoder = SKLabelEncoder()
+        self._query_text_type = None
+        self._clf = None
 
     def initialize_resources(self, resource_loader, examples=None, labels=None):
-        raise NotImplementedError
+        del resource_loader, examples, labels
+        # self._resources[
+        #     "text_preparation_pipeline"
+        # ] = resource_loader.get_text_preparation_pipeline()
```
vembar, line 758: Remove commented code

mindmeld/models/model.py
```diff
@@ -350,12 +365,19 @@ def register_resources(self, **kwargs):  # pylint: disable=no-self-use
...
         return self._resources.get(name)
 
-    @staticmethod
-    def _get_model_config_save_path(path):
-        head, ext = os.path.splitext(path)
-        model_config_save_path = head + ".config" + ext
-        os.makedirs(os.path.dirname(model_config_save_path), exist_ok=True)
-        return model_config_save_path
+    @property
+    def text_preparation_pipeline(self):
```
vembar, line 369: Unsure why we need this here. Isn't the input to the models already a list of ProcessedQueries?

mindmeld/models/nn_utils/sequence_classification.py
```diff
@@ -0,0 +1,394 @@
...
+    def _forward_with_batching_and_no_grad(self, examples):
+        logits = None
+        was_training = self.training
+        self.eval()
+        with torch.no_grad():
+            for start_idx in range(0, len(examples), self.params.batch_size):
+                this_examples = examples[start_idx:start_idx + self.params.batch_size]
+                batch_data = self.encoder.batch_encode(this_examples)
+                this_logits = self.forward(batch_data)["logits"]
+                logits = torch.cat((logits, this_logits)) if logits is not None else this_logits
```
vembar, line 138:
```suggestion
                batch_examples = examples[start_idx:start_idx + self.params.batch_size]
                batch_encoded = self.encoder.batch_encode(batch_examples)
                batch_logits = self.forward(batch_encoded)["logits"]
                logits = torch.cat((logits, batch_logits)) if logits is not None else batch_logits
```

mindmeld/models/nn_utils/sequence_classification.py
```diff
@@ -0,0 +1,394 @@
...
+                self.params.embedder_output_pooling_type
+            )
+        self.out_dim = self.params.emb_dim
+
+    def _forward_core(self, batch_data):
+
+        bert_outputs = self.bert_model(**batch_data["hgf_encodings"], return_dict=True)
+
+        if self.params.embedder_output_pooling_type != "first":
+            last_hidden_state = bert_outputs.get("last_hidden_state")  # [BS, SEQ_LEN, EMD_DIM]
```
vembar, line 365: Can you add a comment here that `last_hidden_state` actually means the output of the last layer of the model?

mindmeld/models/nn_utils/classification.py
```diff
@@ -0,0 +1,490 @@
...
+                ):
+                    # update weights when it is the last batch in the epoch or
+                    # when specified step is reached or
+                    if self.params.max_grad_norm:  # clip (accumlated) gradients if required
+                        nn.utils.clip_grad_norm_(self.parameters(), self.params.max_grad_norm)
+                    optimizer.step()
+                    scheduler.step()
+                    optimizer.zero_grad()
+                progress_bar_msg = f"Epoch: {epoch} | Mean loss: " \
+                                   f"{train_loss / (start_idx / self.params.batch_size + 1):.4f}"
```
vembar, line 230: `train_loss/train_batches`?

mindmeld/models/nn_utils/classification.py
```diff
@@ -0,0 +1,490 @@
...
+
+        # update params upon preparing encoder and embedder
+        params = {**self.get_default_params(), **params}
+        params = self._prepare_input_encoder(examples, **params)
+        params = self._prepare_embedder(**params)
+
+        # update params upon identifying unique labels
+        try:
+            num_labels = len(set(labels))
+        except TypeError:  # raised in cased on token classification
```
vijay120, line 129: typo: in case

mindmeld/models/nn_utils/classification.py
```diff
@@ -0,0 +1,490 @@
...
+            num_labels = len(set(sum(labels, [])))
```
vijay120, line 130: Not clear how this solve the type issue

mindmeld/models/nn_utils/classification.py
```diff
@@ -0,0 +1,490 @@
...
+        self.ready = False  # True when .fit() is called or loaded from a checkpoint, else False
+        self.dirty = False  # True when the model weights aren't saved to disk yet, else False
+
+    def __repr__(self):
+        return f"<{self.name}> ready:{self.ready} dirty:{self.dirty}"
+
+    def get_default_params(self) -> Dict:
+        return get_default_params(self.__class__.__name__)
+
+    def who_am_i(self) -> str:
```
vijay120, line 85: print_and_return_system_config

mindmeld/models/nn_utils/classification.py
```diff
@@ -0,0 +1,490 @@
...
+        self.who_am_i()
+
+        # dumping weights during training process into a temp folder instead of keeping in
+        # memory to reduce memory usage
+        temp_folder = os.path.join(USER_CONFIG_DIR, "tmp", "pytorch_models", str(uuid.uuid4()))
+        temp_weights_save_path = os.path.join(temp_folder, "pytorch_model.bin")
+        os.makedirs(temp_folder, exist_ok=True)
+
+        # split input data into train & dev splits, and get data loaders
+        indices = np.arange(len(examples))
```
vijay120, line 155: Could we use this https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html

mindmeld/models/nn_utils/classification.py
```diff
@@ -0,0 +1,490 @@
...
+                this_labels_targetted = dev_labels[start_idx:start_idx + self.params.batch_size]
+                # validation
+                if len(this_labels_predicted) != len(this_labels_targetted):
+                    msg = f"Number of predictions ({len(this_labels_predicted)}) " \
+                          f"not equal to number of targets ({len(this_labels_targetted)})"
+                    logger.error(msg)
+                    raise AssertionError(msg)
+                # flatten if required
+                try:
+                    this_labels_predicted = sum(this_labels_predicted, [])
```
vijay120, line 248: I am unclear what this computation accomplishes

murali1996: Because the fit method is used by both sequence classification as well as sequence labeling tasks, this step is required to flatten sequence labeling task's results (i.e from [[], [], ..., []] --> [...]). Once flattened, the results data structure is same in case of both tasks and hence a generic way of scoring is possible.

mindmeld/models/nn_utils/classification.py
murali1996: changed to `print_and_return_model_info`

mindmeld/models/nn_utils/classification.py
murali1996: Similar to the [other related comment below](https://github.com/cisco/mindmeld/pull/325#discussion_r771063528)

mindmeld/models/model.py
murali1996: Here's the use-case I found (something that has been there wrt sklearn models classes we have already):
- in `view_extracted_features()`, `predict()`, `predict_proba()`methods of _tagger\_models.py_,  we refer to `self.text_preparation_pipeline` 
- similarly in `get_feature_matrix()` and `inspect()` methods in _text\_models.py_'s sklearn class too.

mindmeld/models/model_factory.py
murali1996: Hey Kunal, so I had to return None and deal with the result (NoneType or not) directly in components/ folder because of backwards compatibilities wrt the way we store and load sklearn models. I specifically set the `return None` because a NoneType output here explicitly means no model has been found and the classifier.py, role_classifier.py and entity_recognizer.py utilizes this info. 

The reason I did not raise the error and catch it as a means to obtain the same info is because, in mindmeld's architecture, helpers.py's load_model() acts as a bridge between components/ files and models/ files and the two modules don't communicate directly nor have to, hence no dependencies. So in order to propogate the presence of model information from models/ to components/, one design choice i thought is that the components/ files know that a FileNotFoundError is raised by models/ files and do a try-catch for that specific Error type, but then what if we change the Error type in future, we'll not be able to track and debug the dependency created across modules right? Alternate choice i thought is to just return a NoneType, which can remain same irrespective of what error type we create. This will not make the components/ files dependent on the choices of models/ files and will be easy to debug later on. Creating a dummy model class for this purpose seemed to complicate things and I refrained from that idea.

What do you think of this? Please let me know if you think I can do a better solution. And Thanks for pointing this out.

mindmeld/models/model.py
murali1996: hey @KKsharma99 this actually refers to global one only. Upon some digging, I found the following snippet in `initialize_resources()` methods in model.py file. So in absence of a global text_preparation_pipeline, a new one is created and hence the error message. (Also, this message already existed in the codebase and is not something that I newly added. )

```
        # Always initialize the global resource for tokenization, which is not a
        # feature-specific resource
        self._resources[
            "text_preparation_pipeline"
        ] = resource_loader.get_text_preparation_pipeline()
```

mindmeld/models/nn_utils/layers.py
murali1996: Sure, I'll see how I can change. I will provide type hints that num_kernels is a List specifying number of kernels per kernel size specified.

mindmeld/models/nn_utils/helpers.py
```diff
@@ -0,0 +1,43 @@
...
+except ImportError:
+    pass
+
+logger = logging.getLogger(__name__)
+
+
+# utils
+
+def get_disk_space_of_model(pytorch_module):
+    filename = "temp.bin"
```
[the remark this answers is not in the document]
murali1996: > You can use sys.getsizeof instead of writing to disk to check size of the object.

based on the description here https://docs.python.org/2/library/sys.html#sys.getsizeof and from simple check with a bert-base model, found both solutions gives different results

- using `sys.getsizeof`: 19880
- using `get_disk_space_of_model`: 433325247 (i.e. 433 MB) 

the latter being correct result!

> This will prevent unnecessary spikes in memory to write to disk
The quantity of memory is anyway used when actually saving the pytorch model. So in that sense, it is not using additional disk space anyway. But I'll make sure that this computation of disk space method is called only when logging level is set to info or if the verbose flag is on.

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+
+logger = logging.getLogger(__name__)
+
+LABEL_PAD_TOKEN_IDX = -1  # value set based on default label padding idx in pytorch
+
+
+# abstract encoder; derives from torch nn.Module
+
+
+class AbstractEncoder(nn_module):
```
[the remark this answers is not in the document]
murali1996: Added!

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+    def get_num_tokens(self):
+        if not hasattr(self, "token2idx"):
+            return None
+        return len(self.token2idx)
+
+    def get_emb_dim(self):
+        return self.emb_dim
+
+    def get_pad_token_idx(self):
+        if not hasattr(self, "pad_token_idx"):
```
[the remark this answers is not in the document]
murali1996: Added

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+        "pad_token": "<PAD>",
+        "unk_token": "<UNK>",
+        "start_token": "<START>",
+        "end_token": "<END>",
+    }
+    DEFAULT_PARAMS = {
+        "emb_dim": 300,
+    }
+
+    def fit(
```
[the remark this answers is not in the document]
murali1996: while this is not an ML model, encoders are might sometimes need training like in case of sentence pice or BPE. training here means creating a stateful object which splits inputs based on some algo that is learnt with the trained data.

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+        elif self.embedder_type == "glove":
+            self.tokenizer_type = params.get("tokenizer_type")
+            if self.tokenizer_type not in [None, "whitespace-tokenizer"]:
+                msg = f"Provided 'tokenizer_type':'{params.get('tokenizer_type')}' cannot be " \
+                      f"used with the provided 'embedder_type':'{self.embedder_type}'. " \
+                      f"Discarding 'tokenizer_type' information"
+                logger.warning(msg)
+                self.tokenizer_type = "whitespace-tokenizer"
+            self._tokenizer = self._get_tokenizer(self.tokenizer_type)
+            self.params_keys.update(["tokenizer_type"])
```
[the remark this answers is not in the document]
murali1996: Moved params_keys from encoders module to the clasification modules within nn_utils.

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+            if not _disable_loading_embedder:
+                glove_container = GloVeEmbeddingsContainer()
+                self.token2emb = glove_container.get_pretrained_word_to_embeddings_dict()
```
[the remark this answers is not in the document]
murali1996: Thats the reason I modified the codebase of containers.py (GloVeEmbeddingsContainer, SentenceTransformersContainer and HuggingfaceTransformersContainer) with a class variable `CONTAINER_LOOKUP` that acts as a cache. Also made sure that there is a feature to instantiate a separate object and not retrieve from the lookup if need be (eg. we might want one huggingface bert model for each of our domain classifiers as they are trained from same checkpoint but will lead to multiple trained models, one per domain)

mindmeld/models/nn_utils/input_encoders.py
```diff
@@ -0,0 +1,937 @@
...
+        self.special_vocab_dict = dict(special_vocab_dict_)
+        self._add_special_vocab(self.special_vocab_dict)
+
+        # add vocab from tokens inputted directly (useful when loading the encoder from a dump)
+        if tokens:
+            self._add_vocab(tokens)
+
+        # add vocab from examples upon tokenizing each text item
+        if examples:
+            all_tokens = set(sum([self._tokenizer(text) for text in examples], []))
```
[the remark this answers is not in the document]
murali1996: Sure, replaced the following
```
       all_tokens = dict.fromkeys(sum([self._tokenize(text) for text in examples], []))
```
with this snippet
```
        all_tokens = set()
        for text in examples:
            tokens = [token for token in self._tokenize(text) if token not in all_tokens]
            all_tokens.update(tokens)
```

mindmeld/models/nn_utils/input_encoders.py
murali1996: modified and the assumption is not included anymore

mindmeld/models/nn_utils/input_encoders.py
murali1996: Common notation used in huggingface transformers.

mindmeld/models/nn_utils/input_encoders.py
murali1996: variable removed after refactors

mindmeld/models/nn_utils/layers.py
murali1996: replied in this comment <https://github.com/cisco/mindmeld/pull/325#discussion_r771109109>

mindmeld/models/nn_utils/input_encoders.py
murali1996: After some refactors, felt `prepare` to be a suitable name for input encoders. Modified to it from `fit`

mindmeld/models/containers.py
```diff
@@ -0,0 +1,462 @@
...
+                        raise EmbeddingDownloadError("Failed to download embeddings.")
+                    with zip_file_object.open(file_name) as embedding_file:
+                        word_to_embedding = self._extract_and_map(embedding_file)
+
+            GloVeEmbeddingsContainer.CONTAINER_LOOKUP[self.model_id] = word_to_embedding
+
+        return GloVeEmbeddingsContainer.CONTAINER_LOOKUP[self.model_id]
+
+
+class SentenceTransformersContainer:
```
ritvikshrivastava, line 202: Do you think we can add some consistency to the containers file? Like an abstract class with the extraction, get_model, encoder functionalities etc. set up in a default class and then inherited by these different model classes? Might be good to have that base structure in place for adding containers or modifying overall structure across the set of containers at a later point.

mindmeld/models/nn_utils/classification.py
ritvikshrivastava, line 85: Can this be model_status or a little more informative variable name?

mindmeld/models/nn_utils/classification.py
```diff
@@ -0,0 +1,490 @@
...
+            num_labels = len(set(labels))
+        except TypeError:  # raised in cased on token classification
+            num_labels = len(set(sum(labels, [])))
+        params.update({"num_labels": num_labels})
+
+        # update params with label pad idx
+        try:  # used by token classifiers in crf masks and for training/evaluation
+            self.label_padding_idx = params["label_padding_idx"]
+            params.update({"label_padding_idx": params["label_padding_idx"]})
+        except KeyError:  # raised in cased on sequence classification
```
ritvikshrivastava, line 137: Nit: Can you elaborate on what this exception is exactly (what does missing label_padding_idx signify maybe?) and which fail gets it here? Not sure if the comment is a typo but not completely clear!

mindmeld/models/nn_utils/classification.py
```diff
@@ -0,0 +1,490 @@
...
+                break
+            # set modules to train phase, reset gradients, do forward-backward propogations
+            self.train()
+            optimizer.zero_grad()
+            train_loss, train_batches = 0.0, 0.0
+            t = tqdm(
+                range(0, len(train_examples), self.params.batch_size), disable=not verbose
+            )
+            for start_idx in t:
+                this_examples = train_examples[start_idx:start_idx + self.params.batch_size]
```
ritvikshrivastava, line 199:
```suggestion
                batch_examples = train_examples[start_idx:start_idx + self.params.batch_size]
```

rename this_* to batch_*?

mindmeld/models/nn_utils/classification.py
ritvikshrivastava: You can also use chaining from itertools (``from itertools import chain``) to achieve the same, that's supposedly faster for longer lists. I'm considering making the same switch in the AL PR. :) 

```suggestion
                    this_labels_predicted = chain.from_iterable(this_labels_predicted)
```

mindmeld/models/nn_utils/classification.py
murali1996: yup, did that and should be seen in next push!

mindmeld/models/nn_utils/classification.py
murali1996: so in case of token classification, for the labels, we pass a label pad id which is passed to cross entropy loss. However, for seq classification (i.e sentence level text classification), no such idx is required and hence we wont set it. 

Because we have a single fit() method irrespective of the seq or token classification, we are just checking if that pad idx is available or not and if not, we just ignore.

I'll modify the comment to make it clear.

mindmeld/models/nn_utils/classification.py
murali1996: yes, changed in next push!

mindmeld/models/containers.py
murali1996: Yes, we can definitely do that. I worked on it briefly sometime back but kept it as low priority for now because wherever we are using these containers in our codebase, we mostly don't use them interchangeably, meaning we might not need a factory like class structure. Hence I made sure that at least the skeleton of all containers look similar so even if we plan to refactor like you mentioned, it should be mostly staright-forward.

mindmeld/models/nn_utils/classification.py
murali1996: oh great! thanks for mentioning that @ritvikshrivastava

mindmeld/models/text_models.py
```diff
@@ -491,14 +493,134 @@ def _dump(self, path):
...
+        model_eval = StandardModelEvaluation(self.config, evaluations)
+        return model_eval
+
+    def fit(self, examples, labels, params=None):
+
+        if len(set(labels)) <= 1 or not examples:
+            return self
+
+        if not isinstance(examples, PQL.QueryIterator):
+            # pytorch text models are not implemented for role-classifiers, which pass-in an
```
vembar, line 538: I think we should add a check and error message saying NN models are not available for role classifiers if the user selects one in the config. 

I guess this error message won't help the end user identify the problem.

mindmeld/query_factory.py
```diff
@@ -191,9 +191,9 @@ def create_query_factory(
         elif duckling:
             sys_entity_recognizer = DucklingRecognizer.get_instance()
         else:
-            logger.warning(
-                "No System Entity Recognizer selected, set 'duckling=True' for DucklingRecognizer",
-            )
+            # logger.warning(
```
vembar, line 194: Think this is unintended

murali1996: oh yeah!! Thanks for pointing.

mindmeld/models/text_models.py
murali1996: Sure, I couldn't find a better spot to raise this error. I will modify the error message written here so that we hint the user about using NNs for role classification.

mindmeld/models/text_models.py
```diff
@@ -536,10 +536,11 @@ def fit(self, examples, labels, params=None):
...
         if not isinstance(examples, PQL.QueryIterator):
             # pytorch text models are not implemented for role-classifiers, which pass-in an
-            # instance of ListIterator to fit() method as opposed to QueryIterator in case of
+            # instance of ListIterator to this fit() method as opposed to QueryIterator in case of
             # domain- and intent-classifiers
-            msg = f"{self.__class__.__name__}.fit() only accepts QueryIterator type examples " \
-                  f"argument but the inputted type is {type(examples)} "
+            msg = f"{self.__class__.__name__}.fit() only accepts QueryIterator as the first " \
+                  f"argument but found type: {type(examples)}. This might happen if trying to" \
+                  f"create a deep neural net based classifier for role classification."
```
vembar, line 543:
```suggestion
                  f"create a deep neural net based classifier for role classification which is currently not supported."
```

vembar (approved):
Looks okay. Let's make sure we test it thoroughly!

mindmeld/models/helpers.py
```diff
@@ -191,6 +200,7 @@ def register_model(model_type, model_class):
         model_type (str): The model type as specified in model configs
         model_class (class): The model to register
     """
+    # TODO: deprecate MODEL_MAP var in in lieu of ModelFactory
```
ritvikshrivastava, line 203: Is this for the current PR or a later change?

mindmeld/models/nn_utils/classification.py
```diff
@@ -0,0 +1,559 @@
...
+            elif isinstance(v, list):
+                batch_data[k] = [
+                    vv.to(self.params.device) if isinstance(vv, torch.Tensor) else vv
+                    for vv in v
+                ]
+            elif isinstance(v, dict):
+                batch_data[k] = self.to_device(batch_data[k])
+        return batch_data
+
+    # pylint: disable=too-many-locals
```
ritvikshrivastava, line 127: Would it help to break this function down into a helper and main. The main loop can go in a ``_fit`` to contain the local variables and easier readability.

mindmeld/models/nn_utils/sequence_classification.py
```diff
@@ -0,0 +1,448 @@
...
+        elif self.params.num_labels > 2:
+            # cross-entropy criterion
+            self.classifier_head = nn.Linear(self.out_dim, self.params.num_labels)
+            self.criterion = nn.CrossEntropyLoss(reduction='mean')
+        else:
+            msg = f"Invalid number of labels specified: {self.params.num_labels}. " \
+                  f"A valid number is equal to or greater than 2"
+            raise ValueError(msg)
+
+        print(f"{self.name} is initialized")
```
ritvikshrivastava, line 91: Should this be logged instead?

murali1996: yes, corrected that :)

mindmeld/models/model.py
```diff
@@ -335,24 +363,40 @@ def _dump(self, path) -> None:
...
+        # Not implemeneted unless overwritten by child class
         raise NotImplementedError
 
     def register_resources(self, **kwargs):  # pylint: disable=no-self-use
         # Resources for feature extractors are not required for deep neural models
         del kwargs
         pass
 
-    def get_resource(self, name):
+    def get_resource(self, name) -> Any:
```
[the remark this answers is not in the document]
murali1996: I couldn't find a particular data structure that can be associated with all resources. So I left it as a general data structure.

mindmeld/models/helpers.py
murali1996: for a later PR

PR #328, opened 2021-06-22
Feature/MultiLingual-Tokenizer

KKsharma99:
PR for the second half of changes related to the new MultiLingual Tokenizer.

Presentation: https://cisco-my.sharepoint.com/:p:/p/kunshar2/Ea1nCs23oKpEjZ3BxHB3qaMBI8-JmD6xPFq_ndVytLUqzQ?e=Hg6Rqq

Tokenizer Architecture

- [x] Relocate stemmers.py and preprocessor.py into the text_preparation module and update all references to related classes and tests.  
- [x] Create Mindmeld Annotation Detection Regex and entity text extraction functions
- [x] Create the TextPreparationPipeline Class combining Preprocessors, Normalizers, Tokenizers, and Stemmers
- [x] Create the TextPreparationPipeline Factory which converts text config into instantiated instances, create_from_app_path()
- [x] Allow for a TextPreparation Class to be passed in from the top level. 
- [x] Modify the QueryFactory to use a single text_preparation_pipeline object instead of separate preprocessor, stemmer, and tokenizers.
- [x] Build a RegexNormalizers which uses regex_norm_rules from the config ({"pattern": X, "replacement": Y})
- [X] New default config, the TextPreparationPipeline should create language-inferred stemmers and tokenizers if not provided. Handle "rare languages"
- [X] Update and resolve all related tests 
- [x] Add new tests
- [x] Test several Query objects generated before these changes against Query objects generated after

* Will be adding documentation in a separate PR.


mindmeld/text_preparation/normalizers.py
```diff
@@ -127,7 +128,7 @@ class NFD(Normalizer):
     """Unicode NFD Normalizer Class."""
```
ritvikshrivastava, line 128: Should we add examples in the docstrings of the normalizer classes to differentiate between them?

mindmeld/text_preparation/normalizers.py
```diff
@@ -195,28 +196,105 @@ def normalize(self, text):
...
+        self.regex_norm_rules = regex_norm_rules
+
+    def normalize(self, text):
+        """Apply Regex substitutions to the given text.
+        Args:
+            text (str): Input text.
+        Returns:
+            normalized_text (str): Normalized Text.
+        """
+        # TODO: Handle case if the "pattern" or "replacement" is none
```
ritvikshrivastava, line 238: For the to-do:
Maybe consider falling back to the ``MindMeldRegexNormalizer`` and throwing a warning?

Or language subset specific defaults (as discussed).

mindmeld/text_preparation/preprocessor.py
```diff
@@ -33,18 +33,35 @@ def process(self, text):
...
 
+class PreprocessorFactory:
+    """Preprocessor Factory Class"""
+
+    @staticmethod
+    def get_preprocessor(preprocessor):
+        """A static method to get a Preprocessor
+
+        Args:
+            perprocessor (str): Name of the desired Preprocessor class
```
ritvikshrivastava, line 61:
```suggestion
            preprocessor (str): Name of the desired Preprocessor class
```

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+        if not self.normalizers:
+            logger.warning(
+                "This instance of TextPreparationPipeline does not have normalizers."
+                "Returning raw tokens."
+            )
+            return raw_tokens
+
+        normalized_tokens = []
+        for i, raw_token in enumerate(raw_tokens):
+            if not raw_token["text"] or len(raw_token["text"]) == 0:
```
ritvikshrivastava, line 147: Should cover both cases.
```suggestion
            if not raw_token["text"]:
```

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+                    normalized_tokens.append(norm_token)
+        return normalized_tokens
+
+    def _normalize_single_token(self, text):
+        """Normalize an individual token by processing text with all normalizers.
+
+        Args:
+            text (str): Text to normalize.
+        Returns:
+            normalized_text (str): Text to normalize.
```
ritvikshrivastava, line 166:
```suggestion
            normalized_text (str): Normalized text.
```

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+                "This instance of TextPreparationPipeline does not have normalizers."
+                "Returning raw tokens."
+            )
+            return raw_tokens
+
+        normalized_tokens = []
+        for i, raw_token in enumerate(raw_tokens):
+            if not raw_token["text"] or len(raw_token["text"]) == 0:
+                continue
+            normalized_text = self._normalize_single_token(raw_token["text"])
```
ritvikshrivastava, line 149: Maybe consider getting ``normalized_tokens`` as an output of this function directly instead of text? I.e. ``return normalized_text.split()`` from the function?

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+            TextPreparationPipeline.offset_token_start_values(
+                tokens=tokens_after_last_entity, offset=prev_entity_end
+            )
+            tokens.extend(tokens_after_last_entity)
+
+        tokens = TextPreparationPipeline.filter_out_space_text_tokens(tokens)
+        return tokens
+
+    @staticmethod
+    def offset_token_start_values(tokens: List[Dict], offset: int):
```
ritvikshrivastava, line 312: Shouldn't this function be returning the tokens? Doesn't seem to be modifying it in-place?

ritvikshrivastava (commented):
Took a pass, the overall design looks good! Some questions/thoughts/nits! 😄

mindmeld/text_preparation/normalizers.py
```diff
@@ -195,28 +196,105 @@ def normalize(self, text):
...
+                {
+                    "description": "remove_text_in_parantheses",
+                    "pattern": "\(.+?\)",
+                    "replacement": ""
+                }
+        """
+        super().__init__()
+        self.regex_norm_rules = regex_norm_rules
+
+    def normalize(self, text):
```
tmehlinger, line 231: If you take the approach I described above, this method becomes a fair bit simpler:
```python
def normalize(self, text):
    return functools.reduce(lambda fn: fn(text), self.regex_norm_rules)
```

mindmeld/text_preparation/normalizers.py
```diff
@@ -195,28 +196,105 @@ def normalize(self, text):
...
+        # TODO: Handle case if the "pattern" or "replacement" is none
+
+        for regex_norm_rule in self.regex_norm_rules:
+            pattern = regex_norm_rule.get("pattern")
+            replacement = regex_norm_rule.get("replacement")
+            text = re.sub(pattern, replacement, text)
+        return text
+
+
+class MindMeldRegexNormalizer(RegexNormalizer):
```
tmehlinger, line 247: What is this class meant to do that `RegexNormalizer` doesn't already. I feel like we don't need this.

mindmeld/text_preparation/normalizers.py
```diff
@@ -195,28 +196,105 @@ def normalize(self, text):
...
-        elif normalizer == NFC.__name__:
-            return NFC()
-        elif normalizer == NFD.__name__:
-            return NFD()
-        elif normalizer == NFKC.__name__:
-            return NFKC()
-        elif normalizer == NFKD.__name__:
-            return NFKD()
-        raise AssertionError(f" {normalizer} is not a valid Normalizer.")
+        normalizer_classes = [
```
tmehlinger, line 285: This is a perfect use case for maps! If you use a callable for the value, you don't need the special case for `RegexNormalizer`.

```python
normalizer_classes = {
    NoOpNormalizer.__name__: NoOpNormalzer,
    ...,
    RegexNormalizer.__name__: lambda: RegexNormalzer(regex_norm_rules),
}
cls = normalizer_classes.get(normalizer)
if not cls:
    raise TypeError('invalid normalizer type')
return cls()
```

mindmeld/text_preparation/stemmers.py
```diff
@@ -74,46 +85,57 @@ def stem_word(self, word):
...
+        Args:
+            translator (str): Name of the desired translator class
+        Returns:
+            (Translator): Translator Class
+        """
+        if stemmer == EnglishNLTKStemmer.__name__:
+            return EnglishNLTKStemmer()
+        if stemmer == SnowballNLTKStemmer.__name__:
+            return SnowballNLTKStemmer()
+        raise AssertionError(f" {stemmer} is not a valid 'stemmer'.")
```
tmehlinger, line 104: This is probably better conveyed as a `TypeError`.

tmehlinger (changes requested):
I have little domain knowledge so I've looked for places where we can tighten up the code.

One thing I would like you to do (and this won't be any fun, sorry ☹️) is back out a lot of the code style changes. It looks like you ran Black on this with different line length options from an earlier application of Black and it's produce a lot of superfluous style changes. It makes it very difficult to tell what is actually modified code and what is functionally identical.

mindmeld/constants.py
```diff
@@ -325,5 +330,68 @@
...
+        "pattern": "(?<=[^\\s])'[sS]",
+        "replacement": " 's",
+    },
+    {
+        "description": "Remove Punction at Start of Word Skip Exclude From Norm",
+        "pattern": "^[^\w\d&" + CURRENCY_SYMBOLS + EXCEPTION_CHARS + "]+",
+        "replacement": " ",
+    },
+    {
+        "description": "Remove Punction at End of Word Skip Exclude From Norm",
```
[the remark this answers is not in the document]
KKsharma99: Done

mindmeld/constants.py
```diff
@@ -325,5 +330,68 @@
...
+        "replacement": " ",
+    },
+    {"description": "Remove Underscores", "pattern": "_", "replacement": " "},
+    {
+        "description": "Separate s to a separate word",
+        "pattern": "(?<=[^\\s])'[sS]",
+        "replacement": " 's",
+    },
+    {
+        "description": "Remove Punction at Start of Word Skip Exclude From Norm",
```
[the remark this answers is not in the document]
KKsharma99: Done

mindmeld/models/model.py
```diff
@@ -852,15 +852,20 @@ def _get_model_constructor(self):
...
+                "The text_preparation_pipeline resource has not been registered "
+                "to the model. Using default text_preparation_pipeline."
             )
-            tokenizer = Tokenizer()
-        return tokenizer
+            return (
+                TextPreparationPipelineFactory.create_default_text_preparation_pipeline()
+            )
+
+        return text_preparation_pipeline
```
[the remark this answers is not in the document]
KKsharma99: This does make the code more concise, however, I think it's still helpful to give the user the logging error if a text_preparation_pipeline object hasn't been registered

mindmeld/text_preparation/stemmers.py
KKsharma99: Makes sense, fixed

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Done

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Good point, done

mindmeld/text_preparation/preprocessor.py
KKsharma99: Fixed

mindmeld/constants.py
```diff
@@ -325,5 +330,68 @@
...
 
-
 SYSTEM_ENTITY_PREFIX = "sys_"
+
+# TEXT PREPRATION CONSTANTS
+EXCEPTION_CHARS = "\@\[\]\|\{\}'"
+
+DEFAULT_REGEX_NORM_RULES = [
+    {
+        "description": "Remove apostrophe at the end of possessive form",
```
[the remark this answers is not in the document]
KKsharma99: fixed

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Double checked the function in a notebook, it does modify dictionary values in-place

mindmeld/text_preparation/normalizers.py
KKsharma99: Thanks for the advice! Went ahead and applied this strategy for all components of the text_preparation_pipeline

mindmeld/components/_config.py
```diff
@@ -522,6 +514,12 @@
     },
 }
 
+# A language-specific stemmer and tokenizer will be selected during runtime.
```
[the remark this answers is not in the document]
vijay120: Related to rivtik's point, but I would like to see the defaults for english here

mindmeld/components/_config.py
```diff
@@ -1028,135 +1026,46 @@ def get_auto_annotator_config(app_path=None):
...
-    """Gets the active learning configuration for the app at the specified path.
+def get_text_preparation_config(app_path=None):
+    """Gets the text preparation configuration for the app at the specified path.
 
     Args:
         app_path (str, optional): The location of the MindMeld app
 
     Returns:
-        dict: The active learning configuration.
+        dict: The tokenizer configuration.
```
vijay120, line 1060: its unclear why the function states `get_text_preparation_config ` but we give back a tokenizer? I thought the tokenizer was part of the text prep?

vijay120: I think we are returning a text prep pipeline here, if thats the case, lets not mention tokenizer.

mindmeld/components/nlp.py
```diff
@@ -1519,8 +1519,9 @@ def _find_entities_in_text(self, query, dynamic_resource,
         # We extract the `entity` key from each token since that represents the normalized text of
         # each token which we will compare against the normalized text of the query
         def tokenize_to_tuples(text):
```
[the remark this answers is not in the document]
vijay120: isn't it `tokenizer_and_normalize_to_tuples`

mindmeld/models/embedder_models.py
```diff
@@ -757,10 +758,14 @@ class GloveEmbedder(Embedder):
...
     def __init__(self, app_path, **kwargs):
         super().__init__(app_path, **kwargs)
-        self.glove_tokenizer = Tokenizer()
+        self.glove_text_preparation_pipeline = (
+            TextPreparationPipelineFactory.create_default_text_preparation_pipeline()
+        )
 
     def tokenize(self, text):
-        tokens = self.glove_tokenizer.tokenize(text, keep_special_chars=False)
+        tokens = self.glove_text_preparation_pipeline.get_normalized_tokens_from_text(
```
vijay120, line 766: Are we making sure `keep_special_chars ` is being propagated correctly?

mindmeld/query_factory.py
```diff
@@ -91,30 +87,49 @@ def create_query(
...
-        # create raw, processed maps
-        if self.preprocessor:
-            processed_text = self.preprocessor.process(raw_text)
-            maps = self.preprocessor.get_char_index_map(raw_text, processed_text)
-            forward, backward = maps
-            char_maps[(TEXT_FORM_RAW, TEXT_FORM_PROCESSED)] = forward
-            char_maps[(TEXT_FORM_PROCESSED, TEXT_FORM_RAW)] = backward
+        # Step 1: Preprocessing
+        if self.text_preparation_pipeline.preprocessors:
+            processed_text = self.text_preparation_pipeline.preprocess(raw_text)
```
vijay120, line 92: so `self.preprocessor.process` is the same as `self.text_preparation_pipeline.preprocess`?

mindmeld/text_preparation/preprocessors.py
```diff
@@ -33,18 +33,37 @@ def process(self, text):
...
         Returns:
-            (dict, dict): A tuple consisting of two maps, forward and backward
+            (Preprocessor): Preprocessor Class
         """
-        pass
+        preprocessor_classes = {NoOpPreprocessor.__name__: NoOpPreprocessor}
+        preprocessor_class = preprocessor_classes.get(preprocessor)
+        if not preprocessor_class:
+            raise TypeError(f"{preprocessor} is not a valid Preprocessor type.")
+        return preprocessor_class()
```
vijay120, line 69: so this class just provides `NoOpPreprocessor ` objects?

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+        preprocessors: List[Preprocessor] = None,
+        normalizers: List[Normalizer] = None,
+        language: str = ENGLISH_LANGUAGE_CODE,
+    ):
+        """Creates a Pipeline instance."""
+        self.language = language
+        self.preprocessors = preprocessors
+        self.normalizers = normalizers or [NoOpNormalizer()]
+        self.tokenizer = tokenizer
+        self.stemmer = stemmer
```
vijay120, line 66: Do we also want to use a `noopstemmer` here as an or condition? Same with the other noop components

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+            logger.warning(
+                "This instance of TextPreparationPipeline does not have preprocessors."
+                "Returning the original text."
+            )
+            return text
+
+        preprocessed_text = text
+        for preprocessor in self.preprocessors:
+            preprocessed_text = (
+                TextPreparationPipeline.modify_around_mindmeld_annotations(
```
vijay120, line 90: `modify_around_mindmeld_annotations ` -> `modify_around_annotations `

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+            return raw_tokens
+
+        normalized_tokens = []
+        for i, raw_token in enumerate(raw_tokens):
+            if not raw_token["text"]:
+                continue
+            normalized_text = self._normalize_single_token(raw_token["text"])
+            if len(normalized_text) > 0:
+                for token in normalized_text.split():
+                    norm_token = {}
```
vijay120, line 152: Nit: just a dict with the values in it

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+    @staticmethod
+    def find_mindmeld_annotation_re_matches(text):
+        """
+        Args:
+            text (str): The string to find mindmeld annotation instances
+                (" {entity_text|entity_type} ")
+        Returns:
+            matches (List[sre.SRE_Match object]): Regex match objects.
+        """
+        COMPILED_MINDMELD_ANNOTATION_PATTERN = re.compile(MINDMELD_ANNOTATION_PATTERN)
```
vijay120, line 209: I think we should just compile it right after assigning `MINDMELD_ANNOTATION_PATTERN `

mindmeld/text_preparation/preprocessors.py
KKsharma99: Yes, for now MindMeld does not have any preset Preprocessors. I used the same mapping approach as the other 3 components for consistency

mindmeld/query_factory.py
KKsharma99: That's correct. However, `self.text_preparation_pipeline.preprocess` can use multiple preprocessors to initially process the raw query

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Done

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Good point, updated

mindmeld/components/_config.py
KKsharma99: Changed to `dict: The text preparation pipeline configuration.`

mindmeld/components/nlp.py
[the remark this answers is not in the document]
KKsharma99: how about `get_normalized_tokens_as_tuples`?

mindmeld/gazetteer.py
```diff
@@ -216,9 +218,10 @@ def update_with_entity_map(
         if len(self.pop_dict) > 0:
             min_popularity = min(self.pop_dict.values())
         for item in mapping:
-            tokenized_canonical = tuple(
-                token["entity"] for token in self.tokenizer.tokenize(
-                    normalizer(item["cname"])))
+
+            raw_tokens = self.text_preparation_pipeline.tokenize(normalizer(item["cname"]))
```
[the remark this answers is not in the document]
KKsharma99: Looks like the `normalizer` in this function refers to the query_factory's normalize function which uses the text_preparation_pipeline.normalize(text). So I've simplified this function to the following:

```
            normalized_tokens = normalizer(item["cname"]).split()
            tokenized_canonical = tuple(token["entity"] for token in normalized_tokens)
```

mindmeld/text_preparation/normalizers.py
KKsharma99: The examples are better described in images and are a bit too complex to be explained in doc-strings. We can instead direct them to Unicode's official documentation on the Normalization type. Updated the class docstrings to the following:

```
class NFKC(Normalizer):
    """Unicode NFKC Normalizer Class.
    (Compatibility Decomposition, followed by Canonical Composition)

    For more details: https://unicode.org/reports/tr15/#Norm_Forms
    """ 
```

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: I see what you mean, yes, this would be cleaner! To take it a step further, we can better generalize the post-norm tokenization by using the language specific tokenization (self.tokenize) here

An example of post-norm tokenization is “o’clock” -> “o clock” -> [“o”, “clock”]. In English, the .split() works fine since words are separated by space, however, to generalize better for other languages (like Japanese), so it'd make sense to do self.tokenize(norm_text_single_token) instead of .split().

mindmeld/query_factory.py
```diff
@@ -91,30 +87,48 @@ def create_query(
...
+                raw_text, processed_text
+            )
+            char_maps[(TEXT_FORM_RAW, TEXT_FORM_PROCESSED)] = forward_map
+            char_maps[(TEXT_FORM_PROCESSED, TEXT_FORM_RAW)] = backward_map
         else:
             processed_text = raw_text
 
-        normalized_tokens = self.tokenizer.tokenize(processed_text)
+        # Step 2: Tokenization
+        raw_tokens = self.text_preparation_pipeline.tokenize(processed_text)
```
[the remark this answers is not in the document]
KKsharma99: Sure, changed

mindmeld/text_preparation/normalizers.py
KKsharma99: For Regex language-specific defaults, the plan is to split Each Regex rule into its own class (as Travis suggested). Doing so, we would no longer have `MindMeldRegexNormalizer`. This would make it easier to maintain language specific defaults. Here is an example: 

```
{ 
      “normalizers”: [
              “RemoveUnderscores”,
              “SeparateAposSChars”,
              “RemoveSpecialCharsBetweenWords”
      ]
}
```

While it is a bit more verbose, this approach is fully customizable and is what HuggingFace follows as well. As Vijay mentioned, for certain languages such as Arabic we would explicitly require developers to use their own Regex rules. As it is not immediately clear which languages would use which Regex norm rules.

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Makes sense! Now `MINDMELD_ANNOTATION_PATTERN` represents the compiled form

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Sure, added NoOp defaults for the stemmer and preprocessors:

```
self.preprocessors = preprocessors or [NoOpPreprocessor()]
self.normalizers = normalizers or [NoOpNormalizer()]
self.tokenizer = tokenizer
self.stemmer = stemmer or NoOpStemmer()
```

mindmeld/components/_config.py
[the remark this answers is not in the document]
KKsharma99: Below is what the default EN Config would look like based on changes in progress. I'm thinking about the best way to handle these defaults and it may make sense to have language specific normalizer lists in the file `EN_NORMALIZERS` and keep the other elements "dynamic" such that they are inferred based on the language. Ie. For tokenizer, it will select a SpacyTokenizer if the language is supported by Spacy, else, WhiteSpaceTokenier. Similar procedure with Stemmer. Thoughts?

```
DEFAULT_EN_TEXT_PREPARATION_CONFIG = {
    "preprocessors": [],
    "tokenizer": "SpacyTokenizer",
    "normalizers": [
        'RemoveAposAtEndOfPossesiveForm',
        'RemoveAdjacentAposAndSpace',
        'RemoveBeginningSpace',
        'RemoveTrailingSpace',
        'ReplaceSpacesWithSpace',
        'ReplaceUnderscoreWithSpace',
        'SeparateAposS',
        'ReplacePunctuationAtWordStartWithSpace',
        'ReplacePunctuationAtWordEndWithSpace',
        'ReplaceSpecialCharsBetweenLettersAndDigitsWithSpace',
        'ReplaceSpecialCharsBetweenDigitsAndLettersWithSpace',
        'ReplaceSpecialCharsBetweenLettersWithSpace',
        'RemoveSpecialCharsBeforePipe',
        'ReplaceEndBracketAndFollowingSpecialCharsBeforeSWithSpace'
        'Lowercase',
        'ASCIIFold'
    ],
    "stemmer": "EnglishNLTKStemmer"
}
```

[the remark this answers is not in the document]
KKsharma99: If we place English defaults here then the correct (dynamically created) Tokenizers and Stemmers won't be expected. Since a dynamic Tokenizer/Stemmer is only selected if the given attribute is None

mindmeld/text_preparation/normalizers.py
KKsharma99: Makes sense, removed. The rules in `MindMeldRegexNormalizer` are now the default so we no longer need the class to specify them

mindmeld/constants.py
```diff
@@ -325,5 +330,68 @@
...
+    {"description": "Remove Trailing Spaces", "pattern": "\s+$", "replacement": ""},
+    {
+        "description": "Replace Multiple Spaces with Single Space",
+        "pattern": "\s+",
+        "replacement": " ",
+    },
+    {"description": "Remove Underscores", "pattern": "_", "replacement": " "},
+    {
+        "description": "Separate s to a separate word",
+        "pattern": "(?<=[^\\s])'[sS]",
```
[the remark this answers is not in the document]
KKsharma99: Done

mindmeld/constants.py
```diff
@@ -325,5 +330,68 @@
 UNICODE_NON_LATIN_CATEGORY = "Lo"
 UNICODE_SPACE_CATEGORY = "Zs"
 
-
 SYSTEM_ENTITY_PREFIX = "sys_"
+
+# TEXT PREPRATION CONSTANTS
+EXCEPTION_CHARS = "\@\[\]\|\{\}'"
```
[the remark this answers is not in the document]
KKsharma99: Fixed

mindmeld/constants.py
```diff
@@ -325,5 +330,68 @@
...
+
+DEFAULT_REGEX_NORM_RULES = [
+    {
+        "description": "Remove apostrophe at the end of possessive form",
+        "pattern": "^'(?=\S)|(?<=\S)'$",
```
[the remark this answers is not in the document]
KKsharma99: Good point, switched to raw strings

mindmeld/constants.py
```diff
@@ -325,5 +330,68 @@
 UNICODE_NON_LATIN_CATEGORY = "Lo"
 UNICODE_SPACE_CATEGORY = "Zs"
 
-
 SYSTEM_ENTITY_PREFIX = "sys_"
+
+# TEXT PREPRATION CONSTANTS
+EXCEPTION_CHARS = "\@\[\]\|\{\}'"
+
+DEFAULT_REGEX_NORM_RULES = [
```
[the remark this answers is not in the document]
KKsharma99: Thank you for this suggestion Travis! Went ahead and made each Regex rule its own class following this suggestion. Now compiling regex patterns and have moved the rules to the normalizers.py module as suggested.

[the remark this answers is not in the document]
KKsharma99: Switched back to the following to avoid inline comments to ignore style errors from pylint and my IDE.

```
class RegexNormalizerRule(Normalizer):

    def __init__(self, pattern: str, replacement: str):
        """Creates a RegexNormalizerRule instance."""
        self.pattern = pattern
        self.replacement = replacement
        self._expr = re.compile(self.pattern)

    def normalize(self, s):
        return self._expr.sub(self.replacement, s)
```

Here is an example of this being used:

`"SeparateAposS": RegexNormalizerRule(pattern=r"(?<=[^\s])'[sS]", replacement=" 's"),
`

mindmeld/models/embedder_models.py
KKsharma99: Good point, will take a closer look to double check

mindmeld/components/_config.py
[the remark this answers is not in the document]
ritvikshrivastava: That does make sense if we can dynamically select the components.

Although, is it not possible to overwrite the default config here with the dynamic values? Not sure how that affects runtime but one idea could be to load the config from file or default to this, and then create a runtime config based on language selected. If nothing is obtained from the latter, then continue as is or otherwise overwrite the default.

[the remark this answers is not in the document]
ritvikshrivastava: One of the reasons I mentioned default EN config was that most of the other functionalities in MM currently default to that. So this would be consistent, but depending upon consensus we can go the dynamic route as well.

mindmeld/components/_config.py
```diff
@@ -522,6 +514,35 @@
...
+    'ReplacePunctuationAtWordStartWithSpace',
+    'ReplacePunctuationAtWordEndWithSpace',
+    'ReplaceSpecialCharsBetweenLettersAndDigitsWithSpace',
+    'ReplaceSpecialCharsBetweenDigitsAndLettersWithSpace',
+    'ReplaceSpecialCharsBetweenLettersWithSpace',
+    'RemoveSpecialCharsBeforePipe',
+    'ReplaceEndBracketAndFollowingSpecialCharsBeforeSWithSpace',
+    'Lowercase',
+    'ASCIIFold'
+]
```
ritvikshrivastava, line 536: Maybe move these to a ``config_helpers`` file? We can do a refactor of the config file to move a lot of the supporting code later. This file has become quite verbose.

mindmeld/components/_config.py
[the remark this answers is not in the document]
KKsharma99: I think we should avoid having configs overwritten internally as that creates confusion to the user. To your point regarding our use of a default EN config elsewhere in the code: We can make a default en config and then have a check that goes (if language ==“en” return default_en_config). Such logic would be a bit redundant, however, it could be more consistent?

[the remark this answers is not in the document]
KKsharma99: After discussing this with Vijay, we decided to use the existing Normalization rules for English for all languages. If a user would like to update the rule set, they can paste a subset of the default rules into the 'normalizer' section of the config. These changes have been made

mindmeld/components/_config.py
KKsharma99: Good point, however, after discussing this with Vijay, we decided to use the existing Normalization rules for English for all languages. So we are no longer maintaining defaults for each language

mindmeld/models/embedder_models.py
KKsharma99: Many of the regex rules remove "Special Characters" that can be found in between words or numbers. Here is an example rule:

```
    "ReplacePunctuationAtWordEndWithSpace": RegexNormalizerRule(
        pattern=r"[^\w\d&" + CURRENCY_SYMBOLS + EXCEPTION_CHARS + r"]+$",
        replacement=" ",
    )
```

The EXCEPTION_CHARS currently refers to `r"\@\[\]\|\{\}'"`. When keep_special_chars (old code) is used, these EXCEPTION_CHARS are applied such that these special chars are not removed.

I think we may be able to skip the "{" "}" characters since regex detects MindMeld annotation patterns now. Wonder if we could skip "[" and "]" as well? In which case are brackets used.

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+            backward_map (Dict): Reverse mapping from modified text to raw text.
+        """
+        preprocessed_text = text
+        for preprocessor in self.preprocessors:
+            preprocessed_text = TextPreparationPipeline.modify_around_annotations(
+                text=preprocessed_text, function=preprocessor.process
+            )
+        return preprocessed_text
+
+    def normalize(self, text):
```
vijay120, line 82: I think this function should be named `tokenize_and_normalize`

vijay120: Here is how I would separate out the functions:
1. Tokenize: str -> list[str]
2. Normalize: str -> str
3. TokenizeAndNormalize: str -> list[str]
The callers of these primitive functions can then do things like `` `.join`

mindmeld/components/nlp.py
```diff
@@ -1518,9 +1518,10 @@ def _find_entities_in_text(self, query, dynamic_resource,
 
         # We extract the `entity` key from each token since that represents the normalized text of
         # each token which we will compare against the normalized text of the query
-        def tokenize_to_tuples(text):
-            return tuple([token["entity"] for token in
-                          self.resource_loader.query_factory.tokenizer.tokenize(text)])
+        def get_normalized_tokens_as_tuples(text):
+            return tuple(
+                self.resource_loader.query_factory.text_preparation_pipeline.normalize(text).split()
```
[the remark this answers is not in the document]
vijay120: I think `get_normalized_tokens_as_tuples` can be a function in text_preparation_pipeline

mindmeld/constants.py
```diff
@@ -321,9 +321,7 @@
 CURRENCY_SYMBOLS = u"".join(
     chr(i) for i in range(0xFFFF) if unicodedata.category(chr(i)) == "Sc"
 )
-ASCII_CUTOFF = ord("\u0080")
```
vijay120, line 324: why do we remove this?

mindmeld/gazetteer.py
```diff
@@ -216,9 +219,8 @@ def update_with_entity_map(
         if len(self.pop_dict) > 0:
             min_popularity = min(self.pop_dict.values())
         for item in mapping:
-            tokenized_canonical = tuple(
-                token["entity"] for token in self.tokenizer.tokenize(
-                    normalizer(item["cname"])))
+
+            tokenized_canonical = tuple(normalizer(item["cname"]).split())
```
vijay120, line 223: Why do we not do the same thing as line 130 here?

mindmeld/models/embedder_models.py
vijay120: I don't think we skip `[`/`]` with the current changes in the PR, no?

mindmeld/constants.py
KKsharma99: Moved it to line 26 of `text_preparation_pipeline/normalizers.py` since it is solely being used by this file. Should I move it to `constants.py` instead?

mindmeld/text_preparation/preprocessors.py
```diff
@@ -33,18 +33,37 @@ def process(self, text):
...
+        """A static method to get a Preprocessor
+
+        Args:
+            preprocessor (str): Name of the desired Preprocessor class
         Returns:
-            (dict, dict): A tuple consisting of two maps, forward and backward
+            (Preprocessor): Preprocessor Class
         """
-        pass
+        preprocessor_classes = {NoOpPreprocessor.__name__: NoOpPreprocessor}
```
vijay120, line 65: I think these dicts can be added to a contants.py file

mindmeld/text_preparation/preprocessors.py
```diff
@@ -33,18 +33,37 @@ def process(self, text):
...
+        preprocessor_class = preprocessor_classes.get(preprocessor)
```
vijay120, line 66: The logic here can be simplified since we are checking if the instance is the same as `NoOpPreprocessor `. Finally, why are we only giving the option to use a NoOpPreprocessor instead of the other ones?

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+                        {
+                            "entity": token_text,
+                            "raw_entity": raw_token["text"],
+                            "raw_token_index": i,
+                            "raw_start": raw_token["start"],
+                        }
+                    )
+        return normalized_tokens
+
+    def _normalize_single_token(self, text):
```
vijay120, line 141: maybe `_normalize_text`, also add the arg types here.

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+                split the entity if possible.
+        """
+        normalized_text = text
+        for normalizer in self.normalizers:
+            normalized_text = TextPreparationPipeline.modify_around_annotations(
+                text=normalized_text, function=normalizer.normalize
+            )
+        # We tokenize the post-norm text and split the entity if possible
+        # Ex: normalize("o'clock") -> "o clock" -> ["o", "clock"]
+        return [t["text"] for t in self.tokenize(normalized_text)]
```
vijay120, line 157: why do we do this? It seems like it's the caller's job to do tokenization. In general, I would like to keep these functions simple. So `"o'clock"` -> `o clock` here and then the caller can do whatever with it.

mindmeld/text_preparation/text_preparation_pipeline.py
vijay120, line 141: Maybe `_normalize_text `

mindmeld/text_preparation/preprocessors.py
```diff
@@ -33,18 +33,37 @@ def process(self, text):
...
+        if not preprocessor_class:
```
vijay120, line 67: If ANLP defines a custom preprocessor, how do we enable that here?

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+        """
+        matches = TextPreparationPipeline.find_mindmeld_annotation_re_matches(text)
+
+        modified_text = []
+        prev_entity_end = 0
+
+        for match in matches:
+            entity_start, entity_end = match.span()
+
+            # Adds "function(pre_entity_text)" {..
```
vijay120, line 214: Shouldnt this be `# Adds "function(pre_entity_text)"{..`

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+            text_before_entity = text[prev_entity_end:entity_start]
+            modified_text.append(function(text_before_entity))
```
vijay120, line 216: We have to be cautious that no preprocessors eliminate trailing spaces since that would fuse `function(pre_entity_text)` and `{` for latin languages

vijay120: One way we can do this is by checking if the previous chars are spaces. If they are, then modify around text before the space chars.

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+            prev_entity_end = entity_end
+
+        if prev_entity_end < len(text):
+            # Adds the remainder of the text after the last end brace } "function(post_entity_text)"
+            modified_text.append(function(text[prev_entity_end : len(text)]))
+
+        return "".join(modified_text)
+
+    @staticmethod
+    def tokenize_around_mindmeld_annotations(text, function):
```
vijay120, line 241: Why do we have to configure the tokenize function? Seems like we can just use self.tokenizer, no?

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+        Args:
+            text (str): The string to find mindmeld annotation instances
+                (" {entity_text|entity_type} ")
+        Returns:
+            matches (List[sre.SRE_Match object]): Regex match objects.
+        """
+        return list(MINDMELD_ANNOTATION_PATTERN.finditer(text))
+
+    @staticmethod
+    def modify_around_annotations(text, function):
```
vijay120, line 194: I think `modify_around_annotations ` and `tokenize_around_mindmeld_annotations ` can be combined. The main difference is the append vs extend and offset ops. The offset ops can be composed into the function that is passed in.

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,530 @@
...
+        """
+        language, _ = get_language_config(app_path)
+        text_preparation_config = get_text_preparation_config(app_path)
+
+        if (
+            text_preparation_config.get("regex_norm_rules")
+            and "normalizers" not in text_preparation_config
+        ):
+            logger.warning(
+                "Detected 'regex_norm_rules' in TEXT_PREPARATION_CONFIG, however, 'normalizers' "
```
vijay120, line 440: great!

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Yeah, good point, I'll make that change

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: `modify_around_annotations` is also used by normalizers so it may be better to keep them separate

mindmeld/text_preparation/preprocessors.py
KKsharma99: Sure, I can simplify this to a single if statement. Wrote it this way in anticipation of additional Preprocessors but not sure what functionality to put in a preprocessor that wouldn't simply be a normalization class. Simplified logic to:

```
        if preprocessor == NoOpPreprocessor.__name__:
            return NoOpPreprocessor()
        raise TypeError(f"{preprocessor} is not a valid Preprocessor type.")
```

mindmeld/text_preparation/preprocessors.py
KKsharma99: For preprocessors.py, I changed this logic to a single if statement since there is only one preprocessor.

```
if preprocessor == NoOpPreprocessor.__name__:
    return NoOpPreprocessor()
raise TypeError(f"{preprocessor} is not a valid Preprocessor type.")
```

Would you like for us to add the class mapping dicts in other component files into constants.py?

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Sounds good, done

mindmeld/text_preparation/preprocessors.py
KKsharma99: To add custom preprocessors without modifying the MindMeld codebase, a developer can make a custom TextPreparationPipeline and pass that in at the top level (Similar to how preprocessors were being passed in before.) I am working on a section in the docs that showcases this functionality.

The other alternative is to modify MindMeld itself and add the classes. I am open to other ideas on this customization as well

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: As discussed:

Tokenize: str -> list[str]
Normalize: str -> str

TokenizeAndNormalize: str -> list[Dict]

mindmeld/components/nlp.py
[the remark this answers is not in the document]
KKsharma99: Done

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Good catch, fixed

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Added this check. Checks if there is a space character before the start to annotation, if so it decrements entity_start by one.

```
  # Include Space Before Entity Annotation
  if (
      keep_space_before_annotation
      and entity_start > 0
      and text[entity_start - 1].isspace()
  ):
      entity_start -= 1
```

mindmeld/models/embedder_models.py
KKsharma99: Added support for this from the config as discussed.

tests/conftest.py
```diff
@@ -16,16 +16,18 @@
 
 import pytest
 
-from mindmeld.components import NaturalLanguageProcessor, Preprocessor, QuestionAnswerer
+from mindmeld.converter.rasa import RasaConverter
+from mindmeld.converter.dialogflow import DialogflowConverter
+from mindmeld.components import NaturalLanguageProcessor, QuestionAnswerer
```
ritvikshrivastava, line 21: small nit, but can you alphabetically arrange the imports across files?

ritvikshrivastava (approved):
A couple of small nits, but otherwise LGTM once other reviewers' comments have been addressed 😄

mindmeld/text_preparation/stemmers.py
```diff
@@ -74,46 +85,61 @@ def stem_word(self, word):
...
-        "config.py, consider disabling it.",
-        language_code,
-    )
-    return NoOpStemmer()
+    @staticmethod
+    def get_language_from_language_code(language_code):
+        language = None
+        if len(language_code) == 2:
+            language = pycountry.languages.get(alpha_2=language_code)
+        elif len(language_code) == 3:
```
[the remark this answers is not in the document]
KKsharma99: Yeah, these are separate as there are both 2 digit language codes and some 3 digit language codes (non overlapping)

mindmeld/gazetteer.py
```diff
@@ -125,7 +126,9 @@ def _update_entity(self, entity, popularity, keep_max=True):
         """
         # Only update the relevant data structures when the entity isn't
         # already in the gazetteer. Update the popularity either way.
-        tokenized_gaz_entry = tuple(token["entity"] for token in self.tokenizer.tokenize(entity))
+        tokenized_gaz_entry = tuple(
+            token["text"] for token in self.text_preparation_pipeline.tokenize(entity)
```
vijay120, line 130: why text instead of entity?

mindmeld/gazetteer.py
vijay120: Does the normalizer function normalize and tokenize?

mindmeld/text_preparation/normalizers.py
```diff
@@ -195,28 +212,150 @@ def normalize(self, text):
...
+        self.replacement = replacement
+        self._expr = re.compile(self.pattern)
+
+    def normalize(self, s):
+        return self._expr.sub(self.replacement, s)
+
+
+class RegexNormalizerRuleFactory:
+
+    exception_chars = r"\@\[\]'"
```
vijay120, line 241: Could you clarify why we need this as a class var? Why not just do `def __init__` and place `exception_chars ` as an instance var?

mindmeld/text_preparation/normalizers.py
```diff
@@ -195,28 +212,150 @@ def normalize(self, text):
...
+        Args:
+            regex_normalizer (str): Name of the desired RegexNormalizerRule
+        Returns:
+            (RegexNormalizerRule): Default Regex Normalizer Rule
+        """
+        if regex_normalizer in DEFAULT_REGEX_NORM_RULES:
+            regex_rule_dict = DEFAULT_REGEX_NORM_RULES[regex_normalizer]
+            # Inserts current exception_chars in pattern string if applicable
+            regex_rule_dict["pattern"] = regex_rule_dict["pattern"].format(
+                exception_chars=RegexNormalizerRuleFactory.exception_chars
```
vijay120, line 273: great!

mindmeld/text_preparation/tokenizers.py
```diff
@@ -101,6 +111,10 @@ def tokenize(self, text):
                 Keys include "start" (token starting index), and "text" (token text).
                 For example: [{"start": 0, "text":"hello"}]
         """
+        if text == "":
+            return []
+        if not text:
```
vijay120, line 116: why is None ever passed in?

KKsharma99: I don't think it is, the check is there just in case it is, should I remove it?

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,566 @@
...
+        Args:
+            text (str): Text to normalize.
+            keep_special_chars (bool): Whether to prevent special characters (such as @, [, ])
+                from being removed in the normalization process. No longer supported at the
+                function level, can be specified in the config.
+        Returns:
+            normalized_text (str): Normalized text.
+        """
+        if keep_special_chars:
+            logger.warning(
```
vijay120, line 100: great!

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,566 @@
...
+                "'keep_special_chars' is depreciated as a parameter to normalize(). "
```
vijay120, line 101: deprecated

mindmeld/text_preparation/normalizers.py
KKsharma99: I went with a class variable since the Factory only has static methods. Moreover, by updating the class variable the change persists throughout the whole system. However, since Normalizers are only made once anyways, I can switch this to instantiate the Factory instead.

mindmeld/gazetteer.py
KKsharma99: Good point, I'll switch this to "entity" and "tokenize_and_normalize" as that is more consistent with the old "tokenize"

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,566 @@
...
+    def modify_around_annotations(text, function, keep_space_before_annotation=False):
+        """Applied a function around the mindmeld annotation.
+
+        function(pre_entity_text) + { + function(entity_text) + |entity_name}
+            + function(post_entity_text)
+
+        Args:
+            text (str): Original sentence with markup to modify.
+            function (function): Function to apply around the annotation
+            keep_space_before_annotation (bool): If True and if there is a space before entity
```
vijay120, line 222: Unclear why we have this. The logic should be that if the previous char is a space ie "abc {xyz}", then the space should be preserved, so it's implicit.

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,566 @@
...
+        matches = TextPreparationPipeline.find_mindmeld_annotation_re_matches(text)
+
+        modified_text = []
+        prev_entity_end = 0
+
+        for match in matches:
+            entity_start, entity_end = match.span()
+
+            # Include Space Before Entity Annotation
+            if (
```
vijay120, line 236: What happens when there are multiple spaces like "xyz      {abc}". I think in this case, we need to preserve all of the spaces.

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -0,0 +1,566 @@
...
+                dictionaries. ({"pattern":<pattern>, "replacement":<replacement>})
+            normalizers (List[str]): List of normalizer class names.
+            tokenizer (str): Class name of Tokenizer to use.
+            stemmer (str): Class name of Stemmer to use.
+
+        Returns:
+            TextPreparationPipeline: A TextPreparationPipeline class.
+        """
+        preprocessors = (
+            [PreprocessorFactory.get_preprocessor(p) for p in preprocessors]
```
vijay120, line 521: I am not sure how a custom preprocessor will be used here since we only check for noop preprocessor in PreprocessorFactory

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Fixed

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: What is the motivation for preserving multiple spaces? From my understanding, a single space was preserved so that MindMeld entity annotations can be detected downstream in the application. If that is the reason, we don't necessarily need multiple spaces?

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: So this is how it was being done previously

```
from mindmeld import Application
from .stem_processor import StemProcessor

preprocessor = StemProcessor()
app = Application(__name__, preprocessor=preprocessor)
```
https://www.mindmeld.com/docs/userguide/preprocessor.html

Instead of this you would place a 'text_preparation_pipeline' object. Which you would and then you can overwrite the preprocessor section with custom preprocessor objects.

KKsharma99: text_preparation_pipeline.preprocessors = [CustomPreprocessor()]
app = Application(__name__, text_preparation_pipeline=text_preparation_pipeline)

mindmeld/text_preparation/normalizers.py
KKsharma99: Added a comment here

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Preprocessors now process the input text directly

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Preprocessors now process the input text directly

tests/text_preparation/test_normalizers.py
```diff
@@ -135,3 +143,352 @@ def test_nfkc_normalization(nfkc_normalizer, sentence, expected_normalized_text)
...
+
+    regex_normalizer = RegexNormalizerRuleFactory.get_default_regex_normalizer_rule(
+        regex_norm_rule
+    )
+    text_preparation_pipeline.normalizers = [regex_normalizer]
+    normalized_text = text_preparation_pipeline.normalize(input_text)
+    assert normalized_text == expected_text
+
+
+@pytest.mark.parametrize(
```
vijay120, line 172: Could you also do some tests against japanese for the preprocessors and normalizers?

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Modified TextPreparationPipelineFactory.create_text_preparation_pipeline() to be able to accept either List[str] or List[Preprocessor] for each of the components as per our discussion

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -548,3 +565,49 @@ def create_default_text_preparation_pipeline():
...
+    @staticmethod
+    def _instantiate_pipeline_components(  # pylint: disable=W0640
+        expected_component_class, components, language=None
+    ):
+        """Helper method to instantiate the components of a TextPreparationPipeline by iterating
+        through a tuple of either Strings or Component Objects. If the value is a string, the
+        related component_factory_getter will be used to instantiate the required component.
+        Args:
+
+        expected_component_class (Class): The expected type of the component.
```
vijay120, line 578: These docs can be updated

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -548,3 +565,49 @@ def create_default_text_preparation_pipeline():
...
+                    component_factory_getter.get(expected_component_class.__name__)()
+                )
+            elif isinstance(component, expected_component_class):
+                instantiated_components.append(component)
+            else:
+                raise ValueError(
+                    f"{component} must be of type String or {expected_component_class.__name__}."
+                )
+
+        if len(instantiated_components) == 1:
```
vijay120, line 611: why dont we just return list?

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -480,66 +480,83 @@ def create_from_app_path(app_path):
...
+                Preprocessor, preprocessors, language
+            )
             if preprocessors
             else [NoOpPreprocessor()]
         )
 
+        # Update Regex Normalization Exception Characters as Specified in the Config
         if keep_special_chars:
-            RegexNormalizerRuleFactory.EXCEPTION_CHARS = keep_special_chars
+            RegexNormalizerRuleFactory.set_exception_chars(keep_special_chars)
```
vijay120, line 517: Since this is a class var, I dont think we need to have getters and setters. My previous suggestion was if we made it into an instance var.

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -480,66 +480,83 @@ def create_from_app_path(app_path):
...
 
-        normalizers = (
-            [NormalizerFactory.get_normalizer(n) for n in normalizers]
+        # Instantiate Normalizers
+        instantiated_normalizers = (
+            TextPreparationPipelineFactory._instantiate_pipeline_components(
+                Normalizer, normalizers, language
```
vijay120, line 522: I dont think language has to be pass here

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -480,66 +480,83 @@ def create_from_app_path(app_path):
...
             if tokenizer
             else TokenizerFactory.get_tokenizer_by_language(language)
         )
-        stemmer = (
-            StemmerFactory.get_stemmer(stemmer)
+
+        # Instantiate Stemmer
+        instantiated_stemmer = (
+            TextPreparationPipelineFactory._instantiate_pipeline_components(
+                Stemmer, stemmer, language
```
vijay120, line 548: language not needed

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -548,3 +565,49 @@ def create_default_text_preparation_pipeline():
         return TextPreparationPipelineFactory.create_text_preparation_pipeline(
             **DEFAULT_EN_TEXT_PREPARATION_CONFIG
         )
+
+    @staticmethod
+    def _instantiate_pipeline_components(  # pylint: disable=W0640
```
vijay120, line 570: maybe construct_pipeline_components ?

vijay120, line 570: Can we test this function ie pass in varied type classes and make sure it works as expected?

vijay120 (commented):
A few comments

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Ah, looks like I missed this! Updated.

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Returning a list instead

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Got it, removed getters and setters

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Removed the language param input for preprocessors, normalizers, and stemmers

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Removed the language param input for preprocessors, normalizers, and stemmers

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Renamed

KKsharma99: Sure, added tests for valid and invalid input

tests/text_preparation/test_text_preparation_pipeline.py
```diff
@@ -137,3 +137,38 @@ def test_construct_pipeline_components():
...
+        TextPreparationPipelineFactory.create_text_preparation_pipeline(
+            preprocessors = None,
+            normalizers = None,
+            tokenizer = None,
+            stemmer = "NoOpPreprocessor"
+        )
+
+    with pytest.raises(TypeError):
+        TextPreparationPipelineFactory.create_text_preparation_pipeline(
+            preprocessors = ("NoOpNormalizer"),
```
vijay120, line 170: Could you also test passing in a mix, ie string and component type?

KKsharma99: Sure, so as in the example in test_construct_pipeline_components_valid_input() but with string/object components that are of an invalid type?

KKsharma99: Added

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -582,32 +582,48 @@ def _construct_pipeline_components(  # pylint: disable=W0640
...
-            else:
-                raise ValueError(
-                    f"{component} must be of type String or {expected_component_class.__name__}."
-                )
-
+            )
         return instantiated_components
+
+    @staticmethod
+    def _construct_single_pipeline_component(  # pylint: disable=W0640
```
vijay120, line 595: name this `_construct_pipeline_component`

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -567,10 +567,10 @@ def create_default_text_preparation_pipeline():
         )
 
     @staticmethod
-    def _construct_pipeline_components(  # pylint: disable=W0640
+    def _construct_multiple_pipeline_components(  # pylint: disable=W0640
```
vijay120, line 570: `_construct_pipeline_components`

vijay120 (approved):

PR #330, opened 2021-06-25
Update to new Duckling version

serapio:
Replaces duckling path and checksums to point to new duckling release. This drops support for centos 8 and adds support for ubuntu 20.


serapio:
@alhuang10 the "Feb 18 2019" broken test seems to be a consequence of https://sqbu-github.cisco.com/WebexIntelligence/duckling/pull/7
Could you look into that and the "after 3:00" test?

alhuang10:
Discussed offline with Lucien:

There are two Duckling outputs for “Feb 18 2019”. One of them is the correct one with expected resolved time. The other one sees “18” as a time of day and it interprets as “Feburary 2019 18 hours” for a resolved time of '2019-02-01T18:00:00.000-08:00’. This is another issue resulting from numbers being recognized as time-of-day. Likely will need to end up removing the rule for numbers as time of day and readdress the pattern of "today 6", "tomorrow 9" etc.

The “after 3:00” issue is similar where we have two outputs from Duckling, one of which is correct. This actually gets fixed with the removal of rule "ruleIntervalTODDash" since it’s being detected by that rule in addition to the correct one, ruleIntervalAfterFromSinceTime. Works fine when testing against the most recent Duckling master commit.

mindmeld/path.py
```diff
@@ -108,23 +108,25 @@
...
+)
 DUCKLING_OSX_PATH = os.path.join(RESOURCES_FOLDER, "duckling-x86_64-osx")
-DUCKLING_UBUNTU16_MD5 = "1e58b4e91d580d98c8ac4f5a69b0ebd1"
-DUCKLING_UBUNTU18_MD5 = "b9c4891b27731df97d7f01c33441a91c"
-DUCKLING_OSX_MD5 = "d01753261e6e7940b533f09dc17d0c19"
-DUCKLING_CENTOS8_MD5 = "f840176c2b96c0037edd9524faac5e93"
+DUCKLING_UBUNTU16_MD5 = "f32c9c1c8e83526dac430f4a25bdcf80"
+DUCKLING_UBUNTU18_MD5 = "843cc934c41585a875c32bd3f2913df0"
+DUCKLING_UBUNTU20_MD5 = "281aff210f4c0a4d8661f8a1f0aabc13"
+DUCKLING_OSX_MD5 = "ef80f9348a500a0b06cd0de85a414e90"
```
jre21, line 118: Can we change these to SHA256 checksums?  MD5 is pretty badly broken at this point.

serapio: done

jre21 (approved):
Looks good from my perspective, although I can't speak to the pytest failures.

vijay120 (approved):

jre21:
This has been sitting in limbo for nearly a month now, and assistant-nlp is still using the version of mindmeld from this branch and unable to pick up the improvements from 4.3.5rc10 as a result.  How can we unblock this?

alhuang10 (approved):

PR #331, opened 2021-06-29
Feature/disallowed intents

vijay120:
This PR implements https://github.com/cisco/mindmeld/pull/311

### DIFF
--- a/mindmeld/components/_util.py
+++ b/mindmeld/components/_util.py
@@ -17,6 +17,7 @@ import importlib
 import logging
 import enum
 from typing import Union, Optional, List
+from collections import defaultdict
 from ..exceptions import InvalidMaskError
 
 logger = logging.getLogger(__name__)
@@ -47,6 +48,14 @@ def _get_module_or_attr(module_name: str, func_name: str = None):
 
 
 class MaskState(enum.Enum):
+    """
+    This class encoded three NLP states:
+        unset: state when the user has neither allowed/denied the NLP component.
+            This state is needed to propagate state up/down the tree since we only
+            propagate state to unset nodes, never to user-defined nodes
+        allow: state when the user has explicitly allowed a node.
+        deny: state when the user has explicitly denied a node.
+    """
     unset = enum.auto()
     allow = enum.auto()
     deny = enum.auto()
@@ -78,6 +87,11 @@ class TreeNode:
 
 
 class TreeNlp:
+    """
+    This data structure encodes a NLP tree hierarchy where each node
+    encodes a mask state, based on which certain NLP components are allowed
+    or denied based on user input
+    """
     def __init__(self, nlp, mask_state=MaskState.unset):
         # root
         self.root = TreeNode('root', mask_state=mask_state)
@@ -166,27 +180,47 @@ class TreeNlp:
                 domain_node.mask_state = mask_state
                 return
 
-            for intent_node in self.get_intent_nodes(domain_name):
-                if intent_node.nlp_name != intent_name:
+            for intent_node in self.get_intent_nodes(domain_node.nlp_name):
+                if intent_name not in ('*', intent_node.nlp_name):
                     continue
 
                 if not entity_name:
                     intent_node.mask_state = mask_state
+                    # If the intent is * and it's terminal, eg. "domain.*", then
+                    # we mask the intent AND continue to iterate through the other
+                    # intents of the domain
+                    if intent_name == '*':
+                        continue
+                    # If the intent is not *, then it's terminal, eg. "domain.intent",
+                    # then we mask the intent and end the function's operations
                     return
 
-                for entity_node in self.get_entity_nodes(domain_name, intent_name):
-                    if entity_node.nlp_name != entity_name:
+                for entity_node in self.get_entity_nodes(domain_node.nlp_name,
+                                                         intent_node.nlp_name):
+                    if entity_name not in ('*', entity_node.nlp_name):
                         continue
 
                     if not role_name:
                         entity_node.mask_state = mask_state
+
+                        # If the entity is * and it's terminal, eg. "domain.intent.*", then
+                        # we mask the entity AND continue to iterate through the other
+                        # entities of the intent
+                        if entity_name == '*':
+                            continue
+                        # If the entity is not *, then it's terminal, eg. "domain.intent.entity",
+                        # then we mask the entity and end the function's operations
                         return
 
-                    for role_node in self.get_role_nodes(domain_name, intent_name, entity_name):
-                        if role_node.nlp_name != role_name:
+                    for role_node in self.get_role_nodes(domain_node.nlp_name,
+                                                         intent_node.nlp_name,
+                                                         entity_node.nlp_name):
+                        if role_name not in ('*', role_node.nlp_name):
                             continue
 
                         role_node.mask_state = mask_state
+                        if role_name == '*':
+                            continue
                         return
 
     def _sync_nodes(self):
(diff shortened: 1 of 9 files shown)


mindmeld/components/_util.py
```diff
@@ -38,3 +39,207 @@ def _get_module_or_attr(module_name: str, func_name: str = None):
     if func_name not in dir(m):
         raise ImportError(f"Cannot import {func_name} from {module_name}")
     return getattr(m, func_name)
+
+
+class TreeNode:
+    def __init__(self, value: str, parent=None, children=None, allow=None):
```
KKsharma99, line 45: It's not immediately clear that `value` refers to the name of node, can be confused with referring to the child nodes. Would be helpful to add some documentation clarifying this or consider a different attribute name such as `node_name` or `node_label`

mindmeld/components/_util.py
```diff
@@ -38,3 +39,207 @@ def _get_module_or_attr(module_name: str, func_name: str = None):
...
+                entities = nlp.domains[domain].intents[intent].entities
+                for entity in entities:
+                    entity_node = TreeNode(entity, parent=intent_node, allow=allow, children=[])
+                    intent_node.children.append(entity_node)
+                    for role in entities[entity].role_classifier.roles:
+                        role_node = TreeNode(role, parent=intent_node, allow=allow, children=None)
+                        entity_node.children.append(role_node)
+
+    def get_domains(self):
+        return self.root.children or []
```
KKsharma99, line 72: To clarify, the only time the empty list would be returned would be when `self.root.children` is None, or in other words, where there aren't any domains in NLP. However, in this case should we instead raise an exception in the __init__ and prevent the creation of TreeNLP objects without any domain nodes?

mindmeld/components/_util.py
```diff
@@ -38,3 +39,207 @@ def _get_module_or_attr(module_name: str, func_name: str = None):
...
+
+    def get_intents(self, domain: Union[str, TreeNode]):
+        if isinstance(domain, TreeNode):
+            domain = domain.value
+
+        for domain_node in self.root.children:
```
KKsharma99, line 78: If `self.root.children` can be `None`, we should also add a check for that here before the iteration step

mindmeld/components/_util.py
```diff
@@ -38,3 +39,207 @@ def _get_module_or_attr(module_name: str, func_name: str = None):
...
+    def get_entities(self, domain: Union[str, TreeNode],
+                     intent: Union[str, TreeNode]):
+
+        if isinstance(domain, TreeNode):
+            domain = domain.value
+
+        if isinstance(intent, TreeNode):
+            intent = intent.value
+
+        for intent_node in self.get_intents(domain):
```
KKsharma99, line 92: `self.get_intents` can be confused to represent the list of the intent names. Instead, consider clarifying this by renaming to `self.get_intent_nodes(self, domain)`. (This change would also apply to all levels of the TreeNLP hierarchy)

mindmeld/components/_util.py
```diff
@@ -38,3 +39,207 @@ def _get_module_or_attr(module_name: str, func_name: str = None):
...
+            entity = entity.value
+
+        if isinstance(role, TreeNode):
+            role = role.value
+
+        for domain_node in self.get_domains():
+            if domain_node.value != domain:
+                continue
+
+            if not intent:
```
KKsharma99, line 144: If either `entity`/`role` is not `None` and `intent` is `None` when passed into `update`. Then values in `entity`/`role` would be silently ignored and unused. Consider providing warnings for cases like this and/or including this in the documentation

mindmeld/components/_util.py
```diff
@@ -38,3 +39,207 @@ def _get_module_or_attr(module_name: str, func_name: str = None):
...
+        This function updates the NLP tree with mask values
+        Args:
+            allow: True is mask off, False is mask on
+            domain: domain of NLP
+            intent: intent of NLP
+            entity: entity of NLP
+            role: role of NLP
+        """
+
+        if isinstance(domain, TreeNode):
```
KKsharma99, line 128: Since this block of code is called a few times, could we instead make a function `convert_tree_node_to_values(List[Union[str,TreeNode]])` which iterates through a list and converts any TreeNode objects to string values?

mindmeld/components/_util.py
vijay120: done

mindmeld/components/_util.py
vijay120: I made sure in the construct of the TreeNode, we have this: `self.children = children or []`.

mindmeld/components/_util.py
vijay120: done

mindmeld/components/_util.py
vijay120: Added a validation check to the function to make sure all ancestors of a node are passed in.

mindmeld/components/_util.py
vijay120: done

tmehlinger (approved):
Looks good. Your tree implementation is really tight. 🙂

I'm still on the fence about failing open but I understand your rationale. What would you think about having a `strict` option (default off) that would cause NLP to raise an error on a conflict that would otherwise fail open?

tests/components/test_nlp.py
```diff
@@ -316,6 +316,67 @@ def test_allowed_entities(kwik_e_mart_nlp):
...
+    res = kwik_e_mart_nlp.process("please can you tell me if springfield is "
+                                  "possibly open at this time on friday",
+                                  allow_nlp=["store_info"],
+                                  deny_nlp=["store_info.get_store_hours.sys_time",
+                                            "store_info.get_store_hours.store_name"])
+    assert res['intent'] != 'get_store_hours'
+    assert 'sys_time' not in {entity['type'] for entity in res['entities']}
+    assert 'store_name' not in {entity['type'] for entity in res['entities']}
+
+    res = kwik_e_mart_nlp.process("please can you tell me if springfield is "
```
KKsharma99, line 365: Should we also test the case where allow_nlp is a subset of deny_nlp? For Example: `allow_nlp = store_info.get_store_hours` but `deny_nlp = store_info`

KKsharma99 (approved):
Your graph-based solution is comprehensive and elegant! Thanks for addressing previous comments. I recently added a suggestion for an additional test case.

mindmeld/components/_util.py
vijay120: done

serapio (approved):
Looks good to me. 

A related issue that this logic makes apparent, which we should be aware of but is out of scope here: Using allowed/disallowed intents can have pathological behavior because we predict domains and intents in separate stages. e.g. the disallowed intents could include all the most probable intents in the most probable domain, but since the mask is applied logically independent of the intent probabilities, the domain would still be allowed and predicted as most probable.

tests/components/test_nlp.py
```diff
@@ -316,6 +316,67 @@ def test_allowed_entities(kwik_e_mart_nlp):
...
+    assert 'sys_time' in {entity['type'] for entity in res['entities']}
+
+    res = kwik_e_mart_nlp.process("please can you tell me if springfield is "
+                                  "possibly open at this time on friday",
+                                  allow_nlp=["store_info"],
+                                  deny_nlp=["store_info.get_store_hours.sys_time"])
+    assert res['intent'] == 'get_store_hours'
+    assert 'sys_time' not in {entity['type'] for entity in res['entities']}
+
+    # If all entities in an intent are denied, that intent is also denied.
```
ritvikshrivastava, line 355: Is that ideal though? Imposing this as a rule of thumb can lead to bad intent classification. If an application is not performing well on entities and a user tries to block entities out, this would cause the intent classifier to be erroneous as well.

mindmeld/components/_util.py
```diff
@@ -38,3 +43,212 @@ def _get_module_or_attr(module_name: str, func_name: str = None):
...
+                for entity in entities:
+                    roles = self.get_role_nodes(domain, intent, entity)
+                    # sync down
+                    if entity.allow is not None:
+                        for role in roles:
+                            if role.allow is None:
+                                role.allow = entity.allow
+
+                    # sync up
+                    if roles and all(role.allow is False for role in roles):
```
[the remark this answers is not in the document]
vijay120: Such a great design pattern. Thank you!

ritvikshrivastava (approved):
This implementation is really neat! 😄 

A bit unsure about the case where all entities being denied leads to the intent to be denied (mentioned this in a comment). Our classifiers are already tightly coupled in a downward hierarchy, but this might lead them to be coupled the other way around as well.

tests/components/test_nlp.py
vijay120: fixed

vijay120: In order for this to work, I had to break backwards compatibility on a little used API `allowed_nlp_classes`, which is an internal representation of the allowed NLP components in a dict structure. Here, if the entities are not explicitly added as sub-keys, we don't recognize them. However, the children of explicitly stated domain, intent and roles are recognized.

mindmeld/components/_util.py
```diff
@@ -38,3 +43,212 @@ def _get_module_or_attr(module_name: str, func_name: str = None):
...
         raise ImportError(f"Cannot import {func_name} from {module_name}")
     return getattr(m, func_name)
+
+
+class TreeNode:
+    def __init__(self, nlp_name: str, parent=None, children=None, allow=None):
+        """
+        Constructor for the tree node
+        Args:
+            nlp_name: The name of the NLP component
```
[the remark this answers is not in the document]
vijay120: done

mindmeld/components/_util.py
```diff
@@ -38,3 +43,212 @@ def _get_module_or_attr(module_name: str, func_name: str = None):
...
+    def get_intent_nodes(self, domain: Union[str, TreeNode]):
+        domain, _, _, _ = self._convert_tree_node_to_values(domain)
+        for domain_node in self.root.children:
+            if domain_node.nlp_name == domain:
+                return domain_node.children
+        return []
+
+    def get_entity_nodes(self, domain: Union[str, TreeNode],
+                         intent: Union[str, TreeNode]):
+        domain, intent, _, _ = self._convert_tree_node_to_values(domain, intent)
```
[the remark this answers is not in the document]
vijay120: I think this operation is performant in the base case since all of the values are strings in the `update` function, so are just looping through strings and checking their type, so it should be pretty fast. Since this function is getting called elsewhere with TreeNode types, its worth to keep it without adding the extra "I have already checked the type" arg.

tests/components/test_nlp.py
vijay120: done

vijay120:
@tmehlinger I raise exceptions instead and localized the handling of exceptions in one place, hope this is a better design pattern.

PR #336, opened 2021-07-07
Added in-memory query cache

davidmfrey:
# What does this PR do?
Adds a completely in-memory query cache option.

# Why is this needed?
Some users of MindMeld have experienced extreme performance degradations during training due to the disk access patterns of the sqlite3 cache.

# How does this fix the problem?
The on-disk sqlite3 cache is still maintained, but an in-memory copy is also utilized by default.  This retains the previous behavior of the MindMeld framework without any adjustments.  If a user would prefer to use only the disk-based cache in order to reduce memory usage, they may set the env var `MM_QUERY_CACHE_IN_MEMORY=0` to disable the in-memory cache.

The disk-based cache is written to in batches, controlled by the env var `MM_QUERY_CACHE_WRITE_SIZE`.  The default is 1000.

Another change implemented in this PR is the ability to control whether CRF tagger model features are stored on disk or in memory.  The env var `MM_CRF_FEATURES_IN_MEMORY=0` directs the CRF model to store features on disk.  The default is 1, retaining backwards compatibility with previous MindMeld versions.


mindmeld/components/entity_resolver.py
```diff
@@ -139,27 +139,22 @@ def create_resolver(cls, app_path, entity_type, **kwargs):
...
-        )
+        er_config = config or get_classifier_config("entity_resolution", app_path=app_path)
         er_config = cls._correct_deprecated_er_config(er_config)
 
         resolver_type = er_config["model_settings"]["resolver_type"]
         cls._validate_resolver_type(resolver_type)
 
-        resource_loader = kwargs.pop(
-            "resource_loader",
-            ResourceLoader.create_resource_loader(app_path=app_path)
```
davidmfrey, line 153: This line was creating extra ResourceLoaders to pass in for the default value that were then discarded if not needed.

mindmeld/models/taggers/crf.py
```diff
@@ -26,6 +28,10 @@
 
 ZERO = 1e-20
 
+IN_MEMORY_FEATURES = bool(
```
vijay120, line 31: Would it be better worded as `STORE_CRF_FEATURES_IN_MEMORY`?

vijay120 (approved):
Looks good. I am not sure why the `mindmeld_init.sh` file is checked in. Also, could you add the two env vars in `source/userguide/getting_started.rst` in the last section?

PR #341, opened 2021-07-19
Add dump functionality to Entity Resolvers and  NativeQuestionAnswerer

murali1996:
### Update:

- Based on some discussion, calling .fit() in nlp's load process seems un-intuitive and hence necessary modifications are done in entity_resolvers.py and nlp.py to address this issue. See the related comments [here](https://github.com/cisco/mindmeld/pull/341#discussion_r717066197).
- Pending discussion on-- whether KB should be cached by NativeQA or should be data be loaded everytime QA is instantiated?
- Using the .dump() and load() in NLP pipeline leads to creation of new files-- .config.pkl and .pkl.hash for every resolver created in the WxA. Pending discussion on this as well.

---
---

This PR adds new functionality to entity resolvers (and thereby the question answerers built on top of them) to dump their state and load back from the dumped state. In addition to this end use, few disk-space optimizations are also added so that the KB data is not cached by both QA as well as the underlying resolvers. Lastly, minor bugs were fixed in entity resolvers and more comments/type-hints are added. 

Once approved, details will be added to Mindmeld docs for users to take advantage of the resolver dump and load methods.

### Problem:
Before this PR, if someone wishes to use a NativeQuestionAnswerer (i.e. the non-Elasticsearch QA), one should always call the `.load_kb()` method and fit the underlying resolver models to use the QA for search/inference. This behavior is unlike ElasticsearchQA which on other hand needs load-kb called only once. This is because the ElasticsearchQA loads back indices when the .get() and .build_search() methods are called directly. In cases wherein the data as well as the input configurations haven't changed from a previous run, of-course the NativeQuestionAnswerer need not fit the resolvers if there's a way to dump the resolvers' states in the previous run.

### Solution:
- In this PR, dump and load functionality are first added to the entity resolvers. This needed few changes in the resolver path (`path.py` file), ability for an embedder model in embedder based resolvers to dump its embeddings cache in a path derived from `app_path`(related changes also in `embedder_models.py` file).
- In order to keep the NLP pipeline unaltered, a minor change has been made to the `nlp.py` file. Previously, the load methods of resolvers simply redirected to the fit method. So the load method calls to entity resolver has been replaced with fit method in the nlp.py file.
- Note that the NLP pipeline doesn't  use the dump functionality of the resolvers as it leads to creation of more data files on the disk. This is implemented keeping both WxA as well as end users in mind.
- Because QA uses the resolver's dump functionality, some changes were implemented in the `question_answerer.py` file to encapsulate that and resolve the original problem of this PR.

### Discussions:
- See this [comment](https://github.com/cisco/mindmeld/pull/341#issuecomment-903170410) below.
- Related Issues: [361](https://github.com/cisco/mindmeld/issues/361)

### Backwards compatibility of the NLP pipeline
-  The NLP pipeline is mostly left unaltered by the introduction of dump and load methods of resolvers. The only change being that instead of calling `self.entity_resolver.load()`, we call `self.entity_resolver.fit(clean=False)` in the load method of `EntityProcessor`.  
- No issues/backward inconsistencies arise for ExactMatch, TFIDF and EmbedderCosSim/SentenceBert resolvers due to this change as they previously were simply calling .fit in their .load methods anyways.
- However, for the Elasticsearch resolver, that too only when the data or resolver configurations have changed, there is a backwards inconsistency (for good reasons!). Previously w/ Elasticsearch resolver when we called load method of the resolver, it checks if the index exists or not, if not then fits the resolver with latest data and configurations. Now, the resolver is fit with latest data and configurations irrespective of index's existence. So this slightly increases the loading time of the resolver as we are re-fitting/re-ingesting synonyms when we call ElasticsearchResolver.load().
- The only corner case is when a user creates a resolver index with specific configurations and now calls nlp pipeline with a different resolver configuration, which again is very unlikely. And when user changes data, it is always better to ingest the newer data into the index.


vembar (approved):

mindmeld/components/question_answerer.py
```diff
@@ -143,6 +144,14 @@ def _build_search(self, *args, **kwargs):
     def _load_kb(self, *args, **kwargs):
         raise NotImplementedError
 
+    @abstractmethod
+    def _dump(self, pkl_path):
```
vijay120, line 148: doc strings would be helpful here

mindmeld/components/question_answerer.py
```diff
@@ -143,6 +144,14 @@ def _build_search(self, *args, **kwargs):
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def load(self, pkl_path):
```
vijay120, line 152: same

mindmeld/components/question_answerer.py
```diff
@@ -227,6 +236,23 @@ def load_kb(self, index_name, data_file, **kwargs):
...
 
+    def dump(self, pkl_path):
+        folder = os.path.dirname(pkl_path)
+        os.makedirs(folder, exist_ok=True)
+        self._dump(pkl_path)
+
+        metadata = {
+            "app_path": self.app_path,
+            "app_namespace": self.app_namespace,
+            "resource_loader": self.resource_loader,
```
vijay120, line 247: the app_path can be derived from the resource_loader. Moreover, the resource_loader is a huge object with tons of references and gazetteer information that can be derived from the app files itself, so a lot of duplicated memory is going to be stored repeatedly to disk.

Is there more specific information in the resource loader that we can dump instead of the whole object?

vijay120, line 247: With the recent refactor of the QA components, why do we have a reference to the resource loader in the QA classes in the first place?

mindmeld/components/question_answerer.py
```diff
@@ -227,6 +236,23 @@ def load_kb(self, index_name, data_file, **kwargs):
             raise ValueError(msg)
         self._load_kb(index_name, data_file, **kwargs)
 
+    def dump(self, pkl_path):
+        folder = os.path.dirname(pkl_path)
+        os.makedirs(folder, exist_ok=True)
+        self._dump(pkl_path)
+
+        metadata = {
+            "app_path": self.app_path,
```
vijay120, line 245: One issue with this implementation is `_dump ` for `NativeQuestionAnswerer` reference `ALL_INDICES`, which is a class variable. So if multiple apps are loaded into memory at the same time, `ALL_INDICES` will be updated with all of them, not just the one related to this app_path. So the metadata will not accurately partition the indices based on app path.

mindmeld/components/question_answerer.py
```diff
@@ -601,15 +633,21 @@ def _get_location(value: Union[dict, str, List]):
...
                 return ",".join([str(value["lat"]), str(value["lon"])])
-            elif isinstance(value, str) and "," in value and len(value.split(",")) == 2:
-                # eg. "37.77,122.41"
+            # eg. "37.77,122.41"
+            elif isinstance(value, str) and len(value.split(",")) == 2:
+                for val in value.split(","):
+                    try:
+                        val = float(val)
+                    except ValueError:
+                        # eg. "Brown, Mia"
```
vijay120, line 645: log an error here

mindmeld/components/question_answerer.py
```diff
@@ -2741,11 +2789,122 @@ def is_vector_field(self):
...
+    Factory class for creating QuestionAnswerers
+
+    usage
+        >>> question_answerer = QuestionAnswererFactory.create_question_answerer(**kwargs)
+        >>> question_answerer.load_kb(...)
+        >>> question_answerer.get(...) # .get(...) or .build_search(...)
+    """
+
+    @classmethod
+    def create_question_answerer(cls, **kwargs):
```
vijay120, line 2803: I would rather have a clearer set of optional args here like config and app_path. Ideally, config should be a data class with a fixed set of attributes rather than a dict

mindmeld/components/question_answerer.py
```diff
@@ -2741,11 +2789,122 @@ def is_vector_field(self):
...
+        return config
+
+    @staticmethod
+    def _correct_deprecated_qa_config(config):
+        """
+        for backwards compatability
+          if the config is supplied in deprecated format, its format is corrected and returned,
+          else it is not modified and returned as-is
+
+        deprecated usage
```
vijay120, line 2834: great docs!

mindmeld/components/question_answerer.py
```diff
@@ -2741,11 +2789,122 @@ def is_vector_field(self):
...
+                        "query_type": "keyword",  # or "text", "embedder", "embedder_keyword", etc.
+                        ...
+                    }
+                }
+        """
+
+        if not config.get("model_settings", {}).get("query_type"):
+            model_type = config.get("model_type")
+            if not model_type:
+                msg = f"Invalid 'model_type': {model_type} found while creating a QuestionAnswerer"
```
vijay120, line 2855: model_type will always be none here, so it doesnt convey the issue. I think you should emit the whole config and say it's invalid.

mindmeld/components/question_answerer.py
```diff
@@ -2741,11 +2789,122 @@ def is_vector_field(self):
...
+                raise ValueError(msg)
+            if model_type in QUESTION_ANSWERER_MODEL_MAPPINGS:
+                raise ValueError(
+                    "Could not find `query_type` in `model_settings` of question answerer")
```
vijay120, line 2859: emit the whole config here as well

vijay120 (commented):
A few comments. I am still a bit unclear on why people are trying to dump QA models when they can just load_kb it. Is it because load_kb takes a long time?

mindmeld/components/question_answerer.py
murali1996: Thanks for pointing it out. It is unnecessary to dump resource loader and the can be created during instantiation through `ResourceLoader.create_resource_loader(self.app_path)` (app being obtained from the loaded pkl file)

mindmeld/components/question_answerer.py
murali1996: Corrected the code accordingly. The config here is the QUESTION_ANSWERER_CONFIG type dict passed to the QuestionAnswererFactory directly instead of using the QUESTION_ANSWERER_CONFIG from the config.py file. This is to enable QA instantiation without having to create an app.

mindmeld/components/question_answerer.py
murali1996: I don't think so. If the user specifies older format of configs and not the newer ones, the config would be something like 
```python
        config = {
            "model_type": "keyword",  # or "text", "embedder", "embedder_keyword", etc.
            "model_settings": {
                ...
            }
        }
```
in which case, the model_type is not None.

murali1996: This is for the cases where the user fails to supply the `model_type` key in the config. If provided, the model_type is no longer None right?

mindmeld/components/question_answerer.py
murali1996: the `if model_type in QUESTION_ANSWERER_MODEL_MAPPINGS:` conditional is for scenario wherein we anticipated an older config version's `model_type` values (because we did not see `query_type` key which was not available in older config version), but we observed newer config version's `model_type`. This is an edge case when users mix up the versions.

murali1996:
> A few comments. I am still a bit unclear on why people are trying to dump QA models when they can just load_kb it. Is it because load_kb takes a long time?

Based on our offline discussion regarding this PR and some analysis around the runtimes, we decided to add the following changes to QA module as part of this PR to bring _native_ QA usability akin to _elasticsearch_:

- Dump models built as part of _native_ QA's .load_kb() method to the path `~/.cache/mindmeld`, so that users don't have to call.load_kb() in every environment before querying indices.
- Implement a new method `_load_field_info()` as part of _native_ QA, similar to its functionality in _elasticsearch_ QA, to load the previously dumped models when directly calling .get() and .build_search() methods w/o calling .load_kb() method in an user environment.
- Add notes in docs informing users that for replicating the state of the QA module when using _native_ QA, especially when using in a deployment environment, they'll have to copy model dump files in `~/.cache/mindmeld`.

mindmeld/components/question_answerer.py
murali1996: removed resource loader from question answerer dump as well as question answerer itself

mindmeld/components/question_answerer.py
murali1996: This is not an error due to user's KB data, it is just a check whether it is a location type string or a normal text string. I don't think a log is necessary here

mindmeld/components/classifier.py
```diff
@@ -226,12 +226,12 @@ def fit(self,
         queries = self._resolve_queries(queries, label_set)
 
         new_hash = self._get_model_hash(model_config, queries)
-        cached_model = self._resource_loader.hash_to_model_path.get(new_hash)
```
murali1996, line 229: Just a change of variable name for better readability.

mindmeld/components/entity_recognizer.py
```diff
@@ -107,21 +107,21 @@ def fit(self,
...
+        # this PR, this will be set to `cached_model_path=<>.pkl` path and the self.load() takes
+        # care of loading a NoneType model. Had it been `cached_model_path=None` like previously,
+        # the following code skips the `load_cached` check and directly attempts to create a new
+        # model. This is not an issue in domain and intent classifiers as the .fit() method is not
+        # called when there are less than 2 domains/intents.
+
+        if incremental_timestamp and cached_model_path:
             logger.info("No need to fit.  Previous model is cached.")
             if load_cached:
-                self.load(cached_model)
```
murali1996, line 124: Just a change of variable name for better readability.

mindmeld/components/classifier.py
```diff
@@ -226,12 +226,12 @@ def fit(self,
...
+        cached_model_path = self._resource_loader.hash_to_model_path.get(new_hash)
 
-        if incremental_timestamp and cached_model:
+        if incremental_timestamp and cached_model_path:
             logger.info("No need to fit.  Previous model is cached.")
             if load_cached:
-                self.load(cached_model)
```
murali1996, line 234: Just a change of variable name for better readability.

mindmeld/components/role_classifier.py
```diff
@@ -103,13 +103,13 @@ def fit(self,
...
         new_hash = self._get_model_hash(model_config, queries)
-        cached_model = self._resource_loader.hash_to_model_path.get(new_hash)
+        cached_model_path = self._resource_loader.hash_to_model_path.get(new_hash)
 
-        if incremental_timestamp and cached_model:
+        if incremental_timestamp and cached_model_path:
             logger.info("No need to fit. Previous model is cached.")
             if load_cached:
                 # load() sets self.ready = True
-                self.load(cached_model)
```
murali1996, line 112: Just a change of variable name for better readability.

mindmeld/components/classifier.py
```diff
@@ -226,12 +226,12 @@ def fit(self,
...
+                self.load(cached_model_path)
                 return True
```
murali1996, line 235: Just a change of variable name for better readability.

mindmeld/components/entity_recognizer.py
```diff
@@ -107,21 +107,21 @@ def fit(self,
...
+                self.load(cached_model_path)
                 return True
```
murali1996, line 125: Just a change of variable name for better readability, and few changes to comments.

mindmeld/components/entity_resolver.py
```diff
@@ -256,32 +305,44 @@ def _trim_results(self, results, top_n):
...
+        # Obtain upto top_n docs and sort them as final result
         n_scores = len(results)
-        if n_scores < top_n:
-            if top_n != 20:
-                # log only if a value different from default value is specified
-                msg = f"Retrieved only {len(results)} entity resolutions instead of asked " \
-                      f"number {top_n} for entity type {self.type}"
-                logger.info(msg)
-            return results
-        # else, n_scores >= top_n
```
murali1996, line 270: Bug. If there are less than top_n resolutions, they weren't sorted before returning. Fixed this.

mindmeld/components/entity_resolver.py
```diff
@@ -151,36 +153,78 @@ def create_resolver(cls, app_path, entity_type, config=None, resource_loader=Non
...
-        self.ready = False  # bool, True if the model is fit by calling .fit()
+        self.dirty = False  # bool, True if exists any unsaved data/model that can be saved
+        self.ready = False  # bool, True if the model is already fitted or loaded
+        self.hash = None
+
+        # resources for creating and maintaining processed entity map
+        self._aug_lower_case = None
+        self._aug_title_case = None
+        self._aug_normalized = None
+        self._normalize_aliases = None
```
vembar, line 220: Aren't these specific to non-ES resolvers?

mindmeld/components/entity_resolver.py
```diff
@@ -1220,50 +1490,56 @@ def __init__(self, app_path, **kwargs):
...
-            cache_path = get_entity_resolver_cache_file_path(app_path, hashid)
-            settings.update({"cache_path": cache_path})
-            er_config.update({"model_settings": settings})
+        # No `app_path` specified because we now have dump and load methods that take a caching path
+        # as argument. Previously, we just dumped and loaded from a default cache_path ascertained
+        # by that embedder model itself (through model_id and app_poth) i.e. one path for all usages
+        # of that particular embedder model. However, with the introduction of dump and load method
+        # to resolvers, one-path-for-all is no longer an optimal solution and is deprecated.
+        # Note that we do not specify any `cache_path` key field as well in the er_config here!
+        self._embedder_model = create_embedder_model(app_path=None, config=er_config)
```
mhuang2173, line 1513: just wonder shouldn't we be initializing embedder model based on the ER config in `load()`?

mindmeld/components/entity_resolver.py
```diff
@@ -930,8 +1114,52 @@ def _construct_whitelist_query(entity, weight=1, use_phons=False):
...
+            with open(resolver_config_path, "rb") as fp:
+                _resolver_configs = pickle.load(fp)
+                fp.close()
+            self._use_double_metaphone = _resolver_configs["use_double_metaphone"]
+        except (TypeError, FileNotFoundError):
+            msg = f"Unable to load previously used configs for the {self.__class__.__name__} " \
+                  f"resolver while loading."
+            logger.warning(msg)
+
+        # elasticsearch loading
```
mhuang2173, line 1162: i think we might need to have additional check on top of index existence check (maybe based on `dirty` flag and also the `hash`?) because we could have an existing index created based on different config than what we load from serialized file?

Also it looks like we are not loading the `hash` from dump()?

mindmeld/components/nlp.py
```diff
@@ -1674,18 +1675,18 @@ def unload(self):
...
-                self.domain,
-                self.intent,
-                self.type,
-                timestamp=incremental_timestamp,
-            )
-            self.role_classifier.load(
-                incremental_model_path if incremental_timestamp else model_path
-            )
-            self.entity_resolver.load()
+            self.entity_resolver.fit(clean=False)
```
mhuang2173, line 1689: Just wonder do we need this for backward compatibility reasons? I think in general we would prefer not to do `fit()` when loading the models so that the behavior across NLP pipeline is more consistent and predictable. (so that users won't be surprised by any kind of model building happens during `load()`) I think the changes in this PR help us move forward on to that direction since we don't need to do fit() in the resolver implementation of load() anymore and we should keep it that way if we can?

mindmeld/components/entity_resolver.py
```diff
@@ -466,8 +558,78 @@ def _predict(self, nbest_entities, allowed_cnames=None):
...
+            path (str): A .pkl file path where the resolver will be dumped. The model hash will be
+                dumped at {path}.hash file path.
+            cache_kb (bool, default: True): If True, saves the processed entity map in disk and
+                loads it back when the load method is called. If False, the resolver model/algorithm
+                is dumped but not the processed entity map necessitating the need to manualy set it
+                upon loading (useful in cases like NativeQuestionAnswerer!)
+        """
+
+        # dump underlying resolver model/algorithm/embeddings
+        self._dump(path)
```
vembar, line 582: Optional: Maybe we should add a message to the user consistent with other models - "Saving entity resolver: domain=<>, intent=<>, entity=<>"

mindmeld/components/entity_resolver.py
```diff
@@ -955,28 +1183,178 @@ def _load(self):
...
+                    for item in self.processed_entity_map["items"][cname]:
+                        item_value = copy.copy(item)
+                        item_value.pop("whitelist", None)
+                        item_value.update({"score": score})
+                        item_value.update({"top_synonym": synonym})
+                        values.append(item_value)
+        except KeyError:
+            logger.warning(
+                "Failed to resolve entity %r for type %r; "
+                "set 'clean=True' for computing embeddings of newly added items in mappings.json",
```
vembar, line 1308:
```suggestion
                "set 'clean=True' for computing TF-IDF scores of newly added items in mappings.json",
```

mindmeld/components/entity_resolver.py
murali1996: They are! Moving them back to the non-ES resolvers from the base resolver class.

mindmeld/components/nlp.py
murali1996: Yeah, I absolutely agree to you point Marvin. I did so keeping in mind about the WxA; I realized that if we dump and load resolver models (exact match in WxA case), we will be using more disk space as we need to dump the entity map (i.e. a processed KB data) in the dump() call and load it back in the load() call of NLP pipeline. So I would like to confirm if extra disk space is a problem in case of WxA before making the change you suggested. @vijay120 @vembar @mhuang2173

mindmeld/components/entity_resolver.py
murali1996: Because we dump the processed entity map in the dump method and load it back in load method, we need not check for the hash equivalence as it will be the same data. This check is similar to other NLP components like domain classifier, etc.

mindmeld/components/entity_resolver.py
murali1996: Thanks for noticing that, @mhuang2173 . Had to make few changes in entity_resolver.py to account for that. Please take a look now.

At the time of your comment, the QA module actually had the resolver config settings stored and hence it was not leading to any sort of bug. But I realized that could be a problem for embedder resolver when used outside the context of question answerers. Hence, I modified some code related to the way the config is saved in the resolvers and it should resolve this issue.

mindmeld/components/nlp.py
murali1996: Update: It is more logical to not to dump a copy of the KB data when we call dump() for a resolver. In the latest push, I added 'entity_map' argument in the load() method to address this. So what happens now is that, the KB data is loaded in the load() method as well similar to fit() method and hence I can replace `'self.entity_resolver.fit(clean=False)'`with `'self.entity_resolver.load()'`. Also in case of question answerers, we can pass the KB data looked up the question answerers directly to the load() method.

mindmeld/components/question_answerer.py
```diff
@@ -823,194 +1097,159 @@ def _update_data_type(self, value):
...
+            self._embedding_resolver = None  # an embedding based entity resolver
+
+        def __repr__(self):
+            return f"{self.__class__.__name__} " \
+                   f"field_name: {self.field_name} " \
+                   f"data_type: {self.data_type} " \
+                   f"has_text_resolver: {self.has_text_resolver} " \
+                   f"has_embedding_resolver: {self.has_embedding_resolver}"
+
+        def update_resource(
```
mhuang2173, line 1224: maybe we can consider having another method `load_resource()` which can be used in `build_search()` in question answerer class and also here based on hash?

mhuang2173 (commented):
looking good, a few more comments

mindmeld/models/embedder_models.py
```diff
@@ -793,6 +901,12 @@ def _encode(self, text_list):
                 encoded_vecs.append(np.average(vl, axis=0))
         return encoded_vecs
 
+    def _tokenize(self, text):
+        return [
+            t["entity"] for t in
+            self.text_preparation_pipeline.tokenize_and_normalize(text)
+        ]
+
     def dump(self):
```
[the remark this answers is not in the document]
murali1996: So embedder models' dump functionality is just dumping and loading a cache object;, and no configs are dumped. Because all embedder classes have the cache in the same way, I guess we don't need an embedder specific dump? @mhuang2173 

And for the overlapping of dump for Glove, I would like to take that up in PR 325 with some more modifications to the Glove embedder class.

mindmeld/components/entity_resolver.py
```diff
@@ -433,12 +310,132 @@ def predict(self, entity_or_list_of_entities, top_n=20, allowed_cnames=None):
...
+
+        Args:
+            path (str): A .pkl file path where the resolver has been dumped
+            entity_map (Dict[str, Union[str, List]]): Entity map if passed in directly instead of
+                loading from a file path
+        """
+        msg = f"Loading {self.__class__.__name__} entity resolver for entity_type {self.type}"
+        logger.info(msg)
+
+        if self.ready:
```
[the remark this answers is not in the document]
murali1996: Added

mindmeld/components/entity_resolver.py
```diff
@@ -364,31 +228,54 @@ def fit(self, clean=False, entity_map=None):
...
-        # load data: list of canonical entities and their synonyms
-        entity_map = entity_map or self._load_and_get_entity_map()
-        entity_map = self._format_entity_map(entity_map)
-        if not entity_map.get("entities", []):
+        entity_map = entity_map or self._get_entity_map()
+        entities_data = entity_map.get("entities", [])
+        if not entities_data:
             self._no_trainable_canonical_entity_map = True
             self.ready = True
+            self.dirty = True
```
vijay120, line 248: its unclear why dirty is true here

mindmeld/components/entity_resolver.py
```diff
@@ -364,31 +228,54 @@ def fit(self, clean=False, entity_map=None):
...
+            self._fit(clean, entity_map)
+        except Exception as e:
+            msg = f"Error in {self.__class__.__name__} while fitting the resolver model with " \
+                  f"clean={clean}"
+            raise EntityResolverError(msg) from e
+        self.hash = new_hash
+
         self.ready = True
-        return
+        self.dirty = True
```
vijay120, line 276: same here

mindmeld/components/nlp.py
```diff
@@ -1636,60 +1638,92 @@ def __init__(
...
         )
 
         self.progress_bar = progress_bar
         if isinstance(self.progress_bar, tqdm):
             self.progress_bar.total += 1
 
+    @property
+    def ready(self):
+        # either one can trigger a call to dump method in build_resursive()
+        return self._ready_rc or self._ready_er
```
vijay120, line 1651: what is _ready_rc? I think both _ready_rc and _ready_er should be renamed more clearly

mindmeld/components/question_answerer.py
```diff
@@ -454,106 +708,189 @@ class Indices:
...
+            }
+            index_all_ids = self._indices_all_ids[index_name]
+            return metadata, index_all_ids
 
-            # clear field resources
+        def _get_index_metadata_from_disk(self, index_name):
+            cache_path = self._get_index_cache_path(index_name)
+            with open(cache_path, "rb") as opfile:
+                metadata = pickle.load(opfile)
+                opfile.close()
```
vijay120, line 813: I think this line is redundant since the code block takes care of it

mindmeld/components/question_answerer.py
```diff
@@ -454,106 +708,189 @@ class Indices:
...
+            index_all_ids = metadata.pop("__all_ids")
```
vijay120, line 814: Why does this have double unders `__all_ids `?

mindmeld/components/question_answerer.py
```diff
@@ -454,106 +708,189 @@ class Indices:
...
+            if index_name in self:
+                # in-memory indices are always fit indices
+                return
+            elif self.is_available(index_name):
+                index_resources, index_all_ids = self.get_index_metadata(index_name)
+                # Instantiate one resource loader for all resolvers being loaded in Native QA
+                _resource_loader = NativeQuestionAnswerer.get_resource_loader()
+                for field_name, field_resource in index_resources.items():
+                    field_resource.update_resource(
+                        id2value={},
```
vijay120, line 877: why empty dict here?

mindmeld/components/question_answerer.py
```diff
@@ -825,194 +1119,167 @@ def _update_data_type(self, value):
...
-            return []
-
         def _get_resolvers_entity_map(self, id2value):
             """
             converts id2value into an entity map format and returns it
             """
+
+            def _get_resolvers_whitelist(value):
+                if isinstance(value, (set, list)):
+                    return list(value)[1:]
```
vijay120, line 1184: why `[1:]`?

mindmeld/components/question_answerer.py
murali1996: (Related [comment]https://github.com/cisco/mindmeld/pull/341#discussion_r720574972) 

So by passing id2value as en empty dict, we can trigger loading of resolvers. But because it doesn't make a good code style, I now modified the code to have two separate methods- update_resource and load_resource, the latter taking care of only loading resolvers for FieldResources. So this line 877 that you pointed no longer exists.

@mhuang2173 @vijay120

Issue #307, opened 2021-05-14
safe_path annotation changes C:\ to C_\ and path is not found


Efaq:
Hi

While building an example application, I noticed that some files were not being found, even though they were there. Looking closer I noticed that the absolute path was starting with C_\ instead of C:\. Looking closer, this is happening because of this function in _path.py_, which changes ":" with "_":

```
# Helpers
def safe_path(func):
    """A decorator to make the path safe by replacing unsafe characters"""

    @wraps(func)
    def _wrapper(*args, **kwargs):
        res = func(*args, **kwargs)
        if isinstance(res, tuple):
            return tuple(map(lambda x: x.replace(":", "_") if x else x, res))
        elif isinstance(res, str):
            return res.replace(":", "_")
        else:
            return res

    return _wrapper
```

Specifically in my case, the annotated function causing the problem is:

```
@safe_path
def get_entity_gaz_path(app_path, entity):
    """Gets the path to the gazetteer text file for a given entity.

    Args:
        app_path (str): The path to the app data.
        entity (str): An entity under the application.

    Returns:
        (str) The path for a mapping of the entity

    """
    return GAZETTEER_TXT_PATH.format(app_path=app_path, entity=entity)
```

The question is: is this behavior expected? Might I be missing some newbie detail here?

Thank you for the patience.

vijay120:
@vrdn-23  Assigning this to you

vrdn-23:
Hey @Efaq ,

Thanks for pointing this issue out to us! Could you give me some more information about your current OS setup? Are you using Windows? Also could you point out to me which resources were not being found and which resources were being found?

Thanks in advance!

Efaq:
Hi

I'm working on Windows 10, using PyCharm 2019.3 as IDE. The files not found that I noticed are the training files inside of "domains" subfolders.

It should be easy to reproduce it. Pick whichever use case that uses domain models and add some training files to some domain. When training the classifier for this domain, some files would not be found, and when checking the path one can notice that the C_ will start the path, instead of C:.

vrdn-23:
@Efaq The fix for this has been merged. Please do pull the latest code from Mindmeld and please let us know and reopen the issue if it persists.

PR #342, opened 2021-07-20
Fix safe path to update rel path

vrdn-23:
This fix should fix the issue with Windows systems mentioned in #307. The issue only arises when the user specifies the absolute path. Works as expected in Windows systems when given the relative path.

### DIFF
--- a/mindmeld/path.py
+++ b/mindmeld/path.py
@@ -160,10 +160,11 @@ def safe_path(func):
     @wraps(func)
     def _wrapper(*args, **kwargs):
         res = func(*args, **kwargs)
+        # Replace with relpath to avoid replacing : (root dir) with _ in Windows
         if isinstance(res, tuple):
-            return tuple(map(lambda x: x.replace(":", "_") if x else x, res))
+            return tuple(map(lambda x: os.path.relpath(x).replace(":", "_") if x else x, res))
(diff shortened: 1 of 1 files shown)


vrdn-23:
This is still a WIP. Do not MERGE

vrdn-23:
I've tested it across a couple of the mindmeld blueprints and both building and loading the models should work now on Windows with this fix. 
PR can be merged!

mindmeld/path.py
```diff
@@ -160,10 +160,11 @@ def safe_path(func):
     @wraps(func)
     def _wrapper(*args, **kwargs):
         res = func(*args, **kwargs)
+        # Replace with relpath to avoid replacing : (root dir) with _ in Windows
         if isinstance(res, tuple):
-            return tuple(map(lambda x: x.replace(":", "_") if x else x, res))
+            return tuple(map(lambda x: os.path.relpath(x).replace(":", "_") if x else x, res))
         elif isinstance(res, str):
-            return res.replace(":", "_")
+            return os.path.relpath(res).replace(":", "_")
```
vijay120, line 167: Arent we returning relpaths sometimes and abs path on other times? I think we should standardize it, no?

vijay120 (approved):

PR #345, opened 2021-07-26
Add duckling mapping system entities for JA, DE, and IT

alhuang10:
JIRA Ticket: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-252947

Make sure JA/DE/IT have the following mappings:
"sys_duration"
"sys_number"
"sys_ordinal"
"sys_time"
"sys_interval"
"sys_phone-number"

### DIFF
--- a/mindmeld/constants.py
+++ b/mindmeld/constants.py
@@ -140,6 +140,7 @@ DUCKLING_TO_SYS_ENTITY_MAPPINGS = {
         "sys_ordinal",
         "sys_time",
         "sys_interval",
+        "sys_phone-number",
     ],
     "el": ["sys_duration", "sys_number", "sys_ordinal", "sys_time", "sys_interval"],
     "en": [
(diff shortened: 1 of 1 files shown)


melisamson (approved):
These changes will help with the current annotation effort in Japanese and German. Maybe eventually we should align the supported system entities on all of our supported languages? Does it depend on Duckling's coverage for each language?

serapio (approved):
@KKsharma99 does the annotator need this to be language-dependent ?

KKsharma99:
> @KKsharma99 does the annotator need this to be language-dependent ?

The mapping object is used internally by the SpacyAnnotator to determine which sys_entities are supported by duckling. Certain entities that are found by Spacy are required to be supported by duckling in order to be labelled. With this mapping, the SpacyAnnotator determines whether it needs to remove or keep the label. (_remove_unresolvable_entities() in line 438 of auto_annotator.py).

PR #349, opened 2021-08-04
don't strictly validate elasticsearch type

serapio:
elasticsearch-py 7.14 strictly validates its Elasticsearch flavor (https://github.com/elastic/elasticsearch-py/commit/b63d00513802ddefa14720043b9dadb85a9e556e)
which makes it reject older versions of elasticsearch that we recommend. Until we are ready to force users to update, we need to not use elasticsearch-py 7.14.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,7 @@ setup(
             'torch~=1.7.0; python_version>="3.6"',
             'transformers~=3.5.1; python_version>="3.6"',
             'sentence-transformers~=0.3; python_version>="3.6"',
-            "elasticsearch>=7.0",
+            'elasticsearch>=7.0,<7.14',
         ],
         "examples": [
             'connexion>=2.7.0; python_version>="3.6"',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -15,7 +15,7 @@
     "click-log==0.1.8",
     "distro~=1.3",
     "dvc>=1.8.1",
-    "elasticsearch>=5.0",
+    "elasticsearch>=5.0,<7.14",
```
serapio, line 18: Why do we have elasticsearch both here and in extras?

jre21: Also, can we add a comment here on why we're limiting the version?  Something like `# elasticsearch-py 7.14 breaks backwards compatibility with servers prior to 7.11`

vijay120 (approved):

jre21 (approved):

jre21:
One note for future reference: Amazon is [planning](https://aws.amazon.com/blogs/opensource/keeping-clients-of-opensearch-and-elasticsearch-compatible-with-open-source/) to release a fork of the elasticsearch clients that restores compatibility with both the Apache 2.0-licensed Elasticsearch server and their own OpenSearch fork.  Once that's available we'll have a viable upgrade path.

PR #350, opened 2021-08-04
pinned es

vijay120:
We had to pin elasticsearch due to this issue: https://github.com/elastic/elasticsearch-py/issues/1639


serapio:
same as #349

PR #351, opened 2021-08-06
Move dvc to extras-requirements

alhuang10:


vijay120 (commented):
Could you also add a check of the dep's existence in the cli's dvc code and provide a helpful message if the dep is not available?

vijay120 (approved):

Issue #352, opened 2021-08-08
"TypeError: Invalid custom validation function type" when defining a custom evaluation function for Auto-Fill.


ronitkishore:
Trying to run this example from the Dialogue Manage documentation:

```
from mindmeld.core import FormEntity

def test_for_money(request):
    return True if '$' in request.text else False

form_transfermoney = {
    'entities':[
        FormEntity(
            entity='account_type',
            role='account_from',
            responses=['Sure. Transfer from which account?'],
            retry_response=["That account is not correct. Transfer from which account?"],
            ),
        FormEntity(
            entity='account_type',
            role='account_to',
            responses=['To which account?'],
            hints=['checking', 'checkings'] # can be only from this list
            ),
        FormEntity(
            entity='sys_amount-of-money',
            responses=['And, how much do you want to transfer?'],
            custom_eval=test_for_money # validates the user-response for this entity
            ),                         # using this custom developer-defined function
                                       # checking for '$' sign.
        ],
    'max_retries': 1,
    'exit_keys': ['cancel', 'quit', 'exit'],
    'exit_msg': "Sorry I cannot help you with this transfer. Please try again."
    }
```

When I run `nlp = NaturalLanguageProcessor(app_path='test')`, it yields this: `TypeError: Invalid custom validation function type.`

I added some print statements on line 1042 in core.py: `print(custom_eval, type(custom_eval))` and this the where the error happens: `<function test_for_money at 0x14784d200> <class 'str'>`. When `callable()` is called on the `<class 'str'>` type object, the `TypeError` is raised.

PR #353, opened 2021-08-09
AutoFill Bugfix - Custom Evaluation Schema Incorrect

ritvikshrivastava:
Fixes issue #352 .

### DIFF
--- a/mindmeld/components/schemas.py
+++ b/mindmeld/components/schemas.py
@@ -397,7 +397,15 @@ class FormEntitySchema(Schema):
     value = fields.Dict(allow_none=True)
     default_eval = fields.Boolean(default=True)
     hints = fields.List(fields.String(), allow_none=True)
-    custom_eval = fields.String(allow_none=True)
+    custom_eval = fields.Method("serialize_custom_eval",
+                                deserialize="deserialize_custom_eval")
+
+    def serialize_custom_eval(self, form):  # pylint: disable=no-self-use
+        if form:
+            return form.custom_eval
+
+    def deserialize_custom_eval(self, custom_eval):  # pylint: disable=no-self-use
+            return custom_eval
 
     def serialize_value(self, form):  # pylint: disable=no-self-use
         if form:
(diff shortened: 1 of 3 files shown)


mindmeld/components/schemas.py
```diff
@@ -397,10 +398,17 @@ class FormEntitySchema(Schema):
     value = fields.Dict(allow_none=True)
     default_eval = fields.Boolean(default=True)
     hints = fields.List(fields.String(), allow_none=True)
-    custom_eval = fields.String(allow_none=True)
+    custom_eval = fields.Method("serialize_custom_eval",
+                                default=None)
+    def serialize_custom_eval(self, form):  # pylint: disable=no-self-use
+        if form.custom_eval:
+            f = open("custom_eval", "wb")
```
tmehlinger, line 405: Better to do this with a context manager to ensure the file is closed without having to do it explicitly:
```python
with open('custom_eval', 'wb') as f:
    pickle.dump(form.custom_eval, f)
```

mindmeld/core.py
```diff
@@ -1043,8 +1043,15 @@ def __init__(
 
         if not self.entity or not isinstance(self.entity, str):
             raise TypeError("Entity cannot be empty.")
+
         if self.custom_eval and not callable(custom_eval):
-            raise TypeError("Invalid custom validation function type.")
+            try:
+                f = open(custom_eval, "rb")
```
tmehlinger, line 1049: same here as above

mindmeld/core.py
```diff
@@ -1059,6 +1059,21 @@ def to_dict(self):
...
+        self.func_registry = {}
+
+    def add_to_func_registry(self, func_name, func):
+        """Populates the function registry map.
+
+        Args:
+            func_name (str): Name to be used as key for the function.
+            func (func): Callable function.
+        """
+        self.func_registry[func_name] = func
```
vijay120, line 1074: Its probably a good idea to implement a getter here instead of exposing the underlying array DS to the client functions

vijay120 (approved):

PR #356, opened 2021-08-11
Few fixes related to entity recognizer classifier

murali1996:
Update:
1. Changes to deprecate  _self.\_model\_config_ are reverted and instead, a TODO is added to deprecate the variable.
---

This PR implements few changes related to entity recognizer classifier:
1. Models for entity recognition are unnecessarily dumped even when there are no entity types. Similar to domain and intent classifiers, this can be avoided.
1. The variable _self.\_model\_config_ seems to have no use apart from being dumped and loaded in metadata. With Phase-1 neural nets PR, the configs of the model are now already saved in a file `<domain/intent/entity>.config.pkl`
1. While evaluating entity recognizers, we do not evaluate on those recognizers who have only one entity type

### DIFF
--- a/mindmeld/components/classifier.py
+++ b/mindmeld/components/classifier.py
@@ -444,6 +444,7 @@ class Classifier(ABC):
                 self._model.dump(path)
 
             hash_path = path + ".hash"
+            os.makedirs(os.path.dirname(hash_path), exist_ok=True)
             with open(hash_path, "w") as hash_file:
                 hash_file.write(self.hash)
 
--- a/mindmeld/components/classifier.py
+++ b/mindmeld/components/classifier.py
@@ -439,8 +439,8 @@ class Classifier(ABC):
             if self._model:
                 # sometimes a model might be NoneType, eg. in role classifiers, in which case,
                 # no dumping is required. While loading such models, the model_path (.pkl)
-                # will not be found and the helpers.load_model() will return None, whic makes it
-                # backwards compataible to loading a NoneType model
+                # will not be found and the helpers.load_model() will return None, which makes it
+                # backwards compatible to loading a NoneType model
                 self._model.dump(path)
 
             hash_path = path + ".hash"
--- a/mindmeld/components/entity_recognizer.py
+++ b/mindmeld/components/entity_recognizer.py
@@ -54,7 +54,6 @@ class EntityRecognizer(Classifier):
         self.domain = domain
         self.intent = intent
         self.entity_types = set()
-        self._model_config = None
 
     def _get_model_config(self, **kwargs):  # pylint: disable=arguments-differ
         """Gets a machine learning model configuration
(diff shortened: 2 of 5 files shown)


mindmeld/components/nlp.py
```diff
@@ -1134,6 +1134,26 @@ def get_entity_processors(self, label_set=None):
 
     def _build(self, incremental=False, label_set=None, load_cached=True):
         """Builds the models for this intent"""
+        entity_types = self.entity_recognizer.get_entity_types(label_set=label_set)
```
vembar, line 1137: Isn't this just getting overwritten by line 1170?

mindmeld/components/classifier.py
```diff
@@ -444,6 +444,7 @@ def dump(self, model_path, incremental_model_path=None):
                 self._model.dump(path)
```
vembar, line 444:
```suggestion
                # sometimes a model might be NoneType, eg. in role classifiers, in which case,
                # no dumping is required. While loading such models, the model_path (.pkl)
                # will not be found and the helpers.load_model() will return None, which makes it
                # backwards compatible to loading a NoneType model
```

vembar, line 444: Typos in the comment above statement:
whic --> which
compataible --> compatible

mindmeld/components/nlp.py
```diff
@@ -1134,6 +1134,26 @@ def get_entity_processors(self, label_set=None):
...
+
+        # Context:
+        # In cases of len(domains)==1 or len(intents)==1 in NaturalLanguageProcessor and
+        # DomainProcessor respectively, the self.ready flag is unchanged to remain 'False'. This
+        # leads to not triggering the ._dump() and .unload() abstract methods in the
+        # ._build_recursive() method, and subsequently, the .load() method in .build() method takes
+        # care of loading a NoneType model. However, in case of entity recognizers, the _dump method
+        # must be called to ave metadata information which is later used to ascertain if there were
```
vembar, line 1145:
```suggestion
        # must be called to have metadata information which is later used to ascertain if there were
```

mindmeld/components/nlp.py
```diff
@@ -1134,6 +1134,26 @@ def get_entity_processors(self, label_set=None):
...
+        # no entity_types or not. Hence, the self.ready flag must be modified to True.
```
vembar, line 1146:
```suggestion
        # entity_types or not. Hence, the self.ready flag must be modified to True.
```

mindmeld/components/nlp.py
```diff
@@ -1134,6 +1134,26 @@ def get_entity_processors(self, label_set=None):
...
+        # Edge-case:
+        # When incremental build is initiated and when the self.ready flag is obtained as output of
+        # self.entity_recognizer.fit() method, the _dump() method mustn't be called if there are no
+        # new changes to data/config. So to enbale this edge case flag, we do not check for
```
vembar, line 1150:
```suggestion
        # new changes to data/config. So to enable this edge case flag, we do not check for
```

vembar (approved):
Just some typos. Looks good.

mindmeld/components/nlp.py
murali1996: Yeah, it gets overwritten. Missed to remove this line from previous edition of code created as part of PR 325. Removed this line now. Thanks for pointing it out.

mindmeld/components/entity_recognizer.py
```diff
@@ -119,16 +128,20 @@ def fit(self,
...
         examples, labels = self._get_examples_and_labels(queries)
 
-        # Build entity types set
-        self.entity_types = set()
-        for label in labels:
-            for entity in label:
-                self.entity_types.add(entity.entity.type)
+        if examples:
+            # Build entity types set
+            self.entity_types = set()
```
vijay120, line 133: `self.entity_types = {entity.entity.type for label in labels for entity in label}`

mindmeld/resource_loader.py
```diff
@@ -455,8 +455,11 @@ def _load_cached_models(self):
                 hash_val = open(file_path, "r").read()
                 classifier_file_path = file_path.split(".hash")[0]
                 if not os.path.exists(classifier_file_path):
+                    # In some cases, there exists hash file but without a corresponding serialized
```
vijay120, line 458: great docs

vijay120 (approved):

Issue #358, opened 2021-08-18
Query cache does not update with change in Tokenizers config


vrdn-23:
With the work for Multilingual Tokenizers finally merged (PR #328), users will now have an option to set different tokenizer and normalizers for the text preparation pipeline. However, the current Mindmeld query cache implementation does not update the cache with changes in text preparation config. The Mindmeld query cache currently only updates with changes in the training data.

vrdn-23:
Fixed and merged with #362

PR #362, opened 2021-09-07
Update query cache on TextPreparationPipeline config update

vrdn-23:
This is a PR that solves the issue detailed in #358. The main idea of the PR is derived from [this blog post](https://chadrick-kwag.net/json-dumpingserializing-custom-python-classes/). 

This IMO seems to be a pretty neat way to deal with the current issue and also provides support for newer pipeline components/custom components by developers, but I am definitely open to any other design suggestions/changes which everyone might suggest.

~I will probably add a couple of unit tests to go along with this PR~, I have added unit tests to ensure the hash changes with different configurations of Text Preparation Pipeline. I would appreciate everyone's input now whether this indeed is the right approach to take!
TIA!


mindmeld/query_cache.py
```diff
@@ -53,7 +58,7 @@ def __init__(self, app_path):
 
         cursor = self.disk_connection.cursor()
 
-        if not self.compatible_version():
+        if not self.compatible_version() or not self.compatible_text_pipeline_config():
```
vijay120, line 61: `compatible_text_pipeline_config ` can be consolidated into compatible_version by doing the following SQL query:
```
        cursor = self.disk_connection.cursor()
        try:
            cursor.execute("""
            SELECT version_number FROM version WHERE version_number=(?) and text_pipeline_hash=(?);
            """, (ProcessedQuery.version,self.text_pipeline_hash))
            if len(cursor.fetchall()) == 0:
                # version does not match
                return False
            return True
        except Exception:  # pylint: disable=broad-except
            return False
```

vijay120 (commented):
I think it would be easier to just take the hash of the JSON encoding of config.py.

mindmeld/query_cache.py
```diff
@@ -43,6 +44,10 @@ def copy_db(source, dest):
 
     def __init__(self, app_path):
         # make generated directory if necessary
+        text_preparation_pipeline = TextPreparationPipelineFactory.create_from_app_path(
```
davidmfrey, line 47: Are there any side effects to instantiating a TextPreparationPipelineFactory that we should be aware of here?

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -192,6 +195,38 @@ def stem_word(self, word):
...
+        """
+        Method defined to obtain Hash value of TextPreparationPipeline.
+
+        Args:
+            None.
+
+        Returns:
+            256 character hash representation of current TextPreparationPipeline config (str) .
+        """
+        string = json.dumps(self, cls=TPPJSONEncoder, sort_keys=True)
```
davidmfrey, line 227: Nice catch on sorting the keys.  FYI, Python dicts keep items in insertion order since 3.6, but it's not officially documented until 3.7, so sorting these is good form since we support 3.6.

davidmfrey (commented):
Nice work so far!  I have a couple of thoughts that we may want to take into account.

- We're ending up with two version numbers stored when we really only need one.  We could either make the version coming from ProcessedQuery automated as well or dispense with that versioning scheme and simply bump the  MindMeld version when changes are made that would impact the ProcessedQuery schema.  Either way, we can combine the two versions into one text version key via sha265 and only store one in the db.
- If we have the ability to version the entity resolver (Duckling or any other backends), we should go ahead and add that in as well.  This will catch errors that can occur when we upgrade the entity recognizer but don't manually bump the version number.
- I'm not sure how often developers will be switching between configurations that will cause a change to the version; we do have the option of maintaining N versions of the query cache to support multiple configurations if needed.

mindmeld/query_cache.py
```diff
@@ -126,6 +131,24 @@ def compatible_version(self):
...
 
+    def compatible_text_pipeline_config(self):
+        """
+        Checks to see if the cache db file exists and that the text preparation pipeline config
+        matches the current pipeline config.
+        """
+
+        cursor = self.disk_connection.cursor()
+        try:
+            cursor.execute("""
```
tmehlinger, line 142: If you're only counting, you don't want to fetch an arbitrary number of rows. You can also simplify the condition below this.

```python
row = cursor.execute("""
    SELECT COUNT(version_number) FROM version WHERE text_pipeline_hash=(?)
""", (self.text_pipeline_hash,)).fetchone()
return row[0] > 0
```

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -627,3 +662,21 @@ def _construct_pipeline_component(  # pylint: disable=W0640
...
+
+    Args:
+        None.
+
+    Returns:
+        Custom JSON Encoder class (json.JSONEncoder) .
+    """
+
+    def default(self, o):
+        if "tojson" in dir(o):
```
tmehlinger, line 680: We can can make this a bit more robust by check specific conditions we need to satisfy.
```python
tojson = getattr(o, 'tojson', None)
if callable(tojson):
    return tojson()
```

tmehlinger (approved):
a couple minor things, looks good otherwise :)

mindmeld/query_cache.py
```diff
@@ -126,6 +131,24 @@ def compatible_version(self):
...
+            SELECT version_number FROM version WHERE text_pipeline_hash=(?);
+            """, (self.text_pipeline_hash,))
+            if len(cursor.fetchall()) == 0:
```
KKsharma99, line 145:
```
version_matches = len(cursor.fetchall()) != 0
return version_matches
```

mindmeld/query_cache.py
```diff
@@ -126,6 +131,24 @@ def compatible_version(self):
...
+                # version does not match
+                return False
+            return True
+        except Exception:  # pylint: disable=broad-except
```
KKsharma99, line 149: We should try to stay away from broad exceptions when possible as they can make debugging unseen edge cases difficult in the future

KKsharma99:
> I think it would be easier to just take the hash of the JSON encoding of config.py.

I agree with Vijay's comment here. This approach would be more robust to capturing configuration changes across multiple backend components and would be quite maintainable.

mindmeld/query_cache.py
KKsharma99: Only one edge case related to instantiating another TextPreparationPipeline is if a developer creates a custom Application object by passing in a TextPreparationPipeline object directly, this would propagate all the way down to the QueryFactory. By instantiating a separate TextPreparationPipeline here we would not be taking into account a custom object passed in from the top-level.

However, this is a very specific edge case as it requires to create a custom Application using a pre-instantiated TextPreparationPipeline and add new data.

davidmfrey: So is it relatively easy to pass in the TextPreparationPipeline object instead of instantiating it here?  That could also make it easier to test this class.

KKsharma99: If there are class methods to grab the existing object of the Application, Resource Loader or Query Factory then you can get the object through that (assuming that they've been implemented as Singleton classes).

mindmeld/query_cache.py
vrdn-23: Fixed to sqlite3.Error to catch only sqlite DB related exceptions

mindmeld/query_cache.py
vrdn-23: Fixed it to Travis' suggestion!

mindmeld/text_preparation/text_preparation_pipeline.py
vrdn-23: Nice! and done!

mindmeld/query_cache.py
vrdn-23: Updated!

mindmeld/query_cache.py
vrdn-23: Refactored code to pass TextPreparationPipeline into QueryCache via QueryFactory. Nice catch @davidmfrey @KKsharma99 !

mindmeld/query_cache.py
vrdn-23: Removed the dependence on an external ProcessedQuery version number.

vrdn-23:
> > I think it would be easier to just take the hash of the JSON encoding of config.py.
> 
> I agree with Vijay's comment here. This approach would be more robust to capturing configuration changes across multiple backend components and would be quite maintainable.

@KKsharma99 I discussed this with @vijay120 offline. The reason I didn't want to rely on the hash of the config.py is because any small changes related to model configurations would trigger an update of the cache even if the text pipeline hadn't really changed. Seems contradictory to the idea of the cache if we're sure certain changes like n-gram settings or BOW parameters of the model will not affect the values stored in the cache. Changing the cache whenever we have a TPP update or a version update seems to be the best way to make sure we don't update the cache unnecessarily too often

vrdn-23:
> Nice work so far! I have a couple of thoughts that we may want to take into account.
> 
> * We're ending up with two version numbers stored when we really only need one.  We could either make the version coming from ProcessedQuery automated as well or dispense with that versioning scheme and simply bump the  MindMeld version when changes are made that would impact the ProcessedQuery schema.  Either way, we can combine the two versions into one text version key via sha265 and only store one in the db.
> * If we have the ability to version the entity resolver (Duckling or any other backends), we should go ahead and add that in as well.  This will catch errors that can occur when we upgrade the entity recognizer but don't manually bump the version number.
> * I'm not sure how often developers will be switching between configurations that will cause a change to the version; we do have the option of maintaining N versions of the query cache to support multiple configurations if needed.

@davidmfrey 
1) I've removed the dependency on the ProcessedQuery version as any changes to that class would hopefully get reflected in a bump in a MM rc version at least.
2) Nice catch. I've included Duckling version in the hash json to update based on duckling changes.
3) I've created #365 to address the issue of supporting multiple configurations.

vrdn-23:
I've addressed the review comments in the PR. New changes include:
- Pass TextPreparationPipeline into query cache instead of creating new one from app_path
- Removed ProcessedQuery version dependency
- Added duckling version to update as well
- Changed DB structure a teeny tiny bit to make it suitable for #365 
-

mindmeld/query_cache.py
```diff
@@ -68,11 +69,11 @@ def __init__(self, app_path):
         # Create table to store the data version
         cursor.execute("""
         CREATE TABLE IF NOT EXISTS version
-        (version_number INTEGER PRIMARY KEY);
+        (text_pipeline_hash TEXT PRIMARY KEY, config_id INTEGER);
```
davidmfrey, line 72: In light of the comment above I'd like to see this as simply `schema_version_hash TEXT PRIMARY KEY` and we can drop any other versioning schemes.

vrdn-23: Done!

mindmeld/query_cache.py
```diff
@@ -41,8 +41,9 @@ def copy_db(source, dest):
             cursor.execute(statement)
         dest.commit()
 
-    def __init__(self, app_path):
+    def __init__(self, app_path, text_preparation_pipeline):
         # make generated directory if necessary
+        self.text_pipeline_hash = text_preparation_pipeline.get_hashid()
```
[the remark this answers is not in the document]
vrdn-23: That is actually a much better idea! Thanks, fixed accordingly!

mindmeld/query_cache.py
```diff
@@ -108,22 +109,19 @@ def __del__(self):
 
         self.disk_connection.close()
 
-    def compatible_version(self):
+    def compatible_version_or_text_prep_pipeline(self):
```
davidmfrey, line 112: Let's revert this name back since we don't have any knowledge of the text_prep_pipeline in this module anymore.

vrdn-23: Fixed

davidmfrey (approved):
Looks good.  Just one more comment for you.  Thanks for your patience with all the requested changes!

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -627,3 +663,20 @@ def _construct_pipeline_component(  # pylint: disable=W0640
             raise TypeError(
                 f"{component} must be of type String or {expected_component_class.__name__}."
             )
+
+
+class TPPJSONEncoder(json.JSONEncoder):
```
vijay120, line 668: I think its better to state the full form here: `TextPreparationPipelineJSONEncoder`

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -627,3 +663,20 @@ def _construct_pipeline_component(  # pylint: disable=W0640
...
+    Custom Encoder class defined to obtain recursive JSON representation of a TextPreparationPipeline.
+
+    Args:
+        None.
+
+    Returns:
+        Custom JSON Encoder class (json.JSONEncoder) .
+    """
+
+    def default(self, o):
```
vijay120, line 679: Do we want to handle the "else" case here where tojson is not callable?

vijay120 (approved):
great work!

mindmeld/text_preparation/text_preparation_pipeline.py
vrdn-23: Done

mindmeld/text_preparation/text_preparation_pipeline.py
vrdn-23: Fixed to throw a TextPreparationPipelineError for missing tojson method

PR #364, opened 2021-09-13
Added script to wait for ElasticSeaech cluster to be in green state before starting the Mindmeld Docker container

Zozman:
Currently if you use the Docker container provided in the Mindmeld repository, there is a chance that Mindmeld will attempt to use the ElasticSearch cluster before it is fully initialized (see [this repository](https://github.com/Zozman/mmError) for an example of this issue when the container is started.  When this happens, the Entity Resolver is not loaded.

Therefore this commit adds a script for the container to run before initializing the Mindmeld instance in the `ENTRYPOINT` so that the ElasticSearch cluster is healthy before starting the instance.

Also added `docker-compose` files for starting containers with `docker-compose up` because I thought they would be helpful.


docker_containers/mindmeld_docker/Dockerfile
```diff
@@ -90,4 +93,5 @@ ENTRYPOINT export LC_ALL=C.UTF-8 && \
     pip3 install mindmeld --upgrade && \
     su mindmeld -c "ES_JAVA_OPTS='-Xms1g -Xmx1g' /elasticsearch/bin/elasticsearch -d" && \
     mindmeld num-parse --start && \
+    /waitForClusterToBeGreen.sh && \
```
vijay120, line 96: Change the file name to `wait_for_cluster_to_turn_green.sh` to reflect the snake case of the other file names in the repo

vijay120 (approved):
Looks good!

docker_containers/mindmeld_docker/Dockerfile
Zozman: Done

PR #367, opened 2021-09-23
Calculating Token Indices for Sentences that Are not Space Delimited

KKsharma99:
Problem: The root issue is in the _get_form_details() function of the NestedEntity class (core.py), this is where the token spans are calculated for raw, processed, and normalized text. This function calculates the starting and ending token based on the number of whitespace characters in the text (see screenshot). This assumption is problematic for Japanese text such as "1時間のミーティングを30分後から予約" which is not delimited by space. Hence, this creates invalid values for token spans and can cause span assertion errors for Japanese text.

Solution: Redesigned the token span calculation method to more robust to Japanese by solely relying on the Normalized Text by default when calculating token indices. Refactored the existing _get_form_details() function to three separate function, one of which uses the Normalized Text as default after making a copy of the current query object (to avoid transformations in the original query)

### DIFF
--- a/mindmeld/core.py
+++ b/mindmeld/core.py
@@ -702,7 +702,12 @@ class NestedEntity:
         def _get_form_details(query_span, offset, form_in, form_out):
             span_out = query.transform_span(query_span, form_in, form_out)
             full_text = query.get_text_form(form_out)
+            orig_text = span_out.slice(full_text)
+
+            span_out = query.transform_span(query_span, form_in, 2)
+            full_text = query.get_text_form(2)
             text = span_out.slice(full_text)
+
             # The span range is till the span_out or max to the second last char
             tok_start = 0
             span_range = min(span_out.start, len(full_text) - 1)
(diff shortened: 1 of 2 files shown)


mindmeld/core.py
```diff
@@ -699,25 +700,89 @@ def from_query(
...
+            if offset is not None:
+                offset_out = query_for_norm_transformation.transform_index(
+                    offset, form_in, TEXT_FORM_NORMALIZED
+                )
+                tok_offset = len(full_norm_text[:offset_out].split())
+                tok_span.shift(-tok_offset)
+
+            return tok_span
+
+        def _get_form_details(query_span, offset, form_in, form_out):
```
vijay120, line 762: I think _get_form_details and _get_token_span can be combined since the `if offset is not None:` cases seem close to each other. There are other ops that are similar as well

vijay120 (commented):
One comment

mindmeld/core.py
```diff
@@ -699,25 +700,75 @@ def from_query(
...
             tok_start = 0
-            span_range = min(span_out.start, len(full_text) - 1)
-            for idx, current_char in enumerate(full_text[:span_range]):
+            span_range = min(span_out.start, len(full_norm_text) - 1)
+            for idx, current_char in enumerate(full_norm_text[:span_range]):
                 # Increment the counter only if a whitespace follows a non-whitespace
-                next_char = full_text[idx + 1]
+                next_char = full_norm_text[idx + 1]
                 if not current_char.isspace() and next_char.isspace():
                     tok_start += 1
```
serapio, line 720: It's still a little worrisome to me that we are depending on whitespace to mark tokenization, even in the normalized text. If someone's custom tokenizer decides "85 %" is a single token or that "co-op" is a single token normalized as "co op", that shouldn't break the whole system. Perhaps we should switch to using a character that is already system-reserved, like two pipes `||` to distinguish it from the entity annotation.

serapio (approved):
This at least fixes the immediate problem.

PR #368, opened 2021-09-23
Adding new English Paraphraser that retains entities 

vembar:
This PR contains changes required to integrate the new English Paraphraser that retains entity information.
The model will be downloaded from https://mindmeld-binaries.s3.amazonaws.com/paraphraser/paraphrase_retain_entities.zip
 which is made public.

We now add an additional key to our augmentation config - `retain_entities`. It is set to `False` by default and will use the original paraphraser model that does not guarantee entity retention.
```
AUGMENTATION_CONFIG = {
    "augmentor_class": "EnglishParaphraser",
    "batch_size": 8,
    "retain_entities": True,
    "paths": [
        {
            "domains": ".*",
            "intents": ".*",
            "files": "train*.txt",
        }
    ],
    "path_suffix": "-augmented.txt"
}
```
The Multilingual model will simple ignore the `retain_entities` keyword as of now.

The new paraphraser follows the steps below:
1) Strips the annotations and converts queries to `<unannotated query text> <EOS> <entity values> ` format. For example: "when will the {main street|store_name} store open?" --> `"when will the main street store open </s> main street"`
2) Generates paraphrases with the entity values retained as is in most of the paraphrases
3) Uses the original entity annotations to find matching spans in the generated paraphrases
4) Additionally replaces found matches with random values from gazetteers (if available for that entity type) to increase n-gram variation in queries
5) Dump the markdown query with entity and role annotations in the appropriate file


mindmeld/augmentation.py
```diff
@@ -241,6 +262,167 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
...
+            processed_entities (list(ProcessedQuery)): List of ProcessedQuery
+                                                       with entity annotations
+        """
+        model_inputs = []
+        processed_entities = []
+        for query in queries:
+            processed_query = load_query(query,
+                                         query_factory=self._resource_loader.query_factory)
+            if self.retain_entities:
+                text = [processed_query.query.text.strip().lower(), EOS_TOKEN]
```
ritvikshrivastava, line 308: I see that this setup lowercases both the text and entities. Would that imply that the default model will be case-sensitive and the entity retention one won't be?

vembar: Yes, that is correct. It shouldn't matter much since most of the feature extractors for mindmeld works on lower case text anyway.

mindmeld/augmentation.py
```diff
@@ -15,19 +15,33 @@
...
+from .path import EMBEDDINGS_FOLDER_PATH, PARAPHRASER_FILE_PATH, PARAPHRASER_MODEL_PATH
+from .models.taggers.embeddings import TqdmUpTo
 
 logger = logging.getLogger(__name__)
 
 # pylint: disable=R0201
 
 SUPPORTED_LANGUAGE_CODES = ["en", "es", "fr", "it", "pt", "ro"]
+EOS_TOKEN = '</s>'
+NUM_PARAPHRASES = 10
```
KKsharma99, line 42: Nit: Consider renaming to `DEFAULT_NUM_PARAPHRASES` for clarity

mindmeld/augmentation.py
```diff
@@ -241,6 +262,167 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
...
+                            reporthook=t.update_to)
+            except ConnectionError as e:
+                logger.error("Model download failed with error: %s", e)
+                return
+        try:
+            with zipfile.ZipFile(PARAPHRASER_FILE_PATH, 'r') as zip_ref:
+                zip_ref.extractall(EMBEDDINGS_FOLDER_PATH)
+        except zipfile.BadZipfile:
+            logger.error("Unable to extract zip file. Try downloading the model again.")
+        return
```
KKsharma99, line 285: Nit: We could remove this last return unless it is intentionally kept to make it clear that None is being returned

mindmeld/augmentation.py
```diff
@@ -62,6 +76,7 @@ def create_augmentor(self):
             )
         try:
             # Validate configuration input
+            app_path = self.config.get("app_path")
```
KKsharma99, line 79: Clarification: .get() for "app_path" and "retain_entities" among other values are all in this try block. However, aside from "app_path" the other .get() invocations have alternate defaults. If this is the case, then the 'except KeyError as e' would only apply if the app_path was missing?

mindmeld/augmentation.py
```diff
@@ -215,23 +232,27 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
...
-
-        model_name = "tuner007/pegasus_paraphrase"
-        torch_device = (
+        self.retain_entities = retain_entities
+        if self.retain_entities:
+            if not os.path.exists(PARAPHRASER_MODEL_PATH):
+                self._download_model()
+            model_name = PARAPHRASER_MODEL_PATH
+        else:
+            model_name = "tuner007/pegasus_paraphrase"
```
KKsharma99, line 241: Consider making this into a global variable? A comment or variable name explaining what this path is or where it comes from would be helpful

mindmeld/augmentation.py
```diff
@@ -241,6 +262,167 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
...
+            processed_query = load_query(query,
+                                         query_factory=self._resource_loader.query_factory)
+            if self.retain_entities:
+                text = [processed_query.query.text.strip().lower(), EOS_TOKEN]
+                for entity in processed_query.entities:
+                    text.append(entity.text.lower())
+                processed_entities.append(processed_query)
+                model_inputs.append(' '.join(text))
+            else:
+                model_inputs.append(processed_query.query.text.strip())
```
KKsharma99, line 314: Nit: Consider adding this line before the if-else block; 
`query_text = processed_query.query.text.strip()`

Then in the if block you can do:
`text = [query_text.lower(), EOS_TOKEN]`

And in the else block:
`model_inputs.append(query_text)`

mindmeld/augmentation.py
```diff
@@ -241,6 +262,167 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
...
+        return model_inputs, processed_entities
+
+    def _replace_with_random_entity(self, paraphrase_text, entity_matches):
```
KKsharma99, line 317:
```suggestion
    def _replace_with_random_gaz_entity(self, paraphrase_text, entity_matches):
```

mindmeld/augmentation.py
ritvikshrivastava: It would apply either in the case or when an incorrect key is invoked for the AUGMENTATION MAP when a wrong augmentor class is provided (line 94)

mindmeld/augmentation.py
```diff
@@ -241,6 +262,167 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
...
+        model_inputs = []
+        processed_entities = []
+        for query in queries:
+            processed_query = load_query(query,
+                                         query_factory=self._resource_loader.query_factory)
+            if self.retain_entities:
+                text = [processed_query.query.text.strip().lower(), EOS_TOKEN]
+                for entity in processed_query.entities:
+                    text.append(entity.text.lower())
+                processed_entities.append(processed_query)
```
ritvikshrivastava, line 311: This variable is called processed_entities and stored processed queries as a list.

Consider renaming it to ``processed_input_queries`` as in line 438?

mindmeld/augmentation.py
```diff
@@ -241,6 +262,167 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
...
+                logger.error("Model download failed with error: %s", e)
+                return
+        try:
+            with zipfile.ZipFile(PARAPHRASER_FILE_PATH, 'r') as zip_ref:
+                zip_ref.extractall(EMBEDDINGS_FOLDER_PATH)
+        except zipfile.BadZipfile:
+            logger.error("Unable to extract zip file. Try downloading the model again.")
+        return
+
+    def _prepare_inputs(self, queries):
```
ritvikshrivastava, line 287: We can add this as an abstract function in the abstract augmentor class? Since it looks like it's being inherited by both the English and multi-lingual augmentors.

mindmeld/augmentation.py
ritvikshrivastava: Also, consider adding "EN" since this number applies only to that model.

mindmeld/augmentation.py
KKsharma99: In that case we may want to move the other get() invocations before the try block to make this more clear

mindmeld/augmentation.py
```diff
@@ -241,6 +262,167 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
...
+            if not running_start:
+                # First entity in the query
+                running_start = span.start
+                # Append text seen so far
+                new_paraphrase_text.append(paraphrase_text[:running_start])
+            else:
+                # Calculate new start based on previously replaced entity length
+                running_start += (span.start - previous_end)
+                # Append text seen between entities
+                new_paraphrase_text.append(paraphrase_text[previous_end:span.start])
```
KKsharma99, line 344: Since value of previous_end in the first iteration is 0, you may consider removing the if block

```suggestion
          # Calculate new start based on previously replaced entity length
          running_start += (span.start - previous_end)
          # Append text before first entity or text seen between entities
          new_paraphrase_text.append(paraphrase_text[previous_end:span.start])
```

mindmeld/augmentation.py
```diff
@@ -241,6 +262,167 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
...
+            gaz = None
+            # If not a system entity and gazetteer is available, load it
+            if not Entity.is_system_entity(entity.type):
+                gaz = self._resource_loader.get_gazetteer(entity.type)['entities']
+            if gaz:
+                # Create new Entity and Span based on random gaz entry for entity type
+                random_entity = random.sample(gaz, 1)[0]
```
KKsharma99, line 351:
```suggestion
                random_gaz_entity_text = random.sample(gaz, 1)[0]
```

mindmeld/augmentation.py
```diff
@@ -241,6 +262,167 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
...
+            previous_end = span.end + 1
+        # Append any leftover text
+        new_paraphrase_text.append(paraphrase_text[previous_end:])
+
+        processed_query = self._resource_loader.query_factory.create_query(
+                                                                  ''.join(new_paraphrase_text))
+        final_entities = []
+        for (entity, span) in replaced_spans_entities:
+            query_entity = QueryEntity.from_query(query=processed_query, span=span, entity=entity)
+            final_entities.append(query_entity)
```
KKsharma99, line 372: Consider separating this section into its own function and also using list comprehension here

```suggestion
        def create_query_entities(replaced_spans_entities):
              return [
                   QueryEntity.from_query(query=processed_query, span=span, entity=entity)
                   for entity, span in replaced_spans_entities
              ]
```

mindmeld/augmentation.py
```diff
@@ -241,6 +262,167 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
...
+
+            Args:
+                paraphrases (list(str)): List of unannotated paraphrases of queries
+                processed_queries (list(ProcessedQuery)): List of their corresponding original ProcessedQuery
+
+            Return:
+                paraphrases (list(str)): List of paraphrased queries.
+        """
+        valid_paraphrases = []
+        for i in range(len(processed_queries)):
```
KKsharma99, line 386: Nit: Alternatively consider

`for i, processed_query in enumerate(processed_queries):`

And changing `processed_queries[i]` to `processed_query` in line 388

mindmeld/augmentation.py
```diff
@@ -359,9 +546,25 @@ def _translate(self, *, queries, model, tokenizer, **kwargs):
...
+
+            Returns:
+                unannotated queries (list(str))
+
+        """
+        processed_queries = []
+        for query in queries:
+            processed_query = load_query(query)
+            processed_queries.append(processed_query.query.text.strip())
+        return processed_queries
```
KKsharma99, line 563: It's a bit unclear here that 'processed_queries' refers to a list of 'str' objects and not 'ProcessedQuery' objects. Perhaps consider renaming this to 'processed_query_texts'?

mindmeld/augmentation.py
vembar: Had some thoughts about this - We can move the method currently implemented in the Multi-Lingual Paraphraser (to return unannotated queries) as the default and let the inheritors override it if they want to. However, I'm not sure if that is the correct default (since in the case of the English paraphraser, we now do require those annotations). On the flip side, making it an abstract method will potential force replication of the method as is in the Multi-Lingual Paraphraser to other future inheritors (maybe language specific models(?)). 

What are your thoughts on what makes a good default? @vijay120 @KKsharma99 @ritvikshrivastava

ritvikshrivastava: If we think future language specific models will be based on the similar MT strategy, then we should make the multi-lingual method as the abstract. 

Otherwise, the current implementation with an empty abstract is fine. We can re-visit this if and when another model is added in this case.

source/userguide/augmentation.rst
```diff
@@ -68,27 +69,36 @@ The following values can be updated in the config to customize it:
...
 
-    mindmeld augment --app-path <app_path> --language "en"
+    mindmeld augment --app_path <app_path> --language "en"
 
 In the config for this paraphraser class, the ``'augmentor_class'`` should be set to :class:`EnglishParaphraser`.
 
+Retaining Entity Annotations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The default English model ignores entity annotations and paraphrases the full query (along with entities). If you wish to retain the entity values and annotations in your query as is in the paraphrases, you can set the ``'retain_entities'`` key to ``True`` in the config.
```
ritvikshrivastava, line 95:
```suggestion
The default English model ignores entity annotations and paraphrases the full query (along with entities). If you wish to retain the original query's entity values and annotations in the paraphrases, you can set the ``'retain_entities'`` key to ``True`` in the config.
```

source/userguide/augmentation.rst
```diff
@@ -68,27 +69,36 @@ The following values can be updated in the config to customize it:
...
+In addition, if a gazetteer is provided for the entity type, the system will automatically replace the entity values with randomly sampled entities from the gazetteer. This is recommended to improve the n-gram diversity in your queries.
```
ritvikshrivastava, line 96:
```suggestion
In addition, if a gazetteer is provided for an entity type, the system will automatically replace entity values with randomly sampled entities from this gazetteer. This is recommended to improve the n-gram diversity in your queries.
```

ritvikshrivastava (approved):
Can discuss the abstract method further, but other changes LGTM! 😄

mindmeld/augmentation.py
KKsharma99: If we do not expect inheritors to use this method as it is, we should avoid making it the default in the base class. A middle ground here is to see if we can refactor this function in a way that the logic we expect to be used by the inheritors is made into a separate function. The function with the common logic can be placed in the abstract class and any custom alterations on top can be done via an overridden method.

What aspects of this method would be specific to EN, is it the syntax of <unannotated text> <EOS> <entity values>?

KKsharma99 (approved):
Looks good!

mindmeld/augmentation.py
```diff
@@ -204,6 +236,12 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
             path_suffix (str): Suffix to be added to new augmented files.
             resource_loader (object): Resource Loader object for the application.
         """
+
+        if language != 'en':
```
KKsharma99, line 240: Nit: Let's use the ENGLISH_LANGUAGE_CODE global variable from components._config instead of 'en'

mindmeld/augmentation.py
```diff
@@ -241,7 +302,160 @@ def __init__(self, batch_size, language, paths, path_suffix, resource_loader):
...
+                                      format required by the models
+            processed_queries (list(ProcessedQuery)): List of ProcessedQuery
+                                                       with entity annotations
+        """
+        model_inputs = []
+        for processed_query in processed_queries:
+            processed_query_text = processed_query.query.text.strip()
+            if self.retain_entities:
+                text = [processed_query_text.lower(), EOS_TOKEN]
+                for entity in processed_query.entities:
```
vijay120, line 347: does it matter what entity comes first (I assume we sort by span start index)?

vijay120 (approved):

mindmeld/augmentation.py
vembar: I sort the list when replacing the entities

PR #374, opened 2021-10-06
Feature/duckling 20211005

serapio:
Update to new Duckling binary, based on Duckling 2.0.0


tests/test_ser.py
```diff
@@ -192,6 +228,48 @@ def test_phone_number(query, predicted_texts, predicted_values):
...
+        ("card number 6011-1111-1111-1117", ["6011-1111-1111-1117"], ["6011111111111117"]),
+        ("1899028724221", ["1899028724221"], ["1899028724221"]),
+    ],
+)
+def test_credit_card(query, predicted_texts, predicted_values):
+    data = {"text": query, "reftime": NOW_TIMESTAMP, "latent": True}
+
+    res = requests.post(DUCKLING_URL, data=data)
+    responses = res.json()
+    print(responses)
```
alhuang10, line 254: Remove?

jre21 (approved):

mindmeld/system_entity_recognizer.py
```diff
@@ -55,6 +55,7 @@ class SystemEntityError(Exception):
 
 class DucklingDimension(Enum):
     AMOUNT_OF_MONEY = "amount-of-money"
+    CREDIT_CARD_NUMBER = "credit-card-number"
```
KKsharma99, line 58: If Duckling can now recognize credit card numbers, we can add this functionality to the auto-annotator as well by adding "sys_credit_card_number" to the DUCKLING_TO_SYS_ENTITY_MAPPINGS for the languages that it is supported for. However, since Spacy doesn't recognize this entity, it would need to be used in the setting where the largest duckling candidate spans are labelled. May be a feature with a limited use case at this time, however, it could be potentially be supported in future PR.

Issue #378, opened 2021-10-18
Mindmeld Docker Containers Currently Unrunnable


Zozman:
Currently, the docker containers provided in the `docker_containers` directory of this repo are unrunnable.  When attempting to run them, the following errors occur:

```
warning: en_core_web_sm not found on disk. Downloading the model.
/usr/bin/python: No module named spacy
```

In an attempt to rectify this, I attempted to install both `spaCy` as well as the `en_core_web_sm` by adding the following to [here in the `Dockerfile`](https://github.com/cisco/mindmeld/blob/master/docker_containers/mindmeld_docker/Dockerfile#L69) to install these:

``` Dockerfile
# Install spaCy and the proper pipeline used by mindmeld for this language
RUN pip3 install -U spacy
RUN export LC_ALL=C.UTF-8 && \
    export LANG=C.UTF-8 && \
    python3 -m spacy download en_core_web_sm
```

This solves the original issue but then the following is seen when running the container:

```
KeyError: "[E002] Can't find factory for 'tok2vec'. This usually happens when spaCy calls `nlp.create_pipe` with a component name that's not built in - for example, when constructing the pipeline from a model's meta.json. If you're using a custom component, you can write to `Language.factories['tok2vec']` or remove it from the model meta and add it via `nlp.add_pipe` instead."
```

From here it seems that `mindmeld` is downgrading `spacy` when it's being installed so I think the required version at https://github.com/cisco/mindmeld/blob/b644c6f843bf96874ef816e0f537eb8ae0862660/setup.py#L39 (and then maybe all you would need is to install the correct pipeline).

This is officially over my head so therefore filing this issue.

Steps to reproduce:

1. Go to `docker_containers/mindmeld_docker` directory of this repo.
2. Run `docker-compose build --no-cache`.

Zozman:
Provided a possible fix at #380 but not sure if this is the better way to fix it or changing what version of `spacy` is installed by mindmeld.

PR #380, opened 2021-10-19
Forced Dockerfiles to use spacy 3.2.1 to fix Docker build issue with Mimdmeld version 4.4.0

Zozman:
As mentioned in #378, currently the Docker container builds fail because the version of `spacy` installed by mindmeld as a dependency is too old to use the `en_core_web_sm` model currently provided by `spacy`.  This change forces the `spacy` version to be upgraded.

Not sure if this is the best solution to the issue or if updating the required version of `spacy` at https://github.com/cisco/mindmeld/blob/b644c6f843bf96874ef816e0f537eb8ae0862660/setup.py#L39 is better but providing this as a PR in case it helps.


Zozman:
Updated to latest `spacy` in case this fix is still needed; currently still can't build Dockerfiles without this.

PR #383, opened 2021-11-16
Active Learning for Entity Recognition

ritvikshrivastava:
Major changes from a user perspective:

instead of tuning strategies, we now provide classifier tuning and tagger tuning as two independent modes. Both of them can be run independently or at the same time. Strategies for them can be defined in ``classifier_tuning_strategies`` and ``tagger_tuning_strategies``.  The modified heuristics for tagger/entity strategies are based on the following paper: https://dl.acm.org/doi/pdf/10.5555/1613715.1613855

To select tuning type, ``tuning_level`` has been updated to a list. Options: ``["domain", "intent", "entity"]``. If both domain and intent are provided, intent runs for classifier tuning. Depending on this list, classifier and/or tagger tuning can be run.

```
DEFAULT_ACTIVE_LEARNING_CONFIG = {
    "output_folder": None,
    "pre_tuning": {
        "train_pattern": ".*train.*.txt",
        "test_pattern": ".*test.*.txt",
        "train_seed_pct": 0.20,
    },
    "tuning": {
        "n_classifiers": 3,
        "n_epochs": 5,
        "batch_size": 100,
        "tuning_level": ["domain"],
        "entity_tuning": False,
        "classifier_tuning_strategies": [
            "LeastConfidenceSampling",
            "MarginSampling",
            "EntropySampling",
            "RandomSampling",
            "DisagreementSampling",
            "EnsembleSampling",
            "KLDivergenceSampling",
        ],
        "tagger_tuning_strategies": [
            "LeastConfidenceSampling",
            "MarginSampling",
            "EntropySampling",
        ],
    },
    "tuning_output": {
        "save_sampled_queries": True,
        "aggregate_statistic": "accuracy",
        "class_level_statistic": "f_beta",
    },
    "query_selection": {
        "classifier_selection_strategy": "EntropySampling",
        "tagger_selection_strategy": "EntropySampling",
        "log_usage_pct": 1.00,
        "labeled_logs_pattern": None,
        "unlabeled_logs_path": "logs.txt",
    },
}
```


mindmeld/active_learning/classifiers.py
```diff
@@ -168,9 +172,99 @@ def _validate_class_level_statistic(class_level_statistic):
...
+                for _ in range(len(queries)):
+                    query_prob_vector_2d = np.zeros((1, len(tag_to_id)))
+
+                    query_prob_vector_2d[0][default_idx] = default_prob
+                    queries_prob_vectors.append(query_prob_vector_2d)
+
+            else:
+                for query in classifier_eval.results:
+
+                    if not len(query.predicted):
```
vembar, line 234: Can you add a note as to what the difference is between the different cases?

mindmeld/active_learning/classifiers.py
```diff
@@ -235,17 +337,31 @@ def train(self, data_bucket: DataBucket, heuristic: Heuristic):
...
         eval_stats = defaultdict(dict)
         eval_stats["num_sampled"] = len(data_bucket.sampled_queries)
-        confidences_2d = self.train_single(data_bucket, eval_stats)
+        confidences_2d, entity_confidences, eval_stats = self.train_single(
+            data_bucket, heuristic, eval_stats
+        )
         return_confidences_3d = isinstance(heuristic, MULTI_MODEL_HEURISTICS)
-        confidences_3d = (
-            self.train_multi(data_bucket) if return_confidences_3d else None
+        confidences_3d, entity_confidences_3d = (
```
vembar, line 344: We should probably make this change for when we use the 3d confidences. Seems like we're not using it at the moment.

mindmeld/active_learning/classifiers.py
```diff
@@ -235,17 +337,31 @@ def train(self, data_bucket: DataBucket, heuristic: Heuristic):
...
+            domain_indices,
+        )
 
-    def train_single(self, data_bucket: DataBucket, eval_stats: defaultdict = None):
+    def train_single(
+        self,
+        data_bucket: DataBucket,
+        heuristic: Heuristic,
+        eval_stats: defaultdict = None,
+    ):
```
vembar, line 364: doc string needs to be updated

mindmeld/active_learning/classifiers.py
```diff
@@ -531,3 +680,122 @@ def _update_eval_stats_intent_level(
...
+                    heuristic=heuristic,
+                )
+
+                for i, index in enumerate(filtered_unsampled_queries_indices):
+                    unsampled_idx_preds_pairs[index] = er_queries_prob_vectors[i]
+
+        indices = list(unsampled_idx_preds_pairs.keys())
+        indices.sort()
+        er_queries_prob_vectors = [unsampled_idx_preds_pairs[index] for index in indices]
+        # import pdb; pdb.set_trace()
```
vembar, line 760: Remove this line

mindmeld/active_learning/classifiers.py
```diff
@@ -531,3 +680,122 @@ def _update_eval_stats_intent_level(
...
+                er = self.nlp.domains[domain].intents[intent].entity_recognizer
+                try:
+                    er.fit(queries=filtered_sampled_queries)
+                except:
+                    pass
+                # Evaluate Test Queries
+                er_eval_test = er.evaluate(queries=filtered_test_queries)
+                er_eval_test_dict[f"{domain}.{intent}"] = er_eval_test
+
+                # Get Probability Vectors
```
vembar, line 745: Why are we setting probability vectors for unsampled queries here?

mindmeld/active_learning/classifiers.py
```diff
@@ -168,9 +172,99 @@ def _validate_class_level_statistic(class_level_statistic):
             )
         return class_level_statistic
 
+    @staticmethod
+    def _get_entity_vector(
```
vembar, line 176: Is this method used anywhere?

mindmeld/models/tagger_models.py
```diff
@@ -315,20 +315,24 @@ def predict_proba(self, examples, dynamic_resource=None):
...
-        for entity in entities:
-            entity_proba = \
-                probas[entity.normalized_token_span.start: entity.normalized_token_span.end + 1]
-            # We assume that the score of the least likely tag in the sequence as the confidence
-            # score of the entire entity sequence
-            entity_confidence.append(min(entity_proba))
-        predicted_labels_scores = tuple(zip(entities, entity_confidence))
-        return predicted_labels_scores
+        # tags, probas = zip(*predicted_tags_probas[0])
+        # import pdb; pdb.set_trace()
```
vembar, line 322: remove this line

vembar:
Did a quick initial pass. Lot of commented lines of code - some for debugging and some maybe for future use? Can you please remove the unnecessary ones and write a one liner description for the necessary ones?

mindmeld/active_learning/classifiers.py
ritvikshrivastava: We're evaluating on unsampled queries given the model trained on the sampled queries so far. These probabilities/confidences are used for active learning selection

mindmeld/active_learning/classifiers.py
ritvikshrivastava: Removed (has been moved to label map class)

mindmeld/active_learning/alp.py
```diff
@@ -311,6 +328,7 @@ def create_from_config(config):
             n_epochs=config.get("tuning", {}).get("n_epochs"),
             batch_size=config.get("tuning", {}).get("batch_size"),
             tuning_strategies=config.get("tuning", {}).get("tuning_strategies"),
+            entity_tuning=config.get("tuning", {}).get("entity_tuning", True),
```
KKsharma99, line 331: Should entity_tuning be True by default?

mindmeld/active_learning/classifiers.py
```diff
@@ -100,6 +101,7 @@ def __init__(
         self,
         app_path: str,
         tuning_level: str,
+        entity_tuning: str,
```
KKsharma99, line 104: Perhaps this should be bool instead of str, as it is a bool in the __init__() method?

mindmeld/active_learning/classifiers.py
```diff
@@ -168,9 +172,70 @@ def _validate_class_level_statistic(class_level_statistic):
...
+            default_idx = tag_to_id[f"{default_key}"]
+
+            if not classifier_eval:
+                for _ in range(len(queries)):
+                    query_prob_vector_2d = np.zeros((1, len(tag_to_id)))
+
+                    query_prob_vector_2d[0][default_idx] = default_prob
+                    queries_prob_vectors.append(query_prob_vector_2d)
+
+            else:
```
KKsharma99, line 210: To reduce indentation, we can remove the "else" condition here and add "return queries_prob_vectors" within the if block

mindmeld/active_learning/classifiers.py
```diff
@@ -168,9 +172,70 @@ def _validate_class_level_statistic(class_level_statistic):
...
+            classifier (MindMeld Classifer): Domain or Intent Classifier
+            queries (ProcessedQueryList): List of MindMeld queries
+            tag_to_id (Dict): Dictionary mapping domain or intent names to vector index
+                positions.
+
+        Returns:
+            prob_vector (List[List]]): Probability distribution vectors for given queries.
+        """
+        queries_prob_vectors = []
+        if queries:
```
KKsharma99, line 193: To reduce indentation, we can change this to:

```
if not queries:
       return queries_prob_vectors
# Rest of the code
```

mindmeld/active_learning/classifiers.py
```diff
@@ -168,9 +172,70 @@ def _validate_class_level_statistic(class_level_statistic):
             )
         return class_level_statistic
 
+    @staticmethod
+    def _get_entity_probs(
+        classifier: Classifier,
+        queries: ProcessedQueryList,
+        tag_to_id: Dict,
+    ):
+        """Get the probability distribution for a query across domains or intents.
```
KKsharma99, line 181: Let's update the description here to explicitly explain how the entity probabilities are calculates and what the rows/columns of the final output represent. An small example here would help

mindmeld/active_learning/classifiers.py
```diff
@@ -168,9 +172,70 @@ def _validate_class_level_statistic(class_level_statistic):
...
+
     @staticmethod
     def _get_probs(
-        classifier: Classifier, queries: ProcessedQueryList, nlp_component_to_id: Dict
+        classifier: Classifier,
+        queries: ProcessedQueryList,
+        nlp_component_to_id: Dict,
+        nlp_component_type=None,
     ):
         """Get the probability distribution for a query across domains or intents.
```
KKsharma99, line 240: Nit: Let's add entities to the description as well

mindmeld/active_learning/classifiers.py
```diff
@@ -183,6 +248,13 @@ def _get_probs(
         Returns:
             prob_vector (List[List]]): Probability distribution vectors for given queries.
         """
+        if nlp_component_type == "entity":
```
KKsharma99, line 251: Since we're using nlp_component_type as a param for this function it would be more consistent to also add another type "classifier", this can be the default for the function

mindmeld/active_learning/classifiers.py
```diff
@@ -183,6 +248,13 @@ def _get_probs(
...
+            return MindMeldALClassifier._get_entity_probs(
+                classifier=classifier,
+                queries=queries,
+                tag_to_id=nlp_component_to_id,
+            )
+
         queries_prob_vectors = []
```
KKsharma99, line 258: Consider placing the code below in a separate helper function as well "_get_classifier_probs()"

mindmeld/models/text_models.py
```diff
@@ -386,7 +386,7 @@ def _get_feature_scaler(self):
         }.get(scale_type)
         return scaler
 
-    def evaluate(self, examples, labels):
+    def evaluate(self, examples, labels, active_learning=False):
```
KKsharma99, line 389: Let's rename this parameter to focus on the updated return type instead of a single application it could be used for. This could be "ret_prob_full" or "ret_prob_all" or anything else you see fit. In this way, this function can easily/intuitively be used for applications beyond AL.

mindmeld/models/taggers/memm.py
```diff
@@ -154,6 +154,32 @@ def _predict_proba_example(self, example, config, resources):
             seq_log_probs.append([prev_tag, prediction[predicted_tag]])
         return seq_log_probs
 
+    def predict_proba_active_learning(self, examples, config, resources):
```
KKsharma99, line 157: We can update the use of "active_learning" in the function name here as well. If the new function name is say predict_proba_all(), we may want to consider adding this to the base class and implementing it for the other classes. Even if predict_proba_all() simply returns predict_proba() across other subclasses it would make for a more consistent application interface. Ultimately, we may want to refactor such that predict_proba() returns the same thing as predict_proba_all()

mindmeld/active_learning/classifiers.py
```diff
@@ -168,9 +172,70 @@ def _validate_class_level_statistic(class_level_statistic):
...
+        Returns:
+            prob_vector (List[List]]): Probability distribution vectors for given queries.
+        """
+        queries_prob_vectors = []
+        if queries:
+            classifier_eval = classifier.evaluate(queries=queries, active_learning=True)
+
+            domain = classifier.domain
+            intent = classifier.intent
+            default_prob = 1.0
```
KKsharma99, line 198: Consider adding a comment here explaining why we default to 100% confidence if entities don't exist as it isn't immediately intuitive

mindmeld/active_learning/classifiers.py
```diff
@@ -168,9 +172,70 @@ def _validate_class_level_statistic(class_level_statistic):
...
+            default_tag = "O|"
+            default_key = f"{domain}.{intent}.{default_tag}"
+            default_idx = tag_to_id[f"{default_key}"]
+
+            if not classifier_eval:
```
KKsharma99, line 203: We can either add a comment here explaining what "not classifier_eval" means and why it happens or include this descriptive variable name (or both).

```
intent_lacks_entities = not classifier:
if intent_lacks_entities:
      # Code here
```

mindmeld/active_learning/classifiers.py
```diff
@@ -168,9 +172,70 @@ def _validate_class_level_statistic(class_level_statistic):
...
+        queries_prob_vectors = []
+        if queries:
+            classifier_eval = classifier.evaluate(queries=queries, active_learning=True)
+
+            domain = classifier.domain
+            intent = classifier.intent
+            default_prob = 1.0
+            default_tag = "O|"
+            default_key = f"{domain}.{intent}.{default_tag}"
+            default_idx = tag_to_id[f"{default_key}"]
```
KKsharma99, line 201: To clearly distinguish this mapping from that for classifiers lets rename this to `entity_tag_to_id`

mindmeld/active_learning/classifiers.py
```diff
@@ -531,3 +647,118 @@ def _update_eval_stats_intent_level(
...
+        Args:
+            eval_stats (defaultdict): Evaluation metrics to be included in accuracies.json.
+            er_eval_test_dict (Dict): Dictionary mapping a domain.intent (str) to the
+                associated er_eval_test object.
+        """
+        for domain_intent, er_eval_test in er_eval_test_dict.items():
+            domain, intent = domain_intent.split(".")
+
+            if er_eval_test:
+                if domain not in eval_stats["accuracies"]:
```
KKsharma99, line 743: Let's move these two if checks into the plotting file itself

mindmeld/active_learning/classifiers.py
```diff
@@ -531,3 +647,118 @@ def _update_eval_stats_intent_level(
...
+                if intent not in eval_stats["accuracies"][domain]:
+                    eval_stats["accuracies"][domain].update({intent: {}})
+
+                eval_stats["accuracies"][domain][intent]["entities"] = {
+                    "overall": er_eval_test.get_stats()["stats_overall"][
+                        self.aggregate_statistic
+                    ]
+                }
+
+                if verbose:
```
KKsharma99, line 754: A comment here explaining that "verbose" gives you B,I,O tags would be helpful

mindmeld/active_learning/data_loading.py
```diff
@@ -300,13 +323,26 @@ def sample_and_update(
...
+
+        newly_sampled_indices = ranked_indices_2d[:sampling_size]
+        remaining_indices = ranked_indices_2d[sampling_size:]
+
+        if entity_tuning:
+            ranked_entity_indices = heuristic.rank_entities(entity_confidences)
+            newly_sampled_indices_entity = ranked_entity_indices[:sampling_size]
+            remaining_indices_entity = ranked_entity_indices[sampling_size:]
+
+            newly_sampled_indices = list(
```
KKsharma99, line 340: By taking the union here we would get a variable batch_size between 100 (all selections overlap) to 200 (no selections overlap). In the config, we have the defined param of "batch_size" and the user would expect this value to be used. Moreover, a consistent batch_size has been important for post-processing scripts. The solution here can be to take a random sample of the union data such that only batch_size num queries are selected.

mindmeld/active_learning/data_loading.py
```diff
@@ -316,21 +352,25 @@ def sample_and_update(
...
         Returns:
             filtered_queries_indices (list): List of indices of filtered queries.
             filtered_queries (list): List of filtered queries.
         """
+
         filtered_queries = []
         filtered_queries_indices = []
         for index, query in enumerate(query_list.processed_queries()):
-            if query.domain == domain:
+            if getattr(query, component_type) == component_name:
```
KKsharma99, line 373: Clever use of getattr()! :) Consider throwing an error as an "else" condition if the component type is not found in the query.

mindmeld/active_learning/alp.py
```diff
@@ -44,6 +45,7 @@ def __init__(  # pylint: disable=R0913
         batch_size: int,
         tuning_strategies: list,
         tuning_level: str,
+        entity_tuning: bool,
```
vembar, line 48: Just to clarify - we only have a single strategy for entities (irrespective of the model) for now?

ritvikshrivastava: No, we still have three strategies for entities. LCS, Entropy and Margin.

mindmeld/active_learning/classifiers.py
ritvikshrivastava: What would this additional type be used for?

mindmeld/models/taggers/memm.py
ritvikshrivastava: updated the flag from active_learning to fetch_distribution and predict_proba_active_learning to predict_proba_distribution

mindmeld/active_learning/classifiers.py
ritvikshrivastava: I checked this, without these checks the line below can't populate this dictionary with the aggregate stats. We would need these or an equivalent here.

mindmeld/active_learning/classifiers.py
ritvikshrivastava: Added equivalent methods for classifier and tagger.

mindmeld/active_learning/alp.py
```diff
@@ -104,6 +114,22 @@ def __init__(  # pylint: disable=R0913
...
+
+        for level in self.tuning_level:
+            if level not in [TUNE_LEVEL_DOMAIN, TUNE_LEVEL_INTENT, TUNE_LEVEL_ENTITY]:
+                raise ValueError(f"Invalid tuning level: {level}")
+
+        if (
+            TUNE_LEVEL_DOMAIN in self.tuning_level
+            and TUNE_LEVEL_INTENT in self.tuning_level
+        ):
+            logger.info(
```
KKsharma99, line 128: Nit: Perhaps we should raise an error here instead of silently selecting intent on behalf of the user.  A log may go undetected in the terminal. Since AL can take a long time to train, it would be ideal if the user is clear on what training level is selected, especially if they were aiming to train at the domain-level and not at the intent-level.

mindmeld/active_learning/alp.py
```diff
@@ -179,32 +210,61 @@ def select_queries(self):
...
-            strategy=self.selection_strategy, select_mode=True
-        )
-        self.results_manager.write_log_selected_queries_json(
-            strategy=self.selection_strategy,
-            queries=newly_sampled_queries,
-        )
+
+        if self.classifier_selection_strategy:
+            newly_sampled_queries = self._run_strategy(
+                tuning_type="classifier",
```
KKsharma99, line 216: Let's make "classifier" and "tagger" into constants or enum.

mindmeld/active_learning/classifiers.py
```diff
@@ -531,3 +678,120 @@ def _update_eval_stats_intent_level(
...
+                for unsampled queries.
+            ic_eval_test_dict (Dict): Dictionary mapping a domain (str) to the
+                associated ic_eval_test object.
+        """
+        er_eval_test_dict = {}
+        unsampled_idx_preds_pairs = {}
+        for domain, intents in domain_to_intents.items():
+            for intent in intents:
+                # Filter Queries
+                (
```
KKsharma99, line 709: Nit: Is there a way to make these similar calls with a for loop to reduce some verbosity/repetition?

mindmeld/active_learning/classifiers.py
```diff
@@ -531,3 +678,120 @@ def _update_eval_stats_intent_level(
...
+                    query_list=test_queries,
+                    component_type=TUNE_LEVEL_INTENT,
+                    component_name=intent,
+                )
+                # Train
+                er = self.nlp.domains[domain].intents[intent].entity_recognizer
+                try:
+                    er.fit(queries=filtered_sampled_queries)
+                except ValueError:
+                    # single class, cannot fit with solver
```
KKsharma99, line 735: Nit: It may be a good idea here to log a message capturing what your comment describes. For example, "Skipping {class}, cannot fit with solver." However, we should only do this if we're certain that a ValueError will always correspond to this. If not, maybe there is another way to find out? (Perhaps you can use the LabelMap for this)

mindmeld/active_learning/plot_manager.py
```diff
@@ -449,6 +434,71 @@ def plot_all_epochs(
...
+                        epoch_dict, ["num_sampled"]
+                    )
+                    y_values = PlotManager.get_across_iterations(
+                        epoch_dict, ["accuracies"] + y_keys
+                    )
+                    self.plt.plot(x_values, y_values)
+                    all_y_values.append(y_values)
+                max_len = max([len(i) for i in all_y_values])
+                for y_values in all_y_values:
+                    y_values = np.pad(
```
KKsharma99, line 471: Nit: From my understanding, this is padding by extending the last value. A comment explaining that may be helpful here

mindmeld/active_learning/plot_manager.py
```diff
@@ -494,8 +544,8 @@ def plot_stacked_bar(
             plot_domain (bool): Whether to generate plots at the domain level.
             plot_intents (bool): Whether to generate plots at the intent level.
         """
-        for strategy in self.queries_data:
-            epoch_dict = self.queries_data[strategy][str(epoch)]
+        for strategy in self.strategies[0]:
```
KKsharma99, line 547: When we're accessing the first index value here, is it guaranteed to exist?

mindmeld/active_learning/heuristics.py
```diff
@@ -241,6 +258,53 @@ def rank_3d(confidences_3d: List[List[List[float]]]) -> List[int]:
         all_ordered_sample_indices = [MarginSampling.rank_2d(c) for c in confidences_3d]
         return Heuristic.ordered_indices_list_to_final_rank(all_ordered_sample_indices)
 
+    @staticmethod
+    def beam_search_decoder(predictions, top_k=3):
```
KKsharma99, line 262: Would be helpful to add some tests for this

mindmeld/active_learning/heuristics.py
```diff
@@ -241,6 +258,53 @@ def rank_3d(confidences_3d: List[List[List[float]]]) -> List[int]:
...
+                new_sequences, key=lambda val: val[1], reverse=True
+            )
+
+            # select top-k based on score
+            output_sequences = output_sequences[:top_k]
+
+        return output_sequences
+
+    @staticmethod
+    def rank_entities(entity_confidences: List[List[List[float]]]) -> List[int]:
```
KKsharma99, line 292: A description here explaining the method by which entities are being ranked and where this is normally used would help.

mindmeld/models/taggers/memm.py
```diff
@@ -154,6 +154,32 @@ def _predict_proba_example(self, example, config, resources):
             seq_log_probs.append([prev_tag, prediction[predicted_tag]])
         return seq_log_probs
 
+    def predict_proba_distribution(self, examples, config, resources):
```
KKsharma99, line 157: Let's add tests for these new memm.py functions

mindmeld/active_learning/plot_manager.py
ritvikshrivastava: Yup, added checks to ensure that!

mindmeld/active_learning/alp.py
ritvikshrivastava: Instead of raising an error, how about creating a warning? While it makes sense that users notice this automatic preference, we should also consider minimal breaking points

mindmeld/active_learning/classifiers.py
ritvikshrivastava: It might become difficult to follow later, and the outputs are independent each time so variables will need to be added on regardless.

mindmeld/active_learning/data_loading.py
```diff
@@ -300,49 +304,59 @@ def sample_and_update(
...
+                ranked_entity_indices = heuristic.rank_entities(confidences_2d)
+            except (TypeError, ValueError):
+                # if heuristic does not have entity AL support default to entropy
+                heuristic = EntropySampling
+                ranked_entity_indices = heuristic.rank_entities(confidences_2d)
+
+            newly_sampled_indices = ranked_entity_indices[:sampling_size]
+            remaining_indices = ranked_entity_indices[sampling_size:]
+
+            # PLACEHOLDER CODE for joint tuning (deprecated for now).
```
vembar, line 352: I think adding an additional script that reads in the json files created for intent/domain and entity for a run and generate a plot showing trends in overall app accuracy resolves this concern. Mostly because as discussed with @ritvikshrivastava, if multiple strategies are selected for each level, we've have to run this eval for all combinations.   

If @KKsharma99, @serapio are in agreement with this, let's remove this commented block of code.

mindmeld/components/_config.py
```diff
@@ -490,9 +490,9 @@
         "n_classifiers": 3,
         "n_epochs": 5,
         "batch_size": 100,
-        "tuning_level": "domain",
+        "tuning_level": ["domain"],
         "entity_tuning": False,
```
vembar, line 494: We don't need this key anymore

vembar (approved):
Some very minor changes needed. Otherwise, LGTM!

PR #385, opened 2021-11-18
Remove Import Error Warning in text_preparation_pipeline.py in create_from_app_path()

KKsharma99:
A user can specify a custom TextPreparationPipeline in the app.py module (at app root level). TextPreparationPipelineFactory.create_from_app_path(app_path) checks if app.py file has a custom TextPrepPipeline. Currently, if a custom pipeline is not found in app.py it will give a warning saying the default pipeline is being used. However, the config.py is checked next for a text_preparation_pipeline_config. Only if a custom config is not found in config.py is the default pipeline used. The warning message for when this happens is already handled in TextPreparationPipelineFactory.create_from_app_config(app_path) so having it in the create_from_app_path() function is inaccurate/redundant.


serapio (approved):
I was scratching my head about that

vijay120 (approved):

PR #386, opened 2021-12-20
Update Elasticsearch dependency.

tmehlinger:
This updates our Elasticsearch dependency in the dependency containers
to mitigate https://www.cve.org/CVERecord?id=CVE-2021-44228.

### DIFF
--- a/docker_containers/mindmeld_dep_docker/Dockerfile
+++ b/docker_containers/mindmeld_dep_docker/Dockerfile
@@ -10,8 +10,8 @@
 # Pull base image.
 FROM ubuntu:18.04
 
-ENV ES_PKG_NAME elasticsearch-7.8.0-linux-x86_64
-ENV ES_FOLDER_NAME elasticsearch-7.8.0
+ENV ES_PKG_NAME elasticsearch-7.16.2-linux-x86_64
+ENV ES_FOLDER_NAME elasticsearch-7.16.2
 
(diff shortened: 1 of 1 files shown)


Zozman:
The same Docker update needs to be done on https://github.com/cisco/mindmeld/blob/1ff40df9863d627b16031f5897e28786ac83d06b/docker_containers/mindmeld_docker/Dockerfile#L13-L14 as well, right?

serapio:
We should fix references in the docs too https://github.com/cisco/mindmeld/blob/master/source/userguide/getting_started.rst#1-install-prerequisites

tmehlinger:
@serapio Agree, we should put a note there about pulling down the latest. However, we also need to update the `mindmeldworkbench/dep` image referred to there. Could someone point me to where/how we push updates to that image?

tmehlinger:
Presently failing because I updated the code to use the latest ES releases... which don't have corresponding Docker images yet. I'll check again later.

vrdn-23:
@tmehlinger The docker builds seem to be available now. 
https://www.docker.elastic.co/r/elasticsearch
Do you want me to take a look at this issue?

Zozman:
> Man, I've been living under a rock. I'd be thrilled if you look into this. :D

Saw this and actually have some work on it so went ahead and filed #425 if y'all are interested.

PR #387, opened 2022-01-11
Incremental build misses cache files and unit test

vrdn-23:
Hey folks,

This PR is meant to fix the issue described in #388.

## The Reason
The preliminary reason for this bug is because once we have a model cache that does not have any data changes, the entity recognizer and the role classifier classes no longer save/update the entity types and roles as their children. Hence when doing an incremental build, these files do not have an incremental build path and do not get copied to the new time stamp.

## The Fix
The fix for this was pretty simple:
We move the parts where the entity types and role labels are getting defined to before the model cache is being loaded.

## Other changes

- I've also added a unit test that verifies the intended behavior. 
- I've updated a unit test in test_role_classifier because I'm not really sure how that was passing all this while. It makes sense that we would need to dump the model files after building them before trying to load files, but the original test case as it was written seems to pass locally 3/5 times for me. The current fix which just ensures an extra dump in the test, passes 5/5 times for me, and I believe this to be the intended behavior. Please do let me know if there is something that I am missing here!

P.S. - Thanks to @davidmfrey and @murali1996 for the debugging sessions

### DIFF
--- a/mindmeld/components/role_classifier.py
+++ b/mindmeld/components/role_classifier.py
@@ -111,8 +111,7 @@ class RoleClassifier(Classifier):
 
         if examples:
             # Build roles set
-            for label in labels:
-                self.roles.add(label)
+            self.roles.update(labels)
 
         if incremental_timestamp and cached_model_path:
             logger.info("No need to fit. Previous model is cached.")
--- a/mindmeld/components/role_classifier.py
+++ b/mindmeld/components/role_classifier.py
@@ -103,7 +103,7 @@ class RoleClassifier(Classifier):
         queries = self._resolve_queries(queries, label_set)
 
         new_hash = self._get_model_hash(model_config, queries)
-        cached_model = self._resource_loader.hash_to_model_path.get(new_hash)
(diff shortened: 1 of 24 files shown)


mindmeld/components/role_classifier.py
```diff
@@ -105,24 +105,24 @@ def fit(self,
...
-            return False
-
         # These examples and labels are flat lists, not
         # a ProcessedQueryList.Iterator
         examples, labels = self._get_examples_and_labels(queries)
 
         if examples:
             # Build roles set
-            self.roles = set()
             for label in labels:
```
davidmfrey, line 114: This operation is more efficient if you simply use `self.roles.update(labels)` rather than the for loop.

davidmfrey (approved):

mindmeld/components/role_classifier.py
vrdn-23: Fixed!

tmehlinger (approved):
LGTM :)

PR #391, opened 2022-01-18
Make few modifications to how resource loader is initialized in NativeQA

murali1996:
In the [question_answerer.py](https://github.com/cisco/mindmeld/blob/master/mindmeld/components/question_answerer.py), a ResourceLoader object is required for building models in NativeQA while it is not required for ElasticsearchQA.

Before this PR, the class variable `RESOURCE_LOADER` of NativeQA is always initialized to a newly created resource loader object. However, it is possible that a question-answerer can be called within an app by passing-in an already created/available resource loader. Through this PR, such scenario is accounted for.

### DIFF
--- a/mindmeld/components/question_answerer.py
+++ b/mindmeld/components/question_answerer.py
@@ -35,6 +35,7 @@ from nltk.corpus import stopwords as nltk_stopwords
 from nltk.stem import PorterStemmer
 from nltk.tokenize import word_tokenize as nltk_word_tokenize
 
+from mindmeld.query_factory import QueryFactory
 from ._config import (
     get_app_namespace,
     get_classifier_config,
@@ -372,13 +373,35 @@ class NativeQuestionAnswerer(BaseQuestionAnswerer):
     """
 
     # a common resource loaded generally used across all resolvers during loading process
-    RESOURCE_LOADER = ResourceLoader.create_resource_loader(
-        app_path=None,
-        text_preparation_pipeline=TextPreparationPipelineFactory.create_text_preparation_pipeline(
-            language="en",
-            tokenizer=WhiteSpaceTokenizer()
-        )
-    )
+    # a class var because it is also used in the underlying `Indices` class
+    RESOURCE_LOADER = None
+
+    def __init__(self, *args, **kwargs):
+        """
+        Args:
+            resource_loader (ResourceLoader, optional): A resource loader object used by the
+                underlying entity resolver models
+        """
+        super().__init__(*args, **kwargs)
+
+        # update class' resource loader; use one if already passed-in
+        resource_loader = kwargs.get("resource_loader")
+        if not resource_loader:
+            text_preparation_pipeline = \
+                TextPreparationPipelineFactory.create_text_preparation_pipeline(
+                    language="en",
+                    tokenizer=WhiteSpaceTokenizer()
+                )
+            query_factory = QueryFactory.create_query_factory(
+                app_path=None,
+                text_preparation_pipeline=text_preparation_pipeline,
+                duckling=True
+            )
+            resource_loader = ResourceLoader.create_resource_loader(
+                app_path=None,
+                query_factory=query_factory
+            )
+        NativeQuestionAnswerer.RESOURCE_LOADER = resource_loader
 
     def _load_kb(self,
                  index_name,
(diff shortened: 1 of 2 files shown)


mindmeld/components/question_answerer.py
```diff
@@ -35,6 +35,7 @@
 from nltk.stem import PorterStemmer
 from nltk.tokenize import word_tokenize as nltk_word_tokenize
 
+from mindmeld.query_factory import QueryFactory
```
ritvikshrivastava, line 38: can we make this a relative import?

mindmeld/components/question_answerer.py
```diff
@@ -372,13 +373,35 @@ class NativeQuestionAnswerer(BaseQuestionAnswerer):
...
+        if not resource_loader:
+            text_preparation_pipeline = \
+                TextPreparationPipelineFactory.create_text_preparation_pipeline(
+                    language="en",
+                    tokenizer=WhiteSpaceTokenizer()
+                )
+            query_factory = QueryFactory.create_query_factory(
+                app_path=None,
+                text_preparation_pipeline=text_preparation_pipeline,
+                duckling=True
```
murali1996, line 398: @vrdn-23 @ritvikshrivastava  I think without explicitly setting `duckling=True`, the error message you see will still keep coming up if no resource_loaded is passed to NativeQA. Hence, instead of passing in a `text_preparation_pipeline` to `create_resource_loader`, I am passing a `query_factory` object directly.

mindmeld/components/question_answerer.py
```diff
@@ -372,13 +373,35 @@ class NativeQuestionAnswerer(BaseQuestionAnswerer):
...
+            )
+            resource_loader = ResourceLoader.create_resource_loader(
+                app_path=None,
+                query_factory=query_factory
+            )
+        NativeQuestionAnswerer.RESOURCE_LOADER = resource_loader
```
ritvikshrivastava, line 404: Curious, why is this variable static instead of instance?

mindmeld/components/question_answerer.py
ritvikshrivastava: Yeah this is fine I think (otherwise we'll have to trickle this change down via resource_loader all the way to query factory)

ritvikshrivastava (approved):
Looks good! (left a couple of comments to consider)

mindmeld/components/question_answerer.py
vrdn-23: Curious as to what happens in the case when we aren't using duckling? In a system that doesn't have the duckling option/binary, what would happen by this statement?

mindmeld/components/question_answerer.py
murali1996: sure, can do that

mindmeld/components/question_answerer.py
murali1996: Because it is also used in classes defined within NativeQuestionAnswerer-- Search and Indices. (for e.g. a resource loader is also required when loading dumped NativeQA's entity resolver objects. Since there are no .dump() and .load() methods for question-answerers, dumping and loading  is done implicitly when load_kb is called which in turn calls a method in Indices class. also, changing this to instance var also needs changes to execute() method of Search)

mindmeld/components/question_answerer.py
murali1996: Good question. Made me think we shouldn't actually hard set duckling=True here. It is because QA component in MindMeld doesn't use "query" objects anywhere and hence no need a duckling recognizer.

vrdn-23: The previous behavior was creating a NoOpSystemEntityRecognizer. This change in flag kind of changes that.  
We should try to identify if there is a way to suppress that warning given in query_factory.py as it's misleading when creating a resource loader just for Native Question Answerer. Or look for an alternative route to creating a resource loader. Or make sure that the duckling flag is indeed trickled down to the Native QA

ritvikshrivastava: If we're okay with a NoOpSystemEntityRecognizer for this use-case, then we can just pass ``system_entity_recognizer = NoOpSystemEntityRecognizer.get_instance()`` as a kwarg to create_query_factory. This would allow for passing the default recognizer while preventing the error message.

ritvikshrivastava:
```suggestion
                system_entity_recognizer = NoOpSystemEntityRecognizer.get_instance()
```

murali1996: Seems like a good alternative.

vrdn-23 (approved):
LGTM!

PR #397, opened 2022-02-15
Setting default tokenizer for All Languages to Whitespace

KKsharma99:
Setting the default tokenizer for English to WhiteSpace. When tokenizing the phrase "what's up" the WhiteSpace tokenizer and SpacyTokenizer will normalize and tokenize this phrase differently such that the apostrophe is removed with the SpacyTokenizer. This PR switches back the default for English and updates the documentation accordingly.

WhiteSpace Tokenizer:
The WhiteSpace Tokenizer will split this phrase to "what's", "up". When MindMeld's default regex normalization rules process the input, the normalized tokens are still "what 's" and "up". Lastly, the tokens are tokenized again (subtokenization) so the final tokens are "what", "'s", "up". The apostrophe is retained.

Spacy Tokenizer:
The Spacy Tokenizer will split this phrase to "what", "'s", "up". Then when passed into the default regex normalization rules, the apostrophe in front of the s is removed because all apostrophes at the start of tokens are removed by default. The resulting normalized tokens are "what", "s", "up".

Reverting the default tokenizer to the EnglishTokenizer will improve backwards compatibility and performance. A user can still explicitly select the Spacy Tokenizer if needed in the config.


mindmeld/text_preparation/tokenizers.py
```diff
@@ -304,7 +304,6 @@ def get_tokenizer_by_language(language):
         Returns:
             (Tokenizer): Tokenizer Class
         """
-        if language in SPACY_SUPPORTED_LANGUAGES:
+        if language != ENGLISH_LANGUAGE_CODE and language in SPACY_SUPPORTED_LANGUAGES:
```
vembar, line 307: Think its better we set the default to whitespace tokenizer for all languages so we know it's backwards compatible?

serapio: Yeah, agree.

KKsharma99: Sure, we can do that

KKsharma99: The WhiteSpaceTokenizer is now the default for all languages, the documentation has also been updated accordingly

vembar (approved):

serapio (approved):

PR #398, opened 2022-02-16
Tokenization Performance Improvements

KKsharma99:
This PR improves the time it takes to create queries if pre-loaded data in the QueryCache is not present.

Problem: The primary prior tokenization technique was to use the WhiteSpaceTokenizer which is about 200 times faster than Spacy if a sentence with 3 entities is passed in all at once. It is about 2000 times faster than a sentence with 3 entities passed in as 5 chunks around the entity label. As a result, applications with a large volume of training data have considerably longer build times as build times include initial query creation times.

Solution: 3 Optimization Techniques are presented in this PR to improve performance.

Optimization 1: Convert Multi-Shot Tokenization to Single-Shot Tokenization. Instead of processing queries in separate chunks and making several calls to Spacy's tokenize method, we can determine the spans of the unannotated text and only pass that into Spacy. We can translate the indices back to those that relate to the annotated text using a mapping vector which lists each value in the index (More details in the presentation below).

Optimization 2: Reduce Sub-Tokenization. The workflow for the TextPreparationPipeline is the following: Text is tokenized, each token is normalized independently, and then each normalized token is tokenized once again. This is to handle cases where, for example a space is added during the normalization step. The phrase "whats's up" gets tokenized as "what's", "up". After normalization, a space is added before the apostrophe but there are still just two tokens. "what 's", "up". Finally, we tokenize all normalized tokens to check for these type of cases and the final tokens are "what", "'s", "up". For large applications, sub-tokenizing every token can become very computationally expensive since making a call to Spacy's tokenizer is time consuming. We can reduce about 80% of tokenization cases by simply checking if the characters have changed after normalization, if they haven't there is no need to use Spacy.

Optimization 3: Disable Unused Spacy Pipeline. By default, Spacy uses "tokenizer", "ner", "tagger", and "parser" on any given text. We disable all components excluding the "tokenizer" to reduce computation. This PR enables a disable list to be propagated when loading the Spacy model using the SpacyModelFactory.

For more details, you can view this presentation: [Link](https://cisco-my.sharepoint.com/:p:/p/kunshar2/EfGQJcghuUxKmNaeJaLF5z4B7eFs5tgGGNU8N75E3-rVmg?e=rVZnbT)


mindmeld/text_preparation/tokenizers.py
```diff
@@ -242,8 +242,9 @@ def __init__(self, language, spacy_model_size="sm"):
             spacy_model_size (str, optional): Size of the Spacy model to use. ("sm", "md", or "lg")
         """
         self.spacy_model = SpacyModelFactory.get_spacy_language_model(
-            language, spacy_model_size
+            language, spacy_model_size, disable=["tagger", "parser", "ner"]
```
serapio, line 245: Since the typical usage doesn't use any pipeline components, perhaps we should just be using `spacy.blank(language)`. This would mean apps wouldn't need to depend on the language model packages, but some languages would still need a tokenizer library, e.g. jieba for zh, pysudachi for ja.

serapio (changes requested):
Since the defaults don't need the language model, we should refactor this to not depend on them.

mindmeld/text_preparation/tokenizers.py
murali1996: A thought: You can also disable the pipes for spacy models upon loading the model.
Example:
```
        self.spacy_model = SpacyModelFactory.get_spacy_language_model(
            language, spacy_model_size
        )
        self.spacy_model.disable_pipes('ner', 'tagger', 'parser')
```
This can help in not having to pass on the information along the hierarchy (pointing this because I don't think we are anyways storing the list of disable pipe names as an instance attribute in [mindmeld/text_preparation/spacy_model_factory.p](https://github.com/cisco/mindmeld/pull/398/files?diff=unified&w=1#diff-a248bdddac776431435cbd12e577828540f8eef0026f5e56eaf3b627b5b33b4a))

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -335,6 +349,105 @@ def find_mindmeld_annotation_re_matches(text):
...
+        """
+        for i in range(len(tokens)):
+            tokens[i]["start"] = unann_to_ann_idx_map[tokens[i]["start"]]
+
+    def tokenize_using_spacy(self, text):
+        """Wrapper function used before tokenizing with Spacy. Combines all unannoted text spans
+        into a single string to pass to spacy for tokenization. Applies the correct offset to
+        the resulting tokens to align with the annotated text. This optimization reduces the overall
+        time needed for tokenization.
+
```
murali1996, line 433: Out of curiosity, what is restricting us from doing the same with other tokenizer types too? I mean, isn't this technique generic enough optimization across all types of tokenizers? Do you think this part of computation 
> Applies the correct offset to the resulting tokens to align with the annotated text.
offsets the computational speed we gain with combined text vs. multiple calls with each span?

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -253,9 +261,13 @@ def tokenize_and_normalize(self, text):
             if not raw_token["text"]:
                 continue
             normalized_text = self._normalize_text(raw_token["text"])
-            # We tokenize the post-norm text and split the entity if possible
+            # We sub-tokenize the post-norm text and split the entity if possible
             # Ex: normalize("o'clock") -> "o clock" -> ["o", "clock"]
-            normalized_texts = [t["text"] for t in self.tokenize(normalized_text)]
+            # Skip sub-tokenization call if characters are not added/removed
+            if normalized_text.lower() == raw_token["text"].lower():
+                normalized_texts = [normalized_text]
```
murali1996, line 268: Just a quick thought. Is there any speedup merits if we maintain a lookup dictionary for (sub-)tokenization phase? Like, we can fetch the tokenization outputs of an input from a table lookup if that input sub-text is already tokenized once previously. I think this technique can help us because there can be a lot of repetitive sub-texts in our training datasets. What do you think?

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -335,6 +349,105 @@ def find_mindmeld_annotation_re_matches(text):
...
+
+        if len(unann_spans) > 0:
+            unann_spans.append((prev_entity_end, len(text)))
+
+        # Filter out spans that have a length of 0
+        unann_spans = [span for span in unann_spans if span[1] - span[0] > 0]
+        return unann_spans
+
+    @staticmethod
+    def unann_to_ann_idx_map(unann_spans):
```
murali1996, line 392: nit: can we expand the unann to unannotated and similarly for ann? Improves readability I think.

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -253,9 +261,13 @@ def tokenize_and_normalize(self, text):
...
             normalized_text = self._normalize_text(raw_token["text"])
-            # We tokenize the post-norm text and split the entity if possible
+            # We sub-tokenize the post-norm text and split the entity if possible
             # Ex: normalize("o'clock") -> "o clock" -> ["o", "clock"]
-            normalized_texts = [t["text"] for t in self.tokenize(normalized_text)]
+            # Skip sub-tokenization call if characters are not added/removed
+            if normalized_text.lower() == raw_token["text"].lower():
+                normalized_texts = [normalized_text]
+            else:
+                normalized_texts = [t["text"] for t in self.tokenize(normalized_text)]
```
murali1996, line 270: Just a quick thought. Is there any speedup merits if we maintain a lookup dictionary for (sub-)tokenization phase? Like, we can fetch the tokenization outputs of an input from a table lookup if that input sub-text is already tokenized once previously. I think this technique can help us because there can be a lot of repetitive sub-texts in our training datasets. What do you think?

murali1996 (commented):
Left few comments. Looks good overall. Thanks for taking care of improving speed ups with Spacy @KKsharma99

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: That's a good point, however, the reason the optimization works for Spacy is that Spacy guarantees that characters are not added or removed. Since the Text Preparation Pipeline has been designed to be extensible and fully customizable there isn't a guarantee that a user's tokenizer implementation will also uphold this property.

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Sure, made the requested change

snow0x2d0 (changes requested):
A couple of pythony nit-picks and suggestions

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -623,7 +744,7 @@ def create_from_app_config(app_path):
         stemmer = (
             "NoOpStemmer"
             if "stemmer" in text_preparation_config
-               and not text_preparation_config["stemmer"]
+            and not text_preparation_config["stemmer"]
             else text_preparation_config.get("stemmer")
         )
```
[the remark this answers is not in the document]
KKsharma99: Thanks for the tip, did this for normalizers as well.

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -335,6 +349,109 @@ def find_mindmeld_annotation_re_matches(text):
...
+        Returns:
+            unannotated_to_annotated_idx_map (List[Tuple(int, int)]): A vector where the value at
+                each index represents the mapping of the position of a single character
+                in the unannotated text to the position in the original text.
+        """
+        unannotated_to_annotated_idx_map = []
+        for unannotated_span in unannotated_spans:
+            start, end = unannotated_span
+            for i in range(start, end):
+                unannotated_to_annotated_idx_map.append(i)
```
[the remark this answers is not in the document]
KKsharma99: Good Catch, fixed

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -321,7 +333,9 @@ def get_hashid(self):
         Returns:
             256 character hash representation of current TextPreparationPipeline config (str) .
         """
-        string = json.dumps(self, cls=TextPreparationPipelineJSONEncoder, sort_keys=True)
+        string = json.dumps(
+            self, cls=TextPreparationPipelineJSONEncoder, sort_keys=True
+        )
```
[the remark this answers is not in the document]
KKsharma99: Fixed

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -335,6 +349,109 @@ def find_mindmeld_annotation_re_matches(text):
...
+            prev_entity_end = entity_end
+
+        if len(unannotated_spans) > 0:
+            unannotated_spans.append((prev_entity_end, len(text)))
+
+        # Filter out spans that have a length of 0
+        unannotated_spans = [
+            span for span in unannotated_spans if span[1] - span[0] > 0
+        ]
+        return unannotated_spans
```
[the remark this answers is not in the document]
KKsharma99: Updated accordingly

mindmeld/text_preparation/spacy_model_factory.py
```diff
@@ -30,12 +30,14 @@ class SpacyModelFactory:
     """Spacy (Language) Model Factory Class"""
 
     @staticmethod
-    def get_spacy_language_model(language, spacy_model_size="lg"):
+    def get_spacy_language_model(language, spacy_model_size="lg", disable=[]):
```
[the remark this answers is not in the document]
serapio: Or have the default be an empty tuple `disable=()` and use type `Iterable[str]` in the doc string.

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -169,16 +170,17 @@ def preprocess(self, text):
         return preprocessed_text
 
     def custom_preprocessors_exist(self):
-        """ Checks if the current TextPreparationPipeline has preprocessors that is not
+        """Checks if the current TextPreparationPipeline has preprocessors that is not
```
serapio, line 173:
```suggestion
        """Checks if the current TextPreparationPipeline has preprocessors that are not
```

serapio (approved):

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: This is a good point, however, from analysis of a couple of PR most sub-tokens are unique. I'm hesitant that storing all unique sub-tokens will cause a build-up in RAM memory especially for large applications that have 50,000+ queries

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Fixed

mindmeld/text_preparation/tokenizers.py
KKsharma99: Opting for the disable param option to retain functionality for selection a partial set of components

mindmeld/text_preparation/text_preparation_pipeline.py
```diff
@@ -335,6 +346,107 @@ def find_mindmeld_annotation_re_matches(text):
         """
         return list(MINDMELD_ANNOTATION_PATTERN.finditer(text))
 
+    @staticmethod
+    def calc_unannotated_spans(text):
+        """Calculates the spans of text that exclude mindmeld entity annotations.
```
vembar, line 351: Just a nice to have: 
Going by the naming, I assumed this function (and other variables used named similarly) will return all spans other than entity spans. 
Can't think of a better alternative though - `calc_spans_without_annotation_markers()` maybe?

mindmeld/text_preparation/tokenizers.py
vembar: Seems like we can also disable `lemmatizer` and `attribute_ruler`

mindmeld/text_preparation/tokenizers.py
```diff
@@ -241,8 +241,9 @@ def __init__(self, language, spacy_model_size="sm"):
...
```
vembar, line 244: Seems like we can also disable `lemmatizer` and `attribute_ruler`

vembar (approved):
Just a couple of comments, LGTM otherwise!

mindmeld/text_preparation/tokenizers.py
KKsharma99: Sure, expanded the disable list

mindmeld/text_preparation/text_preparation_pipeline.py
KKsharma99: Clarified the definition/explanation as discussed

PR #400, opened 2022-02-19
Updated Dockerfiles to build using current Mindmeld version

Zozman:
As stated in Issue #378, currently the Docker builds are unbuildable for both the example docker container and dev container.  This is because currently when mindmeld runs it attempts to install the `en_core_web_sm` model from Spacy but is unable to run the command in the container environment, failes, and the container explodes.  Therefore this PR changes the build by manually having the Docker build add `en_core_web_sm` so that it's already there when mindmeld tests for it and mindmeld itself doesn't have to run it.  Unlike PR #380, this fix is not tied to any poarticular `spacy` version.

Also updated `get-pip.py` scripts to work as the current script will not work with the current container.

To test:

1. Go to `docker_comtainers/mindmeld_docker` and `docker_containers/mindmeld_dev_docker`
2. Run `docker-compose up --build`

### DIFF
--- a/docker_containers/mindmeld_dep_docker/Dockerfile
+++ b/docker_containers/mindmeld_dep_docker/Dockerfile
@@ -55,7 +55,7 @@ COPY ./conf/elasticsearch.yml /elasticsearch/config/elasticsearch.yml
 
 WORKDIR /root
 
-RUN curl -o /tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py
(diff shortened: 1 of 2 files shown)


snow0x2d0 (changes requested):
LGTM. If you wouldn't mind just merging in, or rebasing onto, master first so the checks and tests can all run I'll merge it when they finish. Thanks!

Zozman:
@snow0x2d0 Pulled latest master in.

snow0x2d0 (approved):

snow0x2d0:
:shipit:

PR #402, opened 2022-02-25
Pin markupsafe version

snow0x2d0:
We currently depend on an older flask, which requires an older version of
jinja, which doesn't pin its version of markupsafe, and markupsafe 2.1.0
removed a deprecated function that the older jinja version depends on.
As a stop gap, we'll pin our version of markupsafe to that last 2.0
version. Longer term we need to update our flask dependency, and likely
move it to an extra.


chadoakley (approved):
Well done on the comment.

PR #408, opened 2022-03-11
Fix for letter tokenizer

serapio:
The logic for the letter tokenizer was checking for transitions from Latin character to non-Latin characters, but not for transitions from non-Latin characters to Latin characters. So if a mixed string appeared (without spaces at the boundaries) one non-Latin character was getting attached to the Latin character token. e.g.
`"名前はJohnです" => `
`  ["名", "前", "はJohn", "で", "す"]`
instead of 
`  ["名", "前", "は", "John", "で", "す"]`


mhuang2173 (approved):

vrdn-23 (approved):
LGTM

PR #410, opened 2022-03-24
minor inconsistency in evaluate printout

serapio:
clean up console output of `mindmeld evaluate`, formerly like this:
```
Domain classification accuracy: '1.0'
Intent classification accuracy for the general domain: 1.0
Entity recognition accuracy for the 'general.select' intent: 1.0
```


vrdn-23 (approved):
lgtm!

PR #413, opened 2022-04-29
Integrated working PyTorch-CRF in MM

vrdn-23:
Folks,

This is the first draft PR that I have ready for moving from the old rusty sklearn-crfsuite to the new (hopefully better) pytorch-CRF. A couple of things to note here:

- I haven't really removed the original CRF code from the Mindmeld repo just yet. I still want to make sure we don't have major accuracy drops when it comes to the ANLP repo. My individual experiments on the ANLP repo suggests that we should be fine, but there could always be pitfalls that I did not account for while integrating and I want to be sure before we move on to something new. Hence for now, that way this is structured is that we can switch between the cry models based on "crf" and "torch-crf". Ideally I would either want to remove the old CRF completely by the time this PR is merged or mark it as deprecated for a couple of RCs and refactor it out once we move to the next major release.
- I still need to add some unit tests to this, although I'm not sure if there are particular unit tests that would work for a PyTorch model
- The design for this isn't really perfect. So I would appreciate any suggestions or modifications for a much cleaner codebase. I'm trying to make sure that we stick to the crfsuite implementation that we have as closely as possible so that when we do eventually switch models, it doesn't involve a lot of refactoring and that most things are taken care of by the new PyTorch class. If this might be the wrong approach to look at things, please let me know!
- The functions for compute_marginals which is required for the predict_proba class in CRF has been [borrowed from a PR](https://github.com/kmkurn/pytorch-crf/pull/37) raised against the original pytorch-crf repo. For reasons unknown, it never got merged into the master branch, but I've gone through the implementation a bit and it looks solid and the original author also seemed to approve the implementation. 

Please feel free to comment or let me know if you have any questions regarding the PyTorch code, or if I've missed anything essential with regards to integrating this new monstrosity!
Thanks for all the help in advance! Looking forward to meeting everyone at a newer scikit-learn version soon! :)

### DIFF
--- a/mindmeld/models/tagger_models.py
+++ b/mindmeld/models/tagger_models.py
@@ -27,7 +27,7 @@ from .helpers import (
 )
 from .model import ModelConfig, Model, PytorchModel, AbstractModelFactory
 from .nn_utils import get_token_classifier_cls, TokenClassificationType
-from .taggers.crf import ConditionalRandomFields
+from .taggers.crf import ConditionalRandomFields, PyTorchCRF
 from .taggers.memm import MemmModel
 from ..exceptions import MindMeldError
 
@@ -73,12 +73,13 @@ class TaggerModel(Model):
     CRF_TYPE = "crf"
     MEMM_TYPE = "memm"
     LSTM_TYPE = "lstm"
-    ALLOWED_CLASSIFIER_TYPES = [CRF_TYPE, MEMM_TYPE, LSTM_TYPE]
+    TORCH_CRF_TYPE = "torch-crf"
+    ALLOWED_CLASSIFIER_TYPES = [CRF_TYPE, MEMM_TYPE, LSTM_TYPE, TORCH_CRF_TYPE]
 
     # for default model scoring types
     ACCURACY_SCORING = "accuracy"
     SEQ_ACCURACY_SCORING = "seq_accuracy"
-    SEQUENCE_MODELS = ["crf"]
+    SEQUENCE_MODELS = ["crf", "torch-crf"]
 
     DEFAULT_FEATURES = {
         "bag-of-words-seq": {
@@ -131,6 +132,7 @@ class TaggerModel(Model):
             return {
                 TaggerModel.MEMM_TYPE: MemmModel,
                 TaggerModel.CRF_TYPE: ConditionalRandomFields,
+                TaggerModel.TORCH_CRF_TYPE: PyTorchCRF,
                 TaggerModel.LSTM_TYPE: LstmModel,
             }[classifier_type]
         except KeyError as e:
@@ -231,7 +233,7 @@ class TaggerModel(Model):
                 "There are no labels in this label set, so we don't fit the model."
             )
             return self
-        # Extract labels - label encoders are the same accross all entity recognition models
+        # Extract labels - label encoders are the same across all entity recognition models
         self._label_encoder = get_label_encoder(self.config)
         y = self._label_encoder.encode(labels, examples=examples)
 
@@ -246,8 +248,8 @@ class TaggerModel(Model):
             self._current_params = params
         else:
             # run cross validation to select params
-            if self._clf.__class__ == LstmModel:
-                raise MindMeldError("The LSTM model does not support cross-validation")
+            if self._clf.__class__ in (LstmModel, PyTorchCRF):
+                raise MindMeldError(f"The {self._clf.__class__.__name__} model does not support cross-validation")
 
             _, best_params = self._fit_cv(X, y, groups)
             self._clf = self._fit(X, y, best_params)
--- a/mindmeld/models/taggers/crf.py
+++ b/mindmeld/models/taggers/crf.py
@@ -23,6 +23,7 @@ from sklearn_crfsuite import CRF
 
 from .taggers import Tagger, extract_sequence_features
 from ..helpers import FileBackedList
+from .pytorch_crf import TorchCRF
 
 logger = logging.getLogger(__name__)
 
@@ -176,6 +177,150 @@ class ConditionalRandomFields(Tagger):
                     temp_list.append("{}={}".format(feat_type, str(feature[feat_type])))
                 feat_list.append(temp_list)
             new_X.append(feat_list)
+        return new_X
+
+    def setup_model(self, config):
+        self._feat_binner = FeatureBinner()
+
+
+class PyTorchCRF(Tagger):
+    """A Conditional Random Fields model."""
+
+    @staticmethod
+    def _predict_proba(X):
+        del X
+        pass
+
+    @staticmethod
+    def load(model_path):
+        del model_path
+        pass
+
+    def fit(self, X, y):
+        self._clf.fit(X, y)
+        return self
+
+    def set_params(self, **parameters):
+        self._clf = TorchCRF()
+        self._clf.set_params(**parameters)
+        return self
+
+    def get_params(self, deep=True):
+        return self._clf.get_params()
+
+    def predict(self, X, dynamic_resource=None):
+        return self._clf.predict(X)
+
+    def predict_proba(self, examples, config, resources):
+        """
+        Args:
+            examples (list of mindmeld.core.Query): a list of queries to predict on
+            config (ModelConfig): The ModelConfig which may contain information used for feature
+                                  extraction
+            resources (dict): Resources which may be used for this model's feature extraction
+
+        Returns:
+            list of tuples of (mindmeld.core.QueryEntity): a list of predicted labels \
+             with confidence scores
+        """
+        X, _, _ = self.extract_features(examples, config, resources, in_memory=True)
+        seq = self._clf.predict(X)
+        marginals_dict = self._clf.predict_marginals(X)
+        marginal_tuples = []
+        for query_index, query_seq in enumerate(seq):
+            query_marginal_tuples = []
+            for i, tag in enumerate(query_seq):
+                query_marginal_tuples.append([tag, marginals_dict[query_index][i][tag]])
+            marginal_tuples.append(query_marginal_tuples)
+        return marginal_tuples
+
+    def predict_proba_distribution(self, examples, config, resources):
+        """
+        Args:
+            examples (list of mindmeld.core.Query): a list of queries to predict on
+            config (ModelConfig): The ModelConfig which may contain information used for feature
+                                  extraction
+            resources (dict): Resources which may be used for this model's feature extraction
+
+        Returns:
+            list of tuples of (mindmeld.core.QueryEntity): a list of predicted labels \
+             with confidence scores
+        """
+        X, _, _ = self.extract_features(examples, config, resources, in_memory=True)
+        seq = self._clf.predict(X)
+        marginals_dict = self._clf.predict_marginals(X)
+        predictions = []
+        tag_maps = []
+        for query_index, query_seq in enumerate(seq):
+            tags = []
+            preds = []
+            for i in range(len(query_seq)):
+                tags.append(list(marginals_dict[query_index][i].keys()))
+                preds.append(list(marginals_dict[query_index][i].values()))
+            tag_maps.extend(tags)
+            predictions.extend(preds)
+        return [[tag_maps, predictions]]
+
+    def extract_features(self,
+                         examples,
+                         config,
+                         resources,
+                         y=None,
+                         fit=False,
+                         in_memory=STORE_CRF_FEATURES_IN_MEMORY):
+        """Transforms a list of examples into a feature matrix.
+
+        Args:
+            examples (list of mindmeld.core.Query): a list of queries
+            config (ModelConfig): The ModelConfig which may contain information used for feature
+                                  extraction
+            resources (dict): Resources which may be used for this model's feature extraction
+
+        Returns:
+            (list of list of str): features in CRF suite format
+        """
+        # Extract features and classes
+        feats = [] if in_memory else FileBackedList()
+        for _, example in enumerate(examples):
+            feats.append(self.extract_example_features(example, config, resources))
+        X = self._preprocess_data(feats, fit)
+        return X, y, None
+
+    @staticmethod
+    def extract_example_features(example, config, resources):
+        """Extracts feature dicts for each token in an example.
+
+        Args:
+            example (mindmeld.core.Query): A query.
+            config (ModelConfig): The ModelConfig which may contain information used for feature \
+                                  extraction.
+            resources (dict): Resources which may be used for this model's feature extraction.
+
+        Returns:
+            list[dict]: Features.
+        """
+        return extract_sequence_features(
+            example, config.example_type, config.features, resources
+        )
+
+    def _preprocess_data(self, X, fit=False):
+        """Converts data into formats of CRF suite.
+
+        Args:
+            X (list of dict): features of an example
+            fit (bool, optional): True if processing data at fit time, false for predict time.
+
+        Returns:
+            (list of list of str): features in CRF suite format
+        """
+        if fit:
+            self._feat_binner.fit(X)
+
+        # We want to use a list for in-memory and a LineGenerator for disk based
+        new_X = X.__class__()
+        # Maintain append code structure to make sure it supports in-memory and FileBackedList()
+        for feat_seq in self._feat_binner.transform(X):
+            new_X.append(feat_seq)
         return new_X
 
     def setup_model(self, config):
--- a/mindmeld/models/taggers/pytorch_crf.py
+++ b/mindmeld/models/taggers/pytorch_crf.py
@@ -1,0 +1,381 @@
+import logging
+import os
+import random
+import uuid
+from collections import Counter
+from copy import copy
+from itertools import zip_longest
+from random import randint
+
+import numpy as np
+import torch
+import torch.nn as nn
+from sklearn.feature_extraction import DictVectorizer, FeatureHasher
+from sklearn.metrics import f1_score
+from sklearn.model_selection import train_test_split
+from sklearn.preprocessing import LabelEncoder
+from torch import optim
+from torch.utils.data import Dataset, DataLoader
+from torch_sparse import cat, tensor
+from torchcrf import CRF
+from tqdm import tqdm
+
+from ...exceptions import MindMeldError
+from ...path import USER_CONFIG_DIR
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_PYTORCH_CRF_ER_CONFIG = {
+    "feat_type": "hash",  # ["hash", "dict"]
+    "feat_num": 50000,
+    "stratify": True,
+    "drop_input": 0.2,
+    "train_batch_size": 8,
+    "patience": 3,
+    "epochs": 100,
+    "train_dev_split": 0.15,
+    "optimizer_type": "sgd",  # ["sgd", "adam"]
+}
+
+
+class TaggerDataset(Dataset):
+    def __init__(self, input, seq_lens, labels=None):
+        self.input = input
+        self.labels = labels
+        self.seq_lens = seq_lens
+        self.max_seq_length = max(seq_lens)
+
+    def __len__(self):
+        return len(self.seq_lens)
+
+    def __getitem__(self, index):
+        mask_list = [1] * self.seq_lens[index] + [0] * (self.max_seq_length - self.seq_lens[index])
+
+        mask = torch.as_tensor(mask_list, dtype=torch.bool)
+        if self.labels:
+            return self.input[index], mask, self.labels[index]
+        else:
+            return self.input[index], mask
+
+
+def init_weights(m):
+    if type(m) == nn.Linear:
+        torch.nn.init.xavier_normal_(m.weight)
+        m.bias.data.fill_(0.01)
+
+
+def custom_collate(sequence):
+    if len(sequence[0]) == 3:
+        sparse_mats, masks, labels = zip(*sequence)
+        return cat(sparse_mats, dim=(0, 1)).to_torch_sparse_coo_tensor().coalesce(), torch.stack(masks), torch.stack(
+            labels)
+    elif len(sequence[0]) == 2:
+        sparse_mats, masks = zip(*sequence)
+        return cat(sparse_mats, dim=(0, 1)).to_torch_sparse_coo_tensor().coalesce(), torch.stack(masks)
+
+
+class Encoder:
+    def __init__(self, feature_extractor="dict", num_feats=None):
+
+        self.feat_extractor = DictVectorizer(dtype=np.float32) if feature_extractor == "dict" else FeatureHasher(
+            n_features=num_feats, dtype=np.float32)
+        self.label_encoder = LabelEncoder()
+        self.feat_extract_type = feature_extractor
+        self.num_classes = None
+        self.classes = None
+        self.num_feats = num_feats
+        self.fit_done = False
+
+    def get_tensor_data(self, feat_dicts, labels=None, fit=False):
+        if labels is None:
+            labels = []
+        if fit:
+            if self.feat_extract_type == "dict":
+                comb_dict_list = [x for seq in feat_dicts for x in seq]
+                self.feat_extractor.fit(comb_dict_list)
+                self.num_feats = len(self.feat_extractor.get_feature_names())
+            if labels:
+                self.label_encoder.fit([x for l in labels for x in l])
+                self.pad_index = len(self.label_encoder.classes_) - 1
+                self.classes = self.label_encoder.classes_
+                self.num_classes = len(self.label_encoder.classes_)
+
+            self.fit_done = True
+        feats = []
+        encoded_labels = []
+        seq_lens = [len(x) for x in feat_dicts]
+        max_seq_len = max(seq_lens)
+
+        for i, (x, y) in enumerate(zip_longest(feat_dicts, labels)):
+
+            padded_x = x + [{}] * (max_seq_len - seq_lens[i])
+            sparse_feat = self.feat_extractor.transform(padded_x)
+            sparse_feat_tensor = tensor.from_scipy(sparse_feat)
+            feats.append(sparse_feat_tensor)
+
+            if y:
+                transformed_label = self.label_encoder.transform(y)
+                transformed_label = np.pad(transformed_label, pad_width=(0, max_seq_len - seq_lens[i]),
+                                           constant_values=self.pad_index)
+                label_tensor = torch.as_tensor(transformed_label, dtype=torch.long)
+                encoded_labels.append(label_tensor)
+        return (feats, encoded_labels, seq_lens) if encoded_labels else (feats, seq_lens)
+
+
+class TorchCRF(nn.Module):
+    def __init__(self):
+        super(TorchCRF, self).__init__()
+        self.optimizer = None
+        self.encoder = None
+        self.best_model_save_path = os.path.join(USER_CONFIG_DIR, "tmp", str(uuid.uuid4()), "best_crf_model.pt")
+        os.makedirs(os.path.dirname(self.best_model_save_path), exist_ok=True)
+
+    def set_random_states(self):
+        torch.manual_seed(self.random_state)
+        random.seed(self.random_state + 1)
+        np.random.seed(self.random_state + 2)
+
+    def validate_params(self):
+        if self.optimizer_type not in ["sgd", "adam"]:
+            raise MindMeldError(
+                f"Optimizer type {self.optimizer_type} not supported. Supported options are ['sgd', 'adam']")
+        elif self.feat_type not in ["hash", "dict"]:
+            raise MindMeldError(f"Feature type {self.feat_type} not supported. Supported options are ['hash', 'dict']")
+        elif not 0 < self.train_dev_split < 1:
+            raise MindMeldError(f"Train-dev split should be a value between 0 and 1.")
+        elif not 0 <= self.drop_input < 1:
+            raise MindMeldError(f"Drop Input should be a value between 0 and 1. (inclusive)")
+
+        for x, y in zip([self.feat_num, self.train_batch_size, self.patience, self.epochs],
+                        ["Number of features", "Train Batch size", "Patience", "Number of epochs"]):
+            if not isinstance(x, int):
+                raise MindMeldError(f"{y} should be am integer value.")
+
+    def build_params(self, num_features, num_classes):
+        self.W = nn.Parameter(torch.nn.init.xavier_normal_(torch.empty(size=(num_features, num_classes))),
+                              requires_grad=True)
+        self.b = nn.Parameter(torch.nn.init.constant_(torch.empty(size=(num_classes,)), val=0.01),
+                              requires_grad=True)
+        self.crf_layer = CRF(num_classes, batch_first=True)
+        self.crf_layer.apply(init_weights)
+        self.num_classes = num_classes
+
+    def forward(self, inputs, targets, mask, drop_input=0.0):
+        if drop_input:
+            dp_mask = (torch.FloatTensor(inputs.values().size()).uniform_() > drop_input)
+            inputs.values()[:] = inputs.values() * dp_mask
+        dense_W = torch.tile(self.W, dims=(mask.shape[0], 1))
+        out_1 = torch.addmm(self.b, inputs, dense_W)
+        crf_input = out_1.reshape((mask.shape[0], -1, self.num_classes))
+        if targets is None:
+            return self.crf_layer.decode(crf_input, mask=mask)
+        loss = - self.crf_layer(crf_input, targets, mask=mask)
+        return loss
+
+    # The below implementation is borrowed from https://github.com/kmkurn/pytorch-crf/pull/37
+
+    def _compute_log_alpha(self, emissions, mask, run_backwards):
+        # emissions: (seq_length, batch_size, num_tags)
+        # mask: (seq_length, batch_size)
+
+        assert emissions.dim() == 3 and mask.dim() == 2
+        assert emissions.size()[:2] == mask.size()
+        assert emissions.size(2) == self.crf_layer.num_tags
+        assert all(mask[0].data)
+
+        seq_length = emissions.size(0)
+        mask = mask.float()
+        broadcast_transitions = self.crf_layer.transitions.unsqueeze(0)  # (1, num_tags, num_tags)
+        emissions_broadcast = emissions.unsqueeze(2)
+        seq_iterator = range(1, seq_length)
+
+        if run_backwards:
+            # running backwards, so transpose
+            broadcast_transitions = broadcast_transitions.transpose(1, 2)  # (1, num_tags, num_tags)
+            emissions_broadcast = emissions_broadcast.transpose(2, 3)
+
+            # the starting probability is end_transitions if running backwards
+            log_prob = [self.crf_layer.end_transitions.expand(emissions.size(1), -1)]
+
+            # iterate over the sequence backwards
+            seq_iterator = reversed(seq_iterator)
+        else:
+            # Start transition score and first emission
+            log_prob = [emissions[0] + self.crf_layer.start_transitions.view(1, -1)]
+
+        for i in seq_iterator:
+            # Broadcast log_prob over all possible next tags
+            broadcast_log_prob = log_prob[-1].unsqueeze(2)  # (batch_size, num_tags, 1)
+            # Sum current log probability, transition, and emission scores
+            score = broadcast_log_prob + broadcast_transitions + emissions_broadcast[
+                i]  # (batch_size, num_tags, num_tags)
+            # Sum over all possible current tags, but we're in log prob space, so a sum
+            # becomes a log-sum-exp
+            score = self._log_sum_exp(score, dim=1)
+            # Set log_prob to the score if this timestep is valid (mask == 1), otherwise
+            # copy the prior value
+            log_prob.append(score * mask[i].unsqueeze(1) +
+                            log_prob[-1] * (1. - mask[i]).unsqueeze(1))
+
+        if run_backwards:
+            log_prob.reverse()
+
+        return torch.stack(log_prob)
+
+    def compute_marginal_probabilities(self, inputs, mask):
+        # SWITCHING FOR BATCH FIRST DEFAULT
+        dense_W = torch.tile(self.W, dims=(mask.shape[0], 1))
+        out_1 = torch.addmm(self.b, inputs, dense_W)
+        emissions = out_1.reshape((mask.shape[0], -1, self.num_classes))
+        emissions = emissions.transpose(0, 1)
+        mask = mask.transpose(0, 1)
+        alpha = self._compute_log_alpha(emissions, mask, run_backwards=False)
+        beta = self._compute_log_alpha(emissions, mask, run_backwards=True)
+        z = torch.logsumexp(alpha[alpha.size(0) - 1] + self.crf_layer.end_transitions, dim=1)
+        prob = alpha + beta - z.view(1, -1, 1)
+        return torch.exp(prob).transpose(0, 1)
+
+    @staticmethod
+    def _log_sum_exp(tensor_input, dim):
+        # Find the max value along `dim`
+        offset, _ = tensor_input.max(dim)
+        # Make offset broadcastable
+        broadcast_offset = offset.unsqueeze(dim)
+        # Perform log-sum-exp safely
+        safe_log_sum_exp = torch.log(torch.sum(torch.exp(tensor_input - broadcast_offset), dim))
+        # Add offset back
+        return offset + safe_log_sum_exp
+
+    def set_params(self, **params):
+        self.feat_type = params.get('feat_type', DEFAULT_PYTORCH_CRF_ER_CONFIG['feat_type']).lower()
+        self.feat_num = params.get('feat_num', DEFAULT_PYTORCH_CRF_ER_CONFIG['feat_num'])
+        self.stratify = params.get('stratify', DEFAULT_PYTORCH_CRF_ER_CONFIG['stratify'])
+        self.drop_input = params.get('drop_input', DEFAULT_PYTORCH_CRF_ER_CONFIG['drop_input'])
+        self.train_batch_size = params.get('train_batch_size', DEFAULT_PYTORCH_CRF_ER_CONFIG['train_batch_size'])
+        self.patience = params.get('patience', DEFAULT_PYTORCH_CRF_ER_CONFIG['patience'])
+        self.epochs = params.get('epochs', DEFAULT_PYTORCH_CRF_ER_CONFIG['epochs'])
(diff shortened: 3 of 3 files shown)


snow0x2d0:
EEEEEEeeeeeEEEEE

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,419 @@
...
+    if len(sequence[0]) == 3:
+        sparse_mats, masks, labels = zip(*sequence)
+        return custom_coo_cat(sparse_mats), torch.stack(masks), torch.stack(labels)
+    elif len(sequence[0]) == 2:
+        sparse_mats, masks = zip(*sequence)
+        return custom_coo_cat(sparse_mats), torch.stack(masks)
+
+
+class Encoder:
+    def __init__(self, feature_extractor="dict", num_feats=None):
```
vembar, line 113: Looks like we set the default value for feature_extractor as "dict" here, but a "hash" in the default config?

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,419 @@
...
+        emissions = emissions.transpose(0, 1)
+        mask = mask.transpose(0, 1)
+        alpha = self._compute_log_alpha(emissions, mask, run_backwards=False)
+        beta = self._compute_log_alpha(emissions, mask, run_backwards=True)
+        z = torch.logsumexp(alpha[alpha.size(0) - 1] + self.crf_layer.end_transitions, dim=1)
+        prob = alpha + beta - z.view(1, -1, 1)
+        return torch.exp(prob).transpose(0, 1)
+
+    @staticmethod
+    def _log_sum_exp(tensor_input, dim):
```
vembar, line 277: Is this necessary? Can't we use torch.logsumexp() like in the previous method?
Documentation seems to suggest that it is numerically stable - https://pytorch.org/docs/stable/generated/torch.logsumexp.html#torch.logsumexp

vrdn-23:
Folks,
Just a small update since this hasn't seem any activity from my side. I ran some of the experiments to determine runtime and memory consumption of the new CRF module on my local system last week and the results have been inconsistent based on what I've been running in the background. As such I moved the experiments to an online server this week, and they are being run as we speak. 

I have enough evidence to suggest that the accuracy is not too bad from whence I first ran the experiments and am convinced that this PR is ready to graduate from its "Draft" status. It would be great if everyone could let me know what you think looks good and what you think needs a complete overhaul. I will be giving my full active attention to the PR starting this Monday (05/16) and hope to have the time/memory/accuracy analysis done and updated on the PR by then. 

I hope to have all outstanding comments resolved by Monday EOD and looking forward to addressing new ones!
TIA!

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: You're right! I left it in there cause the original author in the PR used it as such. But I ran some preliminary tests and it doesn't seem like it's any different behavior than what we expect from the in-built torch function!

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Fixed and changed both to hash

vrdn-23:
Screen Shot 2022-05-19 at 15 34 28

I don't have a a better way of formatting this as a markdown table, so I'm going to leave this a picture but these are the main conditions I think are worth knowing about this data:
- These numbers are the averages across 15 runs. 
- The different configurations tried here are the original CRF (sklearn-crfsuite), the torch-crf using the "dict" (aka dictionary) configuration, the torch-crf using the "hash" (aka feature hasher) configuration with a patience of 3, the torch-crf "hash" with a patience of 5. All of them use a batch size of 10 and other defaults are kept unless mentioned otherwise.
- For the time column, the times shown are the total time taken to train and predict on the entire train and test dataset respectively.
- These are the only 3 intents that use CRF in ANLP and hence these were the only ones I ran these numbers on. 
- The memory and time measurements are based on the memory usage and time taken to run the "er.fit()" and "er.predict()" lines for each intent-specific entity recognizer. 

The current configuration that I would support for use in ANLP would be the "hash" configuration with patience 5. The rest of the experimental settings are included to showcase an overall picture of how the torch-crf performs.
Some key takeaways that can be made from these values are:
- We do see an increase in memory used by the torch-crf and it is highest when we use the "dict" feature extractor and this is expected as the dictionary feature extractor takes into account all possible features while the feature hasher, helps reduce the number of sparse features we have. There is a 50-100 MiB increase in memory while training in the other configurations and if this is a problem that needs to be addressed, I can look further into seeing if there is anything that I can do to reduce the memory footprint of the code. However as expected, there is no increase in memory in increasing the patience, so choosing the patience: 5 configuration does not hurt us in this case.
- The times for training are not as bad as I had initially anticipated. Apart from the "dict" configuration, at worst we are probably 3-4 minutes slower than training from the original CRF. Personally, I don't see how we can speed up this and reducing the patience obviously has a positive effect in improving the time it takes to train, but I think that also comes at the cost of the final accuracy. Hence in the case of the pytorch-crf, choosing faster train times might come at the cost of accuracy which is why I'm preferring having a longer patience time as opposed to a smaller one. 
- One good thing I noticed is that it seems for bigger datasets (esp call_person) it does seem like inference is happening much faster than when using the original CRF although the improvement might not be as obvious when comparing on smaller datasets.
- The accuracies are not really that bad in terms on the original model. Atmost we see a 1-2% drop in all configs, but the primary reason for this, is the fact that we are sacrificing 15% of our training data for validation and my initial experiments did show that we could match the accuracies we got with the original CRF setup if allowed to train with the full ANLP data. The best way to fix this would be to invest time into a dev set that we can rely on for ANLP.

I realize that this is a lot to take in. But if you have any questions or comments, please feel free to reach out to me 1-on-1 or drop a comment here! I'm hoping that with enough review comments, we can make this code a little bit more cleaner/faster/lighter or all of the above!

vrdn-23:
6gtal3
cc @murali1996 @mattfrey-cisco @vembar @ritvikshrivastava @snow0x2d0 @tmehlinger @serapio

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,412 @@
...
+    def __len__(self):
+        return len(self.seq_lens)
+
+    def __getitem__(self, index):
+        mask_list = [1] * self.seq_lens[index] + [0] * (self.max_seq_length - self.seq_lens[index])
+
+        mask = torch.as_tensor(mask_list, dtype=torch.bool)
+        if self.labels:
+            return self.inputs[index], mask, self.labels[index]
+        else:
```
tmehlinger, line 55: you don't need `else` if you return early above :)

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,412 @@
...
+        self.feat_type = params.get('feat_type', DEFAULT_PYTORCH_CRF_ER_CONFIG['feat_type']).lower()
+        self.feat_num = params.get('feat_num', DEFAULT_PYTORCH_CRF_ER_CONFIG['feat_num'])
+        self.stratify = params.get('stratify', DEFAULT_PYTORCH_CRF_ER_CONFIG['stratify'])
+        self.drop_input = params.get('drop_input', DEFAULT_PYTORCH_CRF_ER_CONFIG['drop_input'])
+        self.train_batch_size = params.get('train_batch_size', DEFAULT_PYTORCH_CRF_ER_CONFIG['train_batch_size'])
+        self.patience = params.get('patience', DEFAULT_PYTORCH_CRF_ER_CONFIG['patience'])
+        self.epochs = params.get('epochs', DEFAULT_PYTORCH_CRF_ER_CONFIG['epochs'])
+        self.train_dev_split = params.get('train_dev_split', DEFAULT_PYTORCH_CRF_ER_CONFIG['train_dev_split'])
+        self.optimizer_type = params.get('optimizer_type', DEFAULT_PYTORCH_CRF_ER_CONFIG['optimizer_type']).lower()
+        self.verbose = (logger.getEffectiveLevel() == logging.DEBUG)
```
tmehlinger, line 286: You don't need to gate calls to `logger.debug` with a flag. Just emit debug logs and we can control output by setting the debug log level.

tmehlinger (changes requested):

mindmeld/models/taggers/crf.py
```diff
@@ -182,6 +183,150 @@ def setup_model(self, config):
         self._feat_binner = FeatureBinner()
 
 
+class PyTorchCRF(Tagger):
+    """A Conditional Random Fields model."""
+
+    @staticmethod
+    def _predict_proba(X):
```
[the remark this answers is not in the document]
vrdn-23: Fixed. Thanks!

mindmeld/models/taggers/crf.py
```diff
@@ -182,6 +183,150 @@ def setup_model(self, config):
...
+        del X
+        pass
+
+    @staticmethod
+    def load(model_path):
```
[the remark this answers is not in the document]
vrdn-23: Fixed again!

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Gotcha! Fixed!

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Oof. Can't believe I didn't think of that sooner! Thanks a bunch!

mattfrey-cisco:
@vrdn-23 - I don't think the memory usage matters too much anymore, @tmehlinger can weigh in on that.  When we were building for SPLAT, we had to use the env var documented here: https://www.mindmeld.com/docs/userguide/getting_started.html#mm-crf-features-in-memory

If you rerun the tests with that set to '0' the original CRF training will use significantly less memory.  I'm fine removing that option, but we need to verify our build pipelines can handle that memory usage, and we should remove that env var from the documentation.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,407 @@
...
+
+from ...exceptions import MindMeldError
+from ...path import USER_CONFIG_DIR
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_PYTORCH_CRF_ER_CONFIG = {
+    "feat_type": "hash",  # ["hash", "dict"]
+    "feat_num": 50000,
+    "stratify": True,
```
vembar, line 29: Nit: Can we rename to `stratify_train_val_split`?

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,407 @@
...
+    "drop_input": 0.2,
+    "train_batch_size": 8,
+    "patience": 3,
+    "epochs": 100,
+    "train_dev_split": 0.15,
```
vembar, line 34: similar nit: Can we rename to `train_val_split`? Don't want the dev set here to be confused with dev set files you can add as part of the hierarchy. Can be persuaded otherwise though :)

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,407 @@
...
+    def __init__(self, feature_extractor="hash", num_feats=50000):
+
+        self.feat_extractor = DictVectorizer(dtype=np.float32) if feature_extractor == "dict" else FeatureHasher(
+            n_features=num_feats, dtype=np.float32)
+        self.label_encoder = LabelEncoder()
+        self.feat_extract_type = feature_extractor
+        self.num_classes = None
+        self.classes = None
+        self.num_feats = num_feats
+        self.fit_done = False
```
vembar, line 122: To keep things consistent across the repo, can we change this to `self.ready`?

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,407 @@
...
+                comb_dict_list = [x for seq in feat_dicts for x in seq]
+                self.feat_extractor.fit(comb_dict_list)
+                self.num_feats = len(self.feat_extractor.get_feature_names())
+            if labels:
+                self.label_encoder.fit([x for l in labels for x in l])
+                self.pad_index = len(self.label_encoder.classes_) - 1
+                self.classes = self.label_encoder.classes_
+                self.num_classes = len(self.label_encoder.classes_)
+
+            self.fit_done = True
```
vembar, line 138: `self.ready`

tests/models/test_tagging.py
```diff
@@ -271,7 +271,7 @@ def test_get_boundary_counts_sequential(
 
 @pytest.mark.parametrize(
     "model_type,params",
-    [("memm", {"penalty": "l2", "C": 10000}), ("crf", {"c1": 0.01, "c2": 0.01})],
+    [("memm", {"penalty": "l2", "C": 10000}), ("crf", {"c1": 0.01, "c2": 0.01}), ("torch-crf", {"feat_type": "dict"})],
```
vembar, line 274: can we add a test for when feature type is hash?

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Fixed

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Renamed to train_val _split. I do think that would be the better recognized variable name.

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Changed. Although I'm not sure I've done anything special to make sure the self.ready is inherited anywhere. This is just on the encoder side. Not the model side.

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Changed

mindmeld/models/tagger_models.py
```diff
@@ -27,7 +27,7 @@
 )
 from .model import ModelConfig, Model, PytorchModel, AbstractModelFactory
 from .nn_utils import get_token_classifier_cls, TokenClassificationType
-from .taggers.crf import ConditionalRandomFields
+from .taggers.crf import ConditionalRandomFields, PyTorchCRF
```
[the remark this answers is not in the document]
vrdn-23: I've changed it. I was struggling on how to make the distinction between Tagger, Model and Layer. Thanks a lot!

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,407 @@
...
+        mask_list = [1] * self.seq_lens[index] + [0] * (self.max_seq_length - self.seq_lens[index])
+
+        mask = torch.as_tensor(mask_list, dtype=torch.bool)
+        if self.labels:
+            return self.inputs[index], mask, self.labels[index]
+
+        return self.inputs[index], mask
+
+
+def custom_coo_cat(tensors):
```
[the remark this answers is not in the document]
vrdn-23: Changed.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,407 @@
...
+    return torch.sparse_coo_tensor(indices=torch.stack([row, col]), values=value, size=sparse_sizes).coalesce()
+
+
+def init_weights(m):
+    if isinstance(m, nn.Linear):
+        torch.nn.init.xavier_normal_(m.weight)
+        m.bias.data.fill_(0.01)
+
+
+def custom_collate(sequence):
```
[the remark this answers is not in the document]
vrdn-23: Renamed to something more meaningful

mindmeld/models/taggers/pytorch_crf.py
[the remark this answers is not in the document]
vrdn-23: I've renamed it to self.ready. Although I'm not sure if Mindmeld is supposed to pick it up anywhere. This is in the encoder part of the model. Not the actual model itself. I'm not even sure we need it. Hmm

tests/models/test_tagging.py
vrdn-23: Done!

mindmeld/models/taggers/pytorch_crf.py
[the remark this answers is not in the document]
vrdn-23: So the variable in itself is something that I have leftover from porting my initial code. I don't use to anywhere as a flag or anything to check whether training has been done. I can either completely remove it, or should I be adding a check somewhere? I don't see an equivalent for this in the current crf-suite implementation so I'm not sure if maybe it's needed. Thoughts? @serapio @vembar

vrdn-23: Resolving conversation to shift it to under Lucien's comment.

tmehlinger:
On memory consumption, I'm not terribly concerned. Obviously don't allocate all the way to the moon just because it's convenient and keep it in line with present consumption (or even save memory!). However, we have plenty of flexibility to provide resources if you _need_ memory.

mindmeld/models/taggers/pytorch_crf.py
murali1996: Just a thought: It can be easier for MindMeld users/developers if params terminology is similar across different modules where pytorch is being used (see [params](https://www.mindmeld.com/docs/userguide/deep_neural_nets.html#common-configurable-params) used in deep NNs docs). Things like epochs/number_of_epochs, train_val_split/dev_split_ratio, train_batch_size/batch_size, etc.

mindmeld/models/tagger_models.py
```diff
@@ -73,12 +73,13 @@ class TaggerModel(Model):
...
     LSTM_TYPE = "lstm"
-    ALLOWED_CLASSIFIER_TYPES = [CRF_TYPE, MEMM_TYPE, LSTM_TYPE]
+    TORCH_CRF_TYPE = "torch-crf"
+    ALLOWED_CLASSIFIER_TYPES = [CRF_TYPE, MEMM_TYPE, LSTM_TYPE, TORCH_CRF_TYPE]
 
     # for default model scoring types
     ACCURACY_SCORING = "accuracy"
     SEQ_ACCURACY_SCORING = "seq_accuracy"
-    SEQUENCE_MODELS = ["crf"]
+    SEQUENCE_MODELS = ["crf", "torch-crf"]
```
murali1996, line 83: Once the old crf codebase is removed, do we intend to remove the "crf" model name or "torch-crf" ? 
Just a thought: I think because there will be a less chance that we create a different backend based crf in near future (i.e. after this PR is merged), it would be a good idea to retain the old name "crf" as this torch based crf is going to be the de facto crf implementation in MindMeld.

snow0x2d0 (changes requested):
Still looking over the `pytorch_crf.py` file but these are some initial style nitpicks that caught my eye.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+        return (feats, encoded_labels, seq_lens) if encoded_labels else (feats, seq_lens)
+
+
+# pylint: disable=too-many-instance-attributes
+class TorchCrfModel(nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.optimizer = None
+        self.encoder = None
+        self.best_model_save_path = os.path.join(USER_CONFIG_DIR, "tmp", str(uuid.uuid4()), "best_crf_model.pt")
```
murali1996, line 168: As dicussed offline, it is better to save model within the `.generated` folder of an app to avoid any unexpected issues while loading back models from disk.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+from torch import optim
+from torch.utils.data import Dataset, DataLoader
+from torchcrf import CRF
+
+from ...exceptions import MindMeldError
+from ...path import USER_CONFIG_DIR
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_PYTORCH_CRF_ER_CONFIG = {
```
murali1996, line 26: Better to move it to [components/_config.py](https://github.com/cisco/mindmeld/blob/master/mindmeld/components/_config.py) file.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+    "feat_type": "hash",  # ["hash", "dict"]
+    "feat_num": 50000,
+    "stratify_train_val_split": True,
+    "drop_input": 0.2,
+    "train_batch_size": 8,
+    "patience": 3,
+    "epochs": 100,
+    "train_val_split": 0.15,
+    "optimizer_type": "sgd",  # ["sgd", "adam"]
+}
```
murali1996, line 36: Would like to bring up the [comment](https://github.com/cisco/mindmeld/pull/413#discussion_r882145224) about trying to maintain similar terminology across different modules that use pytorch.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+
+        mask = torch.as_tensor(mask_list, dtype=torch.bool)
+        if self.labels:
+            return self.inputs[index], mask, self.labels[index]
+
+        return self.inputs[index], mask
+
+
+def diag_concat_coo_tensors(tensors):
+    assert len(tensors) > 0
```
murali1996, line 60: Might be a good idea to add a `logger.debug(..)` statement here with a brief message.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+        self.feat_type = params.get('feat_type', DEFAULT_PYTORCH_CRF_ER_CONFIG['feat_type']).lower()
+        self.feat_num = params.get('feat_num', DEFAULT_PYTORCH_CRF_ER_CONFIG['feat_num'])
+        self.stratify = params.get('stratify_train_val_split',
+                                   DEFAULT_PYTORCH_CRF_ER_CONFIG['stratify_train_val_split'])
+        self.drop_input = params.get('drop_input', DEFAULT_PYTORCH_CRF_ER_CONFIG['drop_input'])
+        self.train_batch_size = params.get('train_batch_size', DEFAULT_PYTORCH_CRF_ER_CONFIG['train_batch_size'])
+        self.patience = params.get('patience', DEFAULT_PYTORCH_CRF_ER_CONFIG['patience'])
+        self.epochs = params.get('epochs', DEFAULT_PYTORCH_CRF_ER_CONFIG['epochs'])
+        self.train_val_split = params.get('train_val_split', DEFAULT_PYTORCH_CRF_ER_CONFIG['train_val_split'])
+        self.optimizer_type = params.get('optimizer_type', DEFAULT_PYTORCH_CRF_ER_CONFIG['optimizer_type']).lower()
```
murali1996, line 286: I think it is better to raise an error if user mentions uppercased string like "Adam", "ADAM" , etc. instead of "adam" as we'll only mention the latter in the docs. Implementation-wise, I think this is not an issue here but if there was a scenario where we propagate the params further down into another method and missed to do lowercasing there; debugging will become difficult.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+        emissions = emissions.transpose(0, 1)
+        mask = mask.transpose(0, 1)
+        alpha = self._compute_log_alpha(emissions, mask, run_backwards=False)
+        beta = self._compute_log_alpha(emissions, mask, run_backwards=True)
+        z = torch.logsumexp(alpha[alpha.size(0) - 1] + self.crf_layer.end_transitions, dim=1)
+        prob = alpha + beta - z.view(1, -1, 1)
+        return torch.exp(prob).transpose(0, 1)
+
+    def set_params(self, **params):
+        self.feat_type = params.get('feat_type', DEFAULT_PYTORCH_CRF_ER_CONFIG['feat_type']).lower()
```
murali1996, line 277: nit: maybe you can do something like:
```
params = {
    **DEFAULT_PYTORCH_CRF_ER_CONFIG,
    **params
} 
```
and then use the params to initialize all the required vars; for a better readability.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+import os
+import random
+import uuid
+from collections import Counter
+from copy import copy
+from itertools import zip_longest
+from random import randint
+
+import numpy as np
+import torch
```
murali1996, line 11: Note for future: Since we plan to include torch as a requirement in MindMeld, some snippet in neural network codebase can be modified to directly `import torch` and use `nn.Module` and other torch functions directly.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+        self.classes = None
+        self.num_feats = num_feats
+        self.ready = False
+
+    def get_tensor_data(self, feat_dicts, labels=None, fit=False):
+        if labels is None:
+            labels = []
+        if fit:
+            if self.feat_extract_type == "dict":
+                comb_dict_list = [x for seq in feat_dicts for x in seq]
```
snow0x2d0, line 129: Nested comprehensions also tend to be difficult to read

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+                self.classes = self.label_encoder.classes_
+                self.num_classes = len(self.label_encoder.classes_)
+
+            self.ready = True
+        feats = []
+        encoded_labels = []
+        seq_lens = [len(x) for x in feat_dicts]
+        max_seq_len = max(seq_lens)
+
+        for i, (x, y) in enumerate(zip_longest(feat_dicts, labels)):
```
snow0x2d0, line 144: it's unclear from the context what the shape of `x` is

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+                values=torch.as_tensor(sparse_feat.data), size=sparse_feat.shape)
+            feats.append(sparse_feat_tensor)
+
+            if y:
+                transformed_label = self.label_encoder.transform(y)
+                transformed_label = np.pad(transformed_label, pad_width=(0, max_seq_len - seq_lens[i]),
+                                           constant_values=self.pad_index)
+                label_tensor = torch.as_tensor(transformed_label, dtype=torch.long)
+                encoded_labels.append(label_tensor)
+        return (feats, encoded_labels, seq_lens) if encoded_labels else (feats, seq_lens)
```
snow0x2d0, line 159: I'd suggest keeping the shape of the return value the same and just letting the optional value be `None`
```suggestion
        return (feats, seq_lens, encoded_labels)
```

snow0x2d0 (changes requested):

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+    def validate_params(self):
+        if self.optimizer_type not in ["sgd", "adam"]:
+            raise MindMeldError(
+                f"Optimizer type {self.optimizer_type} not supported. Supported options are ['sgd', 'adam']")
+        elif self.feat_type not in ["hash", "dict"]:
+            raise MindMeldError(f"Feature type {self.feat_type} not supported. Supported options are ['hash', 'dict']")
+        elif not 0 < self.train_val_split < 1:
+            raise MindMeldError("Train-dev split should be a value between 0 and 1.")
+        elif not 0 <= self.drop_input < 1:
+            raise MindMeldError("Drop Input should be a value between 0 and 1. (inclusive)")
```
snow0x2d0, line 185: These can just be single `if`'s, since you're raising if they're true.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+        dev_dataloader = DataLoader(dev_dataset, batch_size=512, shuffle=True, collate_fn=collate_tensors_and_masks)
+
+        best_dev_score, best_dev_epoch = -np.inf, -1
+        _patience_counter = 0
+
+        self.build_params(self.encoder.num_feats, self.encoder.num_classes)
+        if self.optimizer_type == "sgd":
+            self.optimizer = optim.SGD(self.parameters(), lr=0.01, momentum=0.9, nesterov=True, weight_decay=1e-5)
+        elif self.optimizer_type == "adam":
+            self.optimizer = optim.Adam(self.parameters(), weight_decay=1e-5)
```
snow0x2d0, line 333: I would just split this into two 'ifs'

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+                raise MindMeldError(f"{y} should be am integer value.")
+
+    def build_params(self, num_features, num_classes):
+        self.W = nn.Parameter(torch.nn.init.xavier_normal_(torch.empty(size=(num_features, num_classes))),
+                              requires_grad=True)
+        self.b = nn.Parameter(torch.nn.init.constant_(torch.empty(size=(num_classes,)), val=0.01),
+                              requires_grad=True)
+        self.crf_layer = CRF(num_classes, batch_first=True)
+        self.crf_layer.apply(init_weights)
+        self.num_classes = num_classes
```
snow0x2d0, line 199: We should just add these to `__init__` with None as the default value.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+            logger.warning(
+                "WARNING: Number of features is compatible with only `hash` feature type. This value is ignored with `dict` setting", )
+
+    # pylint: disable=too-many-locals
+    def fit(self, X, y):
+        self.set_random_states()
+        self.encoder = Encoder(feature_extractor=self.feat_type, num_feats=self.feat_num)
+        stratify_tuples = None
+        if self.stratify:
+            stratify_tuples = [tuple(sorted(list(set(label)))) for label in y]
```
snow0x2d0, line 302: I'd say maybe just pull this out into a simple function where you can provide some comments as to the purpose of each step here.

mindmeld/models/taggers/crf.py
```diff
@@ -182,6 +183,140 @@ def setup_model(self, config):
...
+        marginals_dict = self._clf.predict_marginals(X)
+        marginal_tuples = []
+        for query_index, query_seq in enumerate(seq):
+            query_marginal_tuples = []
+            for i, tag in enumerate(query_seq):
+                query_marginal_tuples.append([tag, marginals_dict[query_index][i][tag]])
+            marginal_tuples.append(query_marginal_tuples)
+        return marginal_tuples
+
+    def predict_proba_distribution(self, examples, config, resources):
```
ritvikshrivastava, line 227: This implementation in the earlier version of CRF used marginals to get the distribution instead of the actual prob distribution because of limitations with the earlier approach. With the torch-crf module we can change this method to output absolute probabilities instead (@vembar)

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+                probs = self.compute_marginal_probabilities(inputs, mask).tolist()
+                mask = mask.tolist()
+                # If anyone has any suggestions on a cleaner way to do this, I am all ears!
+                marginals_dict.extend([[dict(zip(self.encoder.classes, token_probs)) \
+                                        for (token_probs, valid_token) in zip(seq, mask_seq) if valid_token] \
+                                       for seq, mask_seq in zip(probs, mask)])
+
+        return marginals_dict
+
+    def predict(self, X):
```
ritvikshrivastava, line 401: Is this returning the distribution across labels per token or the top label per token? Can you add a docstring specifying that or otherwise?

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+
+        dataloader = DataLoader(torch_dataset, batch_size=512, shuffle=False, collate_fn=collate_tensors_and_masks)
+        marginals_dict = []
+        self.eval()
+        with torch.no_grad():
+            for inputs, mask in dataloader:
+                probs = self.compute_marginal_probabilities(inputs, mask).tolist()
+                mask = mask.tolist()
+                # If anyone has any suggestions on a cleaner way to do this, I am all ears!
+                marginals_dict.extend([[dict(zip(self.encoder.classes, token_probs)) \
```
ritvikshrivastava, line 395: Can we break this nested list comprehension into individual structures? The three level zipping in a single statement might become tedious to follow later

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Raised issue https://github.com/cisco/mindmeld/issues/415 for this!

mindmeld/models/tagger_models.py
vrdn-23: Raised https://github.com/cisco/mindmeld/issues/416 for this

snow0x2d0: I'd say just mention the issue number in the comment.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -273,66 +297,68 @@ def compute_marginal_probabilities(self, inputs, mask):
...
-        self.drop_input = params.get('drop_input', DEFAULT_PYTORCH_CRF_ER_CONFIG['drop_input'])
-        self.train_batch_size = params.get('train_batch_size', DEFAULT_PYTORCH_CRF_ER_CONFIG['train_batch_size'])
-        self.patience = params.get('patience', DEFAULT_PYTORCH_CRF_ER_CONFIG['patience'])
-        self.epochs = params.get('epochs', DEFAULT_PYTORCH_CRF_ER_CONFIG['epochs'])
-        self.train_val_split = params.get('train_val_split', DEFAULT_PYTORCH_CRF_ER_CONFIG['train_val_split'])
-        self.optimizer_type = params.get('optimizer_type', DEFAULT_PYTORCH_CRF_ER_CONFIG['optimizer_type']).lower()
-        self.random_state = params.get('random_state', randint(1, 10000001))
+    def set_params(self, feat_type="hash", feat_num=50000, stratify_train_val_split=True, drop_input=0.2, batch_size=8,
+                   patience=3, number_of_epochs=100, dev_split_ratio=0.2, optimizer="sgd",
+                   random_state=randint(1, 10000001)):
```
snow0x2d0, line 302: Here you'll probably want to set the default of random_state to `None` and then do something like:

```python
random_state = random_state or randint(1, 10000001)
```

As written the randint function will get called when the _class_ is created and then every function call would use the same random value, which I don't think is what you'd want.

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -0,0 +1,408 @@
...
+    elif len(sequence[0]) == 2:
+        sparse_mats, masks = zip(*sequence)
+        return diag_concat_coo_tensors(sparse_mats), torch.stack(masks)
+
+
+class Encoder:
+    def __init__(self, feature_extractor="hash", num_feats=50000):
+
+        self.feat_extractor = DictVectorizer(dtype=np.float32) if feature_extractor == "dict" else FeatureHasher(
+            n_features=num_feats, dtype=np.float32)
```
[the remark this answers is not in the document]
vrdn-23: Fixed

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Refactored to make it class variables so no longer applies.

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Fixed

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Refactored to make it more neat

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Done

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Refactored code and added comment to make it more clear

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Changed to itertool.chain()

mindmeld/models/taggers/pytorch_crf.py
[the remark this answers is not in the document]
vrdn-23: Split into different functions resulting in significantly lesser branches

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Fixed as mentioned

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Renamed to be consistent across NN models

mindmeld/models/tagger_models.py
```diff
@@ -246,8 +248,8 @@ def fit(self, examples, labels, params=None):
             self._current_params = params
         else:
             # run cross validation to select params
-            if self._clf.__class__ == LstmModel:
-                raise MindMeldError("The LSTM model does not support cross-validation")
+            if self._clf.__class__ in (LstmModel, TorchCrfTagger):
+                raise MindMeldError(f"The {self._clf.__class__.__name__} model does not support cross-validation")
```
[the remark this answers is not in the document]
vrdn-23: Fixed to suggested change.

mindmeld/models/taggers/pytorch_crf.py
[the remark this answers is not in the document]
vrdn-23: Resolved offline

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Refactored into a more readable format

mindmeld/models/tagger_models.py
vrdn-23: Done

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Fixed

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Fixed, I think

vrdn-23: We'll see if something breaks

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Done

mindmeld/models/taggers/crf.py
```diff
@@ -182,6 +183,140 @@ def setup_model(self, config):
...
 
+class TorchCrfTagger(Tagger):
+    """A Conditional Random Fields model."""
+
+    def fit(self, X, y):
+        self._clf.fit(X, y)
+        return self
+
+    def set_params(self, **parameters):
+        self._clf = TorchCrfModel()
```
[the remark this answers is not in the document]
vrdn-23: This involves a much larger refactoring across all existing tagger models. I've created https://github.com/cisco/mindmeld/issues/420 to address this.

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Added docstring

mindmeld/models/taggers/crf.py
vrdn-23: Created https://github.com/cisco/mindmeld/issues/421 to address this.

mindmeld/models/taggers/crf.py
```diff
@@ -182,6 +183,155 @@ def setup_model(self, config):
...
+        """
+        Args:
+            examples (list of mindmeld.core.Query): a list of queries to predict on
+            config (ModelConfig): The ModelConfig which may contain information used for feature
+                                  extraction
+            resources (dict): Resources which may be used for this model's feature extraction
+
+        Returns:
+            list of tuples of (mindmeld.core.QueryEntity): a list of predicted labels \
+             with confidence scores
```
snow0x2d0, line 238: A list of tuples doesn't seem to match what's actually being returned here.

snow0x2d0 (changes requested):

vrdn-23:
Created https://github.com/cisco/mindmeld/issues/424 to address the incremental build failure

mindmeld/models/helpers.py
```diff
@@ -534,6 +535,11 @@ def add_resource(func):
     return add_resource
 
 
+def np_encoder(val):
+    if isinstance(val, np.generic):
+        return val.item()
+
+
```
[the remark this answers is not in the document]
vrdn-23: Nice catch! Fixed.

mindmeld/models/taggers/crf.py
vrdn-23: Updated the docs accordingly

mindmeld/models/taggers/crf.py
```diff
@@ -182,6 +183,155 @@ def setup_model(self, config):
...
+        """
+        X, _, _ = self.extract_features(examples, config, resources, in_memory=True)
+        seq = self._clf.predict(X)
+        marginals_dict = self._clf.predict_marginals(X)
+        predictions = []
+        tag_maps = []
+        for query_index, query_seq in enumerate(seq):
+            tags = []
+            preds = []
+            for i in range(len(query_seq)):
```
tmehlinger, line 248: You can give a nicer visual cue using `enumerate` and throwing away the value.
```python
for i, _ in enumerate(query_seq):
    ...
```

tmehlinger (approved):

snow0x2d0 (approved):

ritvikshrivastava (approved):

mindmeld/models/helpers.py
```diff
@@ -534,6 +535,13 @@ def add_resource(func):
     return add_resource
 
 
+def np_encoder(val):
+    if isinstance(val, np.generic):
+        return val.item()
+    else:
```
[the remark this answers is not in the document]
vrdn-23: Thanks Travis! Fixed! :)

mindmeld/models/tagger_models.py
```diff
@@ -392,13 +396,21 @@ def _dump(self, path):
...
             # underneath tagger dump for LSTM model, returned `model_dir` is None for MEMM & CRF
-            self._clf.dump(path)
-            metadata.update({
-                "current_params": self._current_params,
-                "label_encoder": self._label_encoder,
-                "no_entities": self._no_entities,
-                "model_config": self.config
-            })
+            if isinstance(self._clf, TorchCrfTagger):
+                self._clf.dump(path)
```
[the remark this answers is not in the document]
vrdn-23: Done

mindmeld/models/taggers/crf.py
vrdn-23: Done

PR #418, opened 2022-06-11
Loosen marshmallow dependency

jre21:
The current dependency string only allows marshmallow 3.7.x versions.  This change allows it to update to latest 3.x.


jre21:
I'm nixing this change.  Two of our unit tests are only passing because of bugs in marshmallow's validator prior to 3.9.1 (it previously allowed serializations which should have been rejected).  At this point I'm not willing to take a chance on that bug being load-bearing for more than just the unit tests.

PR #426, opened 2022-07-12
Update MM code to support latest scikit-learn version

vrdn-23:
Hey folks,

This is (hopefully) the last PR that we have before we officially move to supporting the a newer (not the latest but hopefully that shouldn't be too large a leap) version of scikit-learn on Mindmeld. 

The main changes that this PR includes are the following:
- Changing the job lib imports to be from the actual job lib library and not from sklearn externals.
- Changing the structure of some inputs that were being passed into the inverse_trasform function of sklearn because it now requires all inputs to be a list.
- Removing the test dependencies to not include any tests from the old CRF library
- Fixing some flaws in the merge param PR that I overlooked previously
- Adding validation for params in the new CRF library in order to drop params that are not relevant.

Once we verify that things work well, with ANLP as well, I will introduce a new PR to remove the existing crf-suite implementation from the Mindmeld library completely.


mindmeld/components/classifier.py
```diff
@@ -403,6 +403,8 @@ def _get_model_config(loaded_config=None, **kwargs):
         Returns:
             ModelConfig: The model configuration corresponding to the provided config name
         """
+        if 'params' in loaded_config and 'params' in kwargs:
+            kwargs['params'].update(loaded_config['params'])
```
snow0x2d0, line 407: This feels backwards. We're overwriting the passed in params with the existing config? 

I think that second line should just be something like:
`merged = {**loaded_config['params'], **kwargs['params']}`

snow0x2d0 (changes requested):
Just one minor issue (unless I misunderstood what we're doing), otherwise LGTM.

mindmeld/components/classifier.py
vrdn-23: Nice catch! 
gifntext-gif
** BLOCKED BY JAMES **

mindmeld/models/taggers/memm.py
```diff
@@ -175,7 +175,7 @@ def _predict_proba_distribution_example(self, example, config, resources):
             predictions.append(list(prediction))
             tag_maps.append(
                 [
-                    self.class_encoder.inverse_transform(i) for i in range(len(prediction))
+                    self.class_encoder.inverse_transform([i]) for i in range(len(prediction))
```
vembar, line 178: Going by the changes in line 153, should this be as follows?
```suggestion
                    self.class_encoder.inverse_transform([i])[0] for i in range(len(prediction))
```

vembar (approved):

mindmeld/components/classifier.py
```diff
@@ -403,6 +403,9 @@ def _get_model_config(loaded_config=None, **kwargs):
         Returns:
             ModelConfig: The model configuration corresponding to the provided config name
         """
+        if 'params' in loaded_config and 'params' in kwargs:
+            merged_params = {**loaded_config['params'], **kwargs['params']}
+            kwargs['params'] = merged_params
```
ritvikshrivastava, line 408: would a one-liner work here?
```suggestion
            kwargs['params'] = {**loaded_config['params'], **kwargs['params']}
```

ritvikshrivastava (approved):

snow0x2d0 (approved):

mindmeld/models/taggers/memm.py
vrdn-23: You're right! I've fixed that! Thanks Varsha!

mindmeld/components/classifier.py
vrdn-23: Refactored!

serapio (approved):

PR #432, opened 2022-08-08
Major bug fix for PyTorch CRF

vrdn-23:
There are two minor fixes in the PyTorch-CRF code that this PR resolves:
1) The best path for a model was not saving the absolute path the first time and hence, the model loading was throwing issues when invoked from a directory that was different from when training was called.
2) The tempfile function I was using returned the same temp directory each time and this would cause the code to save the same model across different entity recognition models.

UPDATE:
This PR has now evolved into changing the way we now dump and load PyTorch CRF models. We no longer dump the entire class object to the entity pickle, but rather save only the config and encoder details, thus following a approach more in line with the LSTM model.

### DIFF
--- a/mindmeld/models/taggers/pytorch_crf.py
+++ b/mindmeld/models/taggers/pytorch_crf.py
@@ -273,8 +273,6 @@ class TorchCrfModel(nn.Module):
         self.optimizer = None
         self.random_state = None
 
-        # self.best_model_save_path = None
-        self.ready = False
         self.tmp_save_path = os.path.join(mkdtemp(), "best_crf_wts.pt")
 
     def get_encoder(self):
@@ -368,7 +366,7 @@ class TorchCrfModel(nn.Module):
             inputs.values()[:] = inputs.values() * dp_mask
         dense_W = torch.tile(self.W, dims=(mask.shape[0], 1))
         out_1 = torch.addmm(self.b, inputs, dense_W)
-        crf_input = out_1.reshape((mask.shape[0], -1, self.num_classes or self.encoder.num_classes))
+        crf_input = out_1.reshape((mask.shape[0], -1, self.num_classes))
         if targets is None:
             return self.crf_layer.decode(crf_input, mask=mask)
         loss = - self.crf_layer(crf_input, targets, mask=mask)
@@ -445,7 +443,7 @@ class TorchCrfModel(nn.Module):
         # SWITCHING FOR BATCH FIRST DEFAULT
         dense_W = torch.tile(self.W, dims=(mask.shape[0], 1))
         out_1 = torch.addmm(self.b, inputs, dense_W)
-        emissions = out_1.reshape((mask.shape[0], -1, self.num_classes or self.encoder.num_classes))
+        emissions = out_1.reshape((mask.shape[0], -1, self.num_classes))
         emissions = emissions.transpose(0, 1)
         mask = mask.transpose(0, 1)
         alpha = self._compute_log_alpha(emissions, mask, run_backwards=False)
--- a/mindmeld/models/taggers/pytorch_crf.py
+++ b/mindmeld/models/taggers/pytorch_crf.py
@@ -6,7 +6,7 @@ from collections import Counter
 from copy import copy
 from itertools import chain
 from random import randint
-from tempfile import gettempdir
+from tempfile import mkdtemp
 
 import numpy as np
 import torch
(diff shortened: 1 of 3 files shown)


snow0x2d0 (approved):
LGTM

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -272,7 +272,7 @@ def __init__(self):
 
         self.best_model_save_path = None
         self.ready = False
-        self.tmp_save_path = os.path.join(gettempdir(), "best_crf_wts.pt")
+        self.tmp_save_path = os.path.join(mkdtemp(), "best_crf_wts.pt")
         # os.makedirs(os.path.dirname(self.tmp_save_path), exist_ok=True)
```
serapio, line 276: rm

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -287,7 +287,7 @@ def save_best_weights_path(self, path):
         Args:
             path (str): Path to save the best model weights.
         """
-        self.best_model_save_path = path
+        self.best_model_save_path = os.path.abspath(path)
```
serapio, line 290: Does this block models from being archived/copied?

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Done!

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Fixed that by changing the way pyTorch-CRF loads and dumps models

snow0x2d0 (approved):

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -287,10 +293,17 @@ def save_best_weights_path(self, path):
...
-            torch.save(best_weights, self.best_model_save_path)
+        torch.save(self.state_dict(), path)
+
+    def load_best_weights_path(self, path):
+        """Saves the best weights of the model to a path in the .generated folder.
+
+        Args:
+            path (str): Path to save the best model weights.
+        """
+        # self.best_model_save_path = os.path.abspath(path)
```
vembar, line 304: stray line

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -485,7 +498,7 @@ def get_dataloader(self, X, y, is_train):
             torch_dataloader (torch.utils.data.dataloader.DataLoader): returns PyTorch dataloader object that can be
             used to iterate across the data.
         """
-        tensor_inputs, input_seq_lens, tensor_labels = self.encoder.get_tensor_data(X, y, fit=is_train)
+        tensor_inputs, input_seq_lens, tensor_labels = self._encoder.get_tensor_data(X, y, fit=is_train)
         tensor_dataset = TaggerDataset(tensor_inputs, input_seq_lens, tensor_labels)
         torch_dataloader = DataLoader(tensor_dataset, batch_size=self.batch_size if is_train else 512, shuffle=is_train,
```
vembar, line 503: Should we store this batch size default as a constant?

vembar (approved):

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Done

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Removed

mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -516,15 +530,15 @@ def fit(self, X, y):
...
         gc.collect()
 
-        self.build_params(self.encoder.num_feats, self.encoder.num_classes)
+        self.build_params(*self._encoder.get_feats_and_classes())
 
         if self.optimizer == "sgd":
             self.optim = optim.SGD(self.parameters(), lr=0.01, momentum=0.9, nesterov=True, weight_decay=1e-5)
         if self.optimizer == "adam":
-            self.optim = optim.Adam(self.parameters(), weight_decay=1e-5)
+            self.optim = optim.Adam(self.parameters(), lr=0.001, weight_decay=1e-5)
```
serapio, line 538: Do we really want these hard-coded?

serapio (approved):

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: I don't anticipate us ever having to change those values in the future, and also I don't know if a user would ever care about those values enough for us to move them into constant variables. For readability's sake, I am inclined to keep them as is.

PR #436, opened 2022-08-31
Improve features for CRF

vrdn-23:
Adding some improvements to the torch-crf so that we can run ANLP experiments quite nicely.

The main changes in this PR include:
-  Adding support for the lbfgs optimizer
- Including a configurable regularization weight for l1 and l2 loss
- Including a learning rate scheduler that kicks in if Dev F1 doesn't increase
- Changing CRF loss reduction from 'sum' to 'mean'

### DIFF
--- a/mindmeld/models/taggers/pytorch_crf.py
+++ b/mindmeld/models/taggers/pytorch_crf.py
@@ -274,7 +274,6 @@ class TorchCrfModel(nn.Module):
         self.dev_split_ratio = None
         self.optimizer = None
         self.random_state = None
-        self.keep_validation_fixed = None
 
         self.tmp_save_path = os.path.join(mkdtemp(), "best_crf_wts.pt")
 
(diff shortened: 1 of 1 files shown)


mindmeld/models/taggers/pytorch_crf.py
```diff
@@ -575,11 +608,24 @@ def train_one_epoch(self, train_dataloader):
...
-            self.optim.step()
+            def closure():
+                nonlocal train_loss
+                self.optim.zero_grad()
+                # pylint: disable=cell-var-from-loop
+                loss = self.forward(inputs, labels, mask, drop_input=self.drop_input)
+                if self.reg_weight > 0:
+                    if self.optimizer == "lbfgs":
+                        loss += self.compute_regularized_loss(l1=False)
+                    loss += self.compute_regularized_loss(l1=True)
```
serapio, line 619: It feels a little weird to hard-code that they always use balanced L1+L2. Is there any problem with using separate `reg_weight_l1` and `reg_weight` (L2)?

serapio (approved):

mindmeld/models/taggers/pytorch_crf.py
vrdn-23: Nope. You're right! Ideally we should have separate params for both! I've fixed that!

snow0x2d0 (approved):

Issue #416, opened 2022-06-08
Remove existing dependency of CRFSuite


vrdn-23:
Once the [PyTorch CRF PR](https://github.com/cisco/mindmeld/pull/413) gets merged, and we are good to make sure nothing fails, we can safely remove the dependency on CRFSuite and also remove/refactor the old CRFSuite code out of the mindmeld repo. 
It would also make sense to rename the "torch-crf" as "crf" once it becomes the default CRF implementation.

This was mentioned as part of [this PR discussion](https://github.com/cisco/mindmeld/pull/413#discussion_r882150030).

Issue #415, opened 2022-06-08
Fix import statements for torch in neural model files


vrdn-23:
With the introduction of a PyTorch based CRF (https://github.com/cisco/mindmeld/pull/413), the torch library is now a required dependency for mindmeld. This means we can now clean up the [try-except import torch statements](https://github.com/cisco/mindmeld/blob/0eeba592e78720b8097de2979a7788d3447398b2/mindmeld/models/nn_utils/classification.py#L54) found across the neural networks files in Mindmeld. 

This was first raised as a comment [here](https://github.com/cisco/mindmeld/pull/413#discussion_r882221639).

PR #446, opened 2022-10-31
<refactor> remove sklearn crf and update torch checks

ritvikshrivastava:
Addressing issues #415 and #416


vrdn-23:
We also need to remove the sklearn-crfsuite dependency from our requirements.txt

vrdn-23 (approved):
lgtm

Issue #424, opened 2022-06-29
Loading fails after two consecutive incremental builds


vrdn-23:
The incremental build of models in Mindmeld currently exhibits a weird behavior where consecutive incremental builds cannot be loaded unless the dump functionality is explicitly called. This behavior can be reproduced by the following code snippet:
```
nlp = NaturalLanguageProcessor(kwik_e_mart_app_path)
nlp.build(incremental=True)
initial_timestamp = nlp.incremental_timestamp

nlp = NaturalLanguageProcessor(kwik_e_mart_app_path)
nlp.build(incremental=True)
new_timestamp = nlp.incremental_timestamp

nlp.load(initial_timestamp)
nlp.load(new_timestamp)
```
The current behavior leads to a FileNotFoundError as we try to load the non-incremental models from the non-timestamp path. 
```
FileNotFoundError: [Errno 2] No such file or directory: 'kwik_e_mart/.generated/cached_models/20220628T172423/domains/banking/transfer_money/entity.pkl 
```
One way to bypass this error, is to explicitly call dump after each build statement, but this does nit align with the expected behavior.

PR #447, opened 2022-11-01
Check if incremental path exists before loading

vrdn-23:
This solves the issue described in #424. We now check if the incremental path exists before trying to load it. If it does not exist, we fall back to loading from the default model path. 

I've also updated a unit test to verify this behavior.

### DIFF
--- a/mindmeld/components/nlp.py
+++ b/mindmeld/components/nlp.py
@@ -510,9 +510,11 @@ class NaturalLanguageProcessor(Processor):
         model_path, incremental_model_path = path.get_domain_model_paths(
             app_path=self._app_path, timestamp=incremental_timestamp
         )
+        incremental_model_path_exist = os.path.exists(
+            incremental_model_path) if incremental_model_path is not None else False
 
         self.domain_classifier.load(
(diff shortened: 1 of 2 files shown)


mindmeld/components/nlp.py
```diff
@@ -510,9 +510,11 @@ def _load(self, incremental_timestamp=None):
         model_path, incremental_model_path = path.get_domain_model_paths(
             app_path=self._app_path, timestamp=incremental_timestamp
         )
+        incremental_model_path_exist = os.path.exists(
+            incremental_model_path) if incremental_model_path is not None else False
```
serapio, line 514: Should this check be done inside `get_domain_model_paths()` instead? It seems it should provide `incremental_model_path` as `None` if there's nothing there.

vrdn-23: Actually get_domain_model_paths() is used to generate the path for both loading and dumping. So it'll be hard without adding another parameter to check if I should return None when we're loading vs before dumping before the file is created

serapio (approved):

snow0x2d0 (approved):

PR #461, opened 2023-01-31
Regenerate docs pages without cookie manager

serapio:
1. Remove the cookie manager from the docs pages template
2. Update the copyright
3. Fix a few broken references from duplicate labels

Note: Does not include history notes for 4.6 and 4.7


vrdn-23 (approved):
lgtm.
Although I am curious to know why we're doing the sentencepiece update?

serapio:
That was from it not finding an installable package. We probably want to switch to `~=` match, but I've taken it out since doesn't belong here.

